PR #1, opened 2016-03-28
use one internal constructor, many external ones

jeffreysarnoff-dev:
This is best practice.  
One internal constructor is enough to disengage provision of implicit constructors.
There is no incremental benefit to putting others inside the type, as they behave just the same written externally.  Julia expects none or one or possibly a very few of the constructors for a type to be internally given and the rest to be external.  External constructors are not `special`, so require no special attention.


thofma:
1. I don't think the constructors should be external constructors.
2. This code is not working. The function `new` as you use it is only defined in type declarations.
3. The file mag.jl (as well as other arb related files) is not used anymore. Most of the stuff is now in Nemo. I will delete it in the near future.

jeffreysarnoff-dev:
one of the rare occasions that I posted something I wanted to consider myself

thank you for being responsive to my dev questions 
 
 this will disappear shortly

PR #16, opened 2017-05-02
only define factor and isprime for Integer if not extending Base

tkelman:
otherwise importing Hecke can globally modify method tables and change the behavior of unrelated code that's trying to use Primes.jl, for example

### DIFF
--- a/src/Misc/Integer.jl
+++ b/src/Misc/Integer.jl
@@ -4,8 +4,12 @@
 #
 ################################################################################
 
-isprime(a::Integer) = isprime(fmpz(a))
-factor(a::Integer) = factor(fmpz(a))
+if !isdefined(Base, :isprime)
+  isprime(a::Integer) = isprime(fmpz(a))
+end
+if !isdefined(Base, :factor)
+  factor(a::Integer) = factor(fmpz(a))
+end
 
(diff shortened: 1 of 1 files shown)


thofma:
Unfortunately not that easy. Now tests are failing:
https://travis-ci.org/thofma/Hecke.jl/jobs/228014244

tkelman:
Could alternately change Nemo to not import these from Base, as another solution. That way you only have side effects for code that uses Nemo, not code that uses Primes (which was the formerly-in-base code).

thofma:
Hm, so actually the "problem" is in Nemo, where the functions are overloaded, right?

tkelman:
You also have https://github.com/thofma/Hecke.jl/blob/73abfd83777fbef3cf94cdef72ce279ec1026b9a/src/NfMaxOrd/Ideal.jl#L36.

It's a combination of both - Nemo imports it from Base but only extends it in package-specific ways as far as I can tell, but then you import it from Nemo (which ends up being the same Base method, if Nemo is importing it from there) and extend it on Base types causing the global side effects.

thofma:
Yeah, thanks for the reminder. Our imports/exports are a mess. I will think of something. Don't spend more time on it. I don't want to steal precious julia-0.6.-preparation-time.

PR #42, opened 2019-09-06
remove length(::MatElem) (now in AbstractAlgebra)

rfourquet:
Cf. https://github.com/Nemocas/AbstractAlgebra.jl/pull/383.


thofma:
This breaks Hecke, since we do not develop Hecke against the master version of Nemo.

The process is:
1) Move stuff to Nemo.
2) Make a new Nemo release.
3) Bump version number of Nemo as a dependecy of Hecke.
4) Remove the stuff from Hecke.

I suggest you move as much stuff over as you want and then tag a new Nemo version.

rfourquet:
Ok, will do.

PR #44, opened 2019-09-20
remove undefined export symbols

rfourquet:
These were exported but not defined:
cols, den, pmaximal_order, rows, update.


thofma:
Thanks!

PR #47, opened 2019-10-02
Local fields

a-kulkarn:
Not saying you should merge this now, but just making sure you're aware it exists before I leave for a week. Some notes:

-- Depends on the most recent version of Nemo.
-- I reorganized the "LocalField/Poly.jl" file a lot, moving some functions into "padic/qadic" where they belonged. There will be many differences from master.
-- EisensteinFields can tower and still work fine. Root-finding will work for polynomials over these objects.
-- the files "basic_eis_tests.jl" and "spaghetti_eis.jl" are meant for development, and are not part of the main module yet. I'm planning to absorb the code into the appropriate spot in Hecke.


src/LocalField/Eisenstein.jl
```diff
@@ -0,0 +1,209 @@
...
+  if d > 0
+    print(io, "Real quadratic field by ", q.pol)
+  else
+    print(io, "Imaginary quadratic field by ", q.pol)
+  end
+end
+
+function quadratic_field(d::Integer; cached::Bool = true, check::Bool = true)
+  return quadratic_field(fmpz(d), cached = cached, check = check)
+end
```
thofma: Why did these functions move here?

src/LocalField/basic_eis_tests.jl
```diff
@@ -0,0 +1,89 @@
+
+using Hecke
+using Crayons
```
thofma: Don't think we want this dependency :)

thofma:
Thanks Avi. I just had a quick look and noticed these two things above.

src/LocalField/Eisenstein.jl
a-kulkarn: They didn't actually, the original functions in the NumberField library are still there. The "Eisenstein.jl" file is not included in the module via an "include(...)" in LocalFields.jl. 

Rather, because Eisenstein fields behave similarly to number fields, I adapt and specialize the interface when it makes sense, as well as try to mimic the style to maintain code consistency.

fieker:
On Thu, Oct 03, 2019 at 04:31:07AM -0700, a-kulkarn wrote:
> a-kulkarn commented on this pull request.
> 
> 
> 
> > +  set_special(q, :show => show_quad)
> +  return q, a
> +end
> +
> +function show_quad(io::IO, q::EisensteinField)
> +  d = trail(q.pol)
> +  if d > 0
> +    print(io, "Real quadratic field by ", q.pol)
> +  else
> +    print(io, "Imaginary quadratic field by ", q.pol)
> +  end
> +end
> +
> +function quadratic_field(d::Integer; cached::Bool = true, check::Bool = true)
> +  return quadratic_field(fmpz(d), cached = cached, check = check)
> +end
> 
> They didn't actually, the original functions in the NumberField library are still there. The "Eisenstein.jl" file is not included in the module via an "include(...)" in LocalFields.jl. 
> 
> Rather, because Eisenstein fields behave similarly to number fields, I adapt and specialize the interface when it makes sense, as well as try to mimic the style to maintain code consistency.
> 
However, I vaguely recall that >0, <0 requires an ordered field, hence
is restricted to subfields of the reals. Thus this does not apply to
p-adics...
And I hope that neither padics nor quadics support >0 at all

a-kulkarn:
Indeed, as you can tell, I haven't converted/pruned this part of the
interface yet. As I've pointed out, nothing in this particular file is
actually ever interpreted. As I pointed out, this branch is still a work in
progress.

a-kulkarn:
Polished the interface and added convenience functionality. Tests exist, though I have not installed these in the test directory yet.

thofma:
The code here has helped a lot to explore the design space for local fields and to find an implementation which suits the applications in the Hecke/Oscar world best. There were still some design problems which we discussed the last few weeks and based on that there is now an implementation of local fields and completions by @CarloSircana.

PR #48, opened 2019-10-04
Subfield

martinra:
Is this how you would like _subfield_basis to be? I also replaced the randomized search for the primitive element by the deterministic one from my snippet.


thofma:
Perfect, thanks a lot!

PR #50, opened 2019-10-08
remove zero!(::MatElem)

rfourquet:
It's now available in AbstractAlgebra.
This is causing the test failure in #49, because this version here assumes the entries of the input matrix are not undefined. But now `zero(a)` calls `zero!(similar(a))`, and `similar(a)` has undefined entries.

### DIFF
--- a/src/Misc/Matrix.jl
+++ b/src/Misc/Matrix.jl
@@ -41,15 +41,6 @@ coefficient_type(::Type{Generic.Mat{T}}) where {T} = T
 #
 ################################################################################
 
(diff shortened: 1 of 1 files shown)


rfourquet:
The tests will fail, can they be cancelled to save resources?

thofma:
@rfourquet I am cleaning up Hecke myself. I took your commit for the perm change.

You don't have to work on this, thanks.

rfourquet:
> You don't have to work on this, thanks.

OK
cc @wbhart

PR #73, opened 2020-06-10
Set prec val length

wbhart:
Update set_length!, set_prec!, set_val! and set_scale! to return their argument to support polynomials and series with immutable types.

Requires AbstractAlgebra # 602 and Nemo # 841 to be merged first. It is not urgent that this patch be applied immediately after those two. This patch is also not breaking.

Oscar.jl and Singular.jl are unaffected by these changes and do not need to be updated.


wbhart:
Obviously the tests will not pass until the AbstractAlgebra/Nemo patches have been merged. But they are breaking so should be held until we do breaking releases.

PR #86, opened 2020-09-25
update few rand methods according to the Random API

rfourquet:
This is a modest first step, which concerns only `rand` methods taking only one object (i.e. this does not use RandomExtensions.jl).
This PR depends on the merged https://github.com/Nemocas/Nemo.jl/pull/878, which is not a breaking change (so a Nemo patch release could be made).

### DIFF
--- a/src/AlgAss/AbsAlgAss.jl
+++ b/src/AlgAss/AbsAlgAss.jl
@@ -550,8 +550,11 @@ end
 #
 ################################################################################
 
-function rand(A::AbsAlgAss{T}) where T
-  c = T[rand(base_ring(A)) for i = 1:dim(A)]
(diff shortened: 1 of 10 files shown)


thofma:
Thanks!

thofma:
I bumped the versions, but it is failing.

rfourquet:
I don't think you included the rand-related changes in the 0.18.2 Nemo release...

thofma:
Oh, I typed 0.18.2 instead of 0.18.3 in the PR here, but if you look at the log of the 1.0 run, he is pulling in Nemo 0.18.3, which got released a few minutes ago.

rfourquet:
Ok, I'm not at the computer now but will fix that later today.

PR #92, opened 2020-10-28
fix "docstring odditites"

rfourquet:
Similar to Nemocas/AbstractAlgebra.jl#662:

1. align signature to 4 spaces
2. add a line between the signature and the description
3. remove > characters at the beginning of lines in docstrings.


src/NfOrd/Ideal/Prime.jl
```diff
@@ -525,7 +528,7 @@ end
                        degree_limit::Int = 0, index_divisors::Bool = true) -> Array{NfOrdIdl, 1}
 
 Computes the prime ideals $\mathcal O$ with norm up to $B$.
->
+
```
rfourquet, line 531: In some cases like here, there was a `>` character to separate the pargraphs, which themselves didn't have a leading `>` in their lines. I'm not sure what was the intent, but I can easily revert this (undelete `>` which separate paragraphs).

thofma: I think these are just artifacts from a previous attempt on cleaning up the docstrings :)

PR #93, opened 2020-10-28
use RandomExtensions.make for  NfAbsNS, AlgAssAbsOrd and NfRel

rfourquet:


thofma:
I think there is something missing.

rfourquet:
> I think there is something missing.

Yes sorry, running the tests locally is really hard memory-wise on my laptop, so hadn't catched the problem. I should have the fix soon.

rfourquet:
Tests passed now, except on Julia 1.0 and 1.1, because inference needs some help to determine the return type of `rand` (and this is tested with `@inferred`)... I pushed a fix, the 3rd time should (hopefully) be good!

PR #96, opened 2020-10-31
Fix for julia master

thofma:
@fieker julia changed the hashing for Integers (including BigInt) on master here: https://github.com/JuliaLang/julia/pull/38031, which broke the LocalField tests. My change here make it pass again, but I am wondering if there is an underlying issue we need to fix?

### DIFF
--- a/test/LocalField/Conjugates.jl
+++ b/test/LocalField/Conjugates.jl
@@ -17,16 +17,29 @@
 
    @testset "Completion" begin
    
-     L1, mL1 = completion(k, 37, 1)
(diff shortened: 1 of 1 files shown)


fieker:
There is an issue, I would have hoped that the ordering is independent of hashing, but maybe thats our old problem of nmod_poly factoring/ roots being "random" in the order of returned values. The test is too fragile.
"solutions": make the test more robust (by checking maths) and/or make the result of factoring/ completing independent of hashing?

thofma:
OK, I will merge this here and we will think about it.

PR #99, opened 2020-11-04
type annotation to can_solve and some friends

fieker:
possibly should be replicated/ moved to AA as well?

### DIFF
--- a/src/Misc/Matrix.jl
+++ b/src/Misc/Matrix.jl
@@ -1921,8 +1921,7 @@ end
 Tries to solve $Ax = B$ for $x$ if `side = :right` or $xA = B$ if `side = :left`
 over a euclidean ring.
 """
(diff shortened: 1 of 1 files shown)


fieker:
does not work, can_solve is used explicitly with matrices of different type: a view and matrix.
So I close this for now

thofma:
Do you remember the input for which this was necessary?

fieker:
On Wed, Nov 04, 2020 at 04:30:15AM -0800, thofma wrote:
> Do you remember the input for which this was necessary?

Necessary is difficult: any can_solvve_with_kernel over quo(ZZ, Int)
is not getting the type inference done.
The stuff in "types" was an attempt to help julia.
In the end, it was resolved by "divrem" and the new type annotations
there.

PR #102, opened 2020-11-04
Convert Windows to Unix line ends

fingolfin:


thofma:
People will continue to add Windows line ends, so I don't know how useful this change is.

PR #112, opened 2020-12-11
collection of minor sanitisations

fieker:


src/Misc/Primes.jl
```diff
@@ -116,6 +116,8 @@ struct PrimesSet{T}
       end
       p = next_prime(p, false)
     end
+    mod = abs(mod)
+    val = abs(val)
```
thofma, line 120:
```suggestion
    val = mod(val, mod)
    val = val < 0 ? val + mod : val
```
Does this make sense? I think we allow the restriction to be "primes which are -12 modulo 5".

thofma, line 120:
```suggestion
    val = mod(val, mod)
```
To make sure this works if the input values have mixed signs.

thofma:
You can ignore the tests on 1.1.

src/Misc/Primes.jl
```diff
@@ -117,7 +117,7 @@ struct PrimesSet{T}
       p = next_prime(p, false)
     end
     mod = abs(mod)
-    val = abs(val)
+    val = abs(val % mod)
```
thofma, line 120: I don't think this is quite right.
```
julia> -1 % 3
-1
```
So this code turns -1 mod 3 into 1 mod 3.

fieker:
On Fri, Dec 11, 2020 at 03:51:22AM -0800, thofma wrote:
> 
> 
> 
> > @@ -117,7 +117,7 @@ struct PrimesSet{T}
>        p = next_prime(p, false)
>      end
>      mod = abs(mod)
> -    val = abs(val)
> +    val = abs(val % mod)
> 
> I don't think this is quite right.
> ```
> julia> -1 % 3
> -1
> ```
> So this code turns -1 mod 3 into 1 mod 3.

Good thinking!

src/Misc/Primes.jl
thofma: mod(x, y) will always be positive if y is positive. So you can just use `val = Base.mod(val, mod)`.

fieker:
On Fri, Dec 11, 2020 at 03:53:39AM -0800, thofma wrote:
> 
> 
> 
> > @@ -117,7 +117,7 @@ struct PrimesSet{T}
>        p = next_prime(p, false)
>      end
>      mod = abs(mod)
> -    val = abs(val)
> +    val = abs(val % mod)
> 
> mod(x, y) will always be positive if y is positive. So you can just use `val = Base.mod(val, mod)`.
where is this definition?

thofma:
Docstring of mod, see `?mod`.

fieker:
On Fri, Dec 11, 2020 at 03:58:33AM -0800, thofma wrote:
> Docstring of mod, see `?mod`. 
you win

PR #116, opened 2020-12-17
mpoly factor shuffle

tthsqe12:
Some things to note:

`modular_proj(::Generic.MPoly{nf_elem}, ::Hecke.modular_env)` was projecting to nmod, which doesn't work in general, so I added a `modular_proj(::Type{fq_nmod}, ::Generic.MPoly{nf_elem}, ::Hecke.modular_env)`, which does work in general. Given that `fq_nmod_mpoly` exists now (hopefully?), maybe the nmod things should go away?

The `induce_crt` function now works in the case of zero inputs.


thofma:
Did you intent to remove the test code?

tthsqe12:
Yeah, there is no way it will pass now. I also need `AbstractAlgebra.factor_squarefree`, so that we can extent it.

tthsqe12:
Feel free to move the test code to the right place.

PR #129, opened 2021-01-14
Better hash for fmpz_mat

thofma:


rfourquet:
Could this be done in `Nemo`?

thofma:
For this it would be good to remove the one missing allocation.

fieker:
hashing of fmpz is done in julia, hence there would be no problem in removing the allocation/ copying...

fieker:
can't you just call hash(unsafe_pointer_to_objref(m)) without the fmpz_set?

fieker:
OK, this works:
       m = ccall((:fmpz_mat_entry, libflint), Ptr{fmpz}, (Ref{fmpz_mat}, Int, Int), M, i - 1, j - 1)
        b = xor(b, xor(hash(unsafe_load(m), h), h))
and doesn't need ant fmpz_set or allocation

fieker:
it allocates - why, I don't know (yet)

fieker:
Sorry I am confused.
unsafe_load in hash -> hasing without allocation
in other applications -> allocations, but here, it seems fine and seems to solve the problem

thofma:
Should call the "shiny new junk" here https://github.com/Nemocas/Nemo.jl/blob/master/src/flint/mpoly.jl#L67.

fieker:
On Thu, Nov 18, 2021 at 03:18:41AM -0800, Tommy Hofmann wrote:
> Should call the "shiny new junk" here https://github.com/Nemocas/Nemo.jl/blob/master/src/flint/mpoly.jl#L67.
> 
Does not work directly. Views have no entries ptr - I learned this the
hard way...

fieker:
yo'd need to call the new stuff on the rows array individually.

thofma:
This is what he suggested.

PR #146, opened 2021-02-01
add IJulia back in and trivia Oscar -> Hecke

fieker:


fieker:
the eval is (currently) not needed I can disable this is 1.1 and 1.0 are important

thofma:
For the time being I would prefer 1.0 still working. If it is not too much hassle.

PR #160, opened 2021-02-12
Don't export functions that don't exist

thofma:
@wbhart This function was removed, but the export statement was not adjusted. Only shows up as warnings in Oscar.


wbhart:
Thanks.

PR #162, opened 2021-02-15
Very first start factor absolute

tthsqe12:
I am in no way claiming this is a good interface for factor_absolute. It is just the first thing I threw together to get the answer out.
There is a problem in recognizing abs irreducible stuff (the code in this pr has a slightly better error message):
```
julia> using Hecke
julia> Hecke.example("MPolyFact.jl")
julia> Qxy,(x,y)=QQ["x","y"];MPolyFact.absolute_bivariate_factorisation(x^2 + 2*y^2 + 2)
ERROR: AssertionError: factorization is not correct mod p^1
Stacktrace:
 [1] lift_prime_power(a::fmpq_mpoly, fac::Vector{AbstractAlgebra.Generic.MPoly{qadic}}, alphas::Vector{Int64}, kstart::Int64, kstop::Int64)

### DIFF
--- a/examples/MPolyFact.jl
+++ b/examples/MPolyFact.jl
@@ -972,7 +972,6 @@ function lift_prime_power(
         return
     end
 
-
     r = length(fac)
     R = parent(fac[1])
     n = nvars(R)
@@ -991,6 +990,13 @@ function lift_prime_power(
 
     for l in kstart:kstop
         error = a - prod(fac)
+
+        for c in coeffs(error)
+          if valuation(c) < l
+            throw(AssertionError("factorization is not correct mod p^$l"))
+          end
+        end
+
         if iszero(error)
             break
         end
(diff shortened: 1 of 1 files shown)


tthsqe12:
@thoma It looks better after rebase. Any ideas on how to return an absolute factorization?

thofma:
Don't know. Create a number field?

fieker:
On Mon, Feb 15, 2021 at 01:11:29PM -0800, thofma wrote:
> Don't know. Create a number field?
> 
A field per factor...

fieker:
On Mon, Feb 15, 2021 at 01:11:29PM -0800, thofma wrote:
> Don't know. Create a number field?

plus some over Q if we're (un)lucky.

good night!

... and thanks a lot
 Claus

thofma:
We could still return a factor object. The types of the factors are the same (may need a trivial number field for this).

thofma:
Should this be merged? I made a suggestion for the return value, but we can also merge as is.

tthsqe12:
The return value is currently a mess. The problem is the factors in extensions. If f is irreducible over QQ, then it is factored into g * f/g where g is abs irreducible and defined over a nf. f/g is not nec abs irreducible. The problem is these two factors should come out as an ordered pair because they should know about each other.

thofma:
I thought the absolute irreducible factorization returns absolutely irreducible factors?

tthsqe12:
no, there are problems with doing that

PR #165, opened 2021-02-16
Fix class group proof

thofma:


fieker (approved):
Looks like a no-brainer to me....

PR #175, opened 2021-02-19
Remove ^(::fmpz, ::fmpz) because of (probable) implementation in Nemo

albinahlback:
See https://github.com/Nemocas/Nemo.jl/pull/985


thofma:
It seems like this is actually not called in the tests. I will still wait for the new Nemo release before merging this.

PR #196, opened 2021-03-03
fix a bug in sub which resulted in different moduli 

simonbrandhorst:
and a bug in normal_form for TorQuadMod


thofma:
Do you want to add a test?

simonbrandhorst:
Alright.

thofma:
Thanks!

PR #223, opened 2021-03-20
Test cleanup action

thofma:


thofma:
This cleans up/cancels old runs on the same branch.

@benlorenz We talked about this yesterday. I found https://github.com/rokroskar/workflow-run-cleanup-action. It is not perfect, since of course the clean up job is itself in the queue, but it helps when quickly pushing commits to the same branch.

@fingolfin You might also be interested (maybe you already knew this).

I plan to roll this out to AA/Nemo/Oscar.

PR #232, opened 2021-03-23
Fix simplify

CarloSircana:


thofma:
Can you add the example as a test?

PR #246, opened 2021-03-30
"fix" factoring of polynomials over non-monic fields

fieker:


thofma:
Some assertions are throwing: https://github.com/thofma/Hecke.jl/pull/246/checks?check_run_id=2227642866

fieker:
On Tue, Mar 30, 2021 at 08:09:55AM -0700, thofma wrote:
> Some assertions are throwing: https://github.com/thofma/Hecke.jl/pull/246/checks?check_run_id=2227642866

I'll look into them, beu they are independent of this PR, the assertions
are in the absolute factorisation - and very random. I'd rather move
this one and deal with the abs. factoring independently

fieker:
Cool example. Problem is mathematical: a "quick" search shwoed that our (my) different function works iff the order is Gorenstein which here it is not. Solution is to not use the different as I am no longer sure
 - that I know "the" definition
 - have an algo that works in the general case...
I have a fix to the factorisation (well the gcd that is used in Trager)

PR #247, opened 2021-03-30
Remove Matrix interface for our matrices

### LINKED ISSUE #245: Hecke changes `collect(::fmpz_mat)` to return a Vector instead of a Matrix
If I just load AbstractAlgebra, I see this:
```
julia> M = MatrixSpace(ZZ, 2, 3); x = M(1:6); collect(x)
2×3 Matrix{BigInt}:
 1  2  3
 4  5  6
```
If I also load Nemo, I get this:
```
julia> M = MatrixSpace(ZZ, 2, 3); x = M(1:6); collect(x)
2×3 Matrix{fmpz}:
 1  2  3
 4  5  6
```
If I also load Hecke, I get this, which arguably is wrong:
```
julia> M = MatrixSpace(ZZ, 2, 3); x = M(1:6); collect(x)
6-element Vector{fmpz}:
 1
 4
 2
 5
 3
 6

```
This also happens if I force it to use AA.ZZ:
```
julia> M = MatrixSpace(AbstractAlgebra.ZZ, 2, 3); x = M(1:6); collect(x)
6-element Vector{BigInt}:
 1
 4
 2
 5
 3
 6
```

I noticed this because it breaks the doctests in the "new" Oscar.jl manual, which includes the AbstractAlgebra manual, which contains the above example.

thofma:
Fixes #245


thofma:
The downstream testing is useful. This is breaking Oscar, which I have to fix first.

PR #252, opened 2021-04-01
Non monic again

fieker:


thofma:
I found one error because of the following:
```julia
julia> Qx, x = PolynomialRing(FlintQQ, "x")
(Univariate Polynomial Ring in x over Rational Field, x)

julia> K, a = number_field(1//2*x^2 + 1, cached = false)
(Number field over Rational Field with defining polynomial 1//2*x^2 + 1, _a)

julia> Hecke.conjugate_quad(a)
ERROR: AssertionError: isone(k.pol_den)
Stacktrace:
 [1] conjugate_quad(a::nf_elem)
   @ Hecke ~/.julia/dev/Hecke/src/NumField/NfAbs/Elem.jl:1092
 [2] top-level scope
   @ REPL[12]:1
```

PR #257, opened 2021-04-06
add some more doc...

fieker:


thofma:
Hm, tests are failing.

fieker:
I think - and I veryfied with Carlo, that the problem was an error in field factory, in stable sugroups: the actio was applied to elements in the wrong group. I did not fail before as both groups seem to have (always, by accident) the same number of generators, so the vec*mat was not failing. Since I added a check to see that elements have to bein the domain, it errored.

PR #262, opened 2021-04-13
Move polynomial function to AbstractAlgebra's Misc directory.

wbhart:
This is the Hecke counterpart to https://github.com/Nemocas/AbstractAlgebra.jl/pull/840

I have no way to test this PR at the moment, but it is expected to fail until that AA PR and other PR's against Hecke are merged.

### DIFF
--- a/src/Misc/Poly.jl
+++ b/src/Misc/Poly.jl
@@ -1108,26 +1108,18 @@ function (f::acb_poly)(x::acb)
   return evaluate(f, x)
 end
 
-function polynomial(A::Array{T, 1}) where {T <: RingElem}
-  P = parent(A[1])
-  @assert all(x->parent(x) == P, A)
-  Pt, t = PolynomialRing(P, cached = false)
-  return Pt(A)
+################################################################################
+#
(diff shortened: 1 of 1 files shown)


fieker:
On Tue, Apr 13, 2021 at 02:26:00AM -0700, wbhart wrote:
> This is the Hecke counterpart to https://github.com/Nemocas/AbstractAlgebra.jl/pull/840
> 
> I have no way to test this PR at the moment, but it is expected to fail until this and other PR&#39;s against Hecke are merged.
Why? it should be that if there pull request with the same branch name
in AA/Nemo/Hecke they will be tested against each other - at least I
thought this to be the case?

wbhart:
I guess @thofma hasn't found time to merge in the other PR's against Hecke, so it currently doesn't pass against AbstractAlgebra master. At least, that's what seems to be happening on my machine with Hecke master. It overwrites AA methods, hangs in the tests and issues deprecation warnings, all of which led me to that conclusion.

No bother, I guess @thofma is quite busy these days. I'm sure he will get to it in time.

PR #269, opened 2021-04-20
Polynomial constructor similar zero

wbhart:
This is the Hecke counterpart to https://github.com/Nemocas/AbstractAlgebra.jl/pull/843 and https://github.com/Nemocas/Nemo.jl/pull/1046

It is expected to fail until those are merged. This PR should not be breaking.


wbhart:
Should we restart the Appveyor tests? I can't tell if they have just timed out or whether there is an actual hang there.

thofma:
I think it is normal that they time out :)

PR #272, opened 2021-04-21
maximal order bug fixed

CarloSircana:


thofma:
Test?

CarloSircana:
Lost it, it was a subfield of a given number field and the subfield function is non deterministic. Sorry.

PR #276, opened 2021-04-23
noDisc

fieker:
- remove printing
- do not export disc - it does not exist (anymore)


thofma:
I think this was already removed. Also the printing was removed in your other PR.

PR #277, opened 2021-04-23
Class group via norm relation

CarloSircana:


src/NfOrd/Unit/FindUnits.jl
```diff
@@ -259,6 +259,7 @@ function _unit_group_find_units(u::UnitGrpCtx, x::ClassGrpCtx; add_orbit::Bool =
   end
   @vprint :UnitGroup 1 "Finished processing\n"
   if has_full_rank(u)
+    u.units = reduce(u.units, u.tors_prec)
```
thofma, line 262: Eh, this is unrelated to the other changes?

CarloSircana: Oh my bad. I forgot to reverse it.

src/NumField/NfAbs/NormRelation/Clgp.jl
```diff
@@ -21,32 +15,68 @@ function get_sunits_from_subfield_data(OK, N; recursive::Bool = false, compact::
...
+        b = Hecke.saturate!(c, UZK, Int(p), 3.5, easy_root = degree(K) < 30)
+      end
+    end
+    idx = Hecke._validate_class_unit_group(c, UZK)[1]
   end
+  @vprint :NormRelation 1 "Index is $idx (should be 1)!\n"
   @assert idx == 1
   @vprint :NormRelation 1 "\n"
   c.finished = true
-  UZK.finished = true
```
thofma, line 49: Why one is finished and the other not?

CarloSircana: UZK contains the full unit group already after the first attempt, with the small factor base. The finished field is set to true there, so 20 lines above

thofma: Right, thanks

PR #287, opened 2021-05-06
homogenous -> homogeneous

fingolfin:
As suggested by @8d1h (thanks)

(Note: both spellings are allowed, but we should settle on one. Right now, the majority uses "homogeneous", hence I picked that. But we could also go the other way).


thofma:
Might have been the sage convention @simonbrandhorst :smile:

PR #292, opened 2021-05-15
Fix norm relations for non-normal fields

thofma:


thofma:
There is still an error. Don't merge.

PR #296, opened 2021-05-21
Abelian exts distinct

CarloSircana:


thofma:
Can you add the example where this makes a difference?

PR #297, opened 2021-05-25
Don't do parallel run on long test :/

thofma:


rfourquet:
I'm curious, what was the problem?

thofma:
See for example https://github.com/thofma/Hecke.jl/runs/2645588225?check_suite_focus=true#step:7:187

I don't fully understand it, but I think the work is being killed because of lack of ressources? At least this is my interpretation. All the other jobs run the tests in parallel but have no problem. The "long tests" job does more tests, which use quite a lot of memory.

PR #299, opened 2021-05-27
[WIP] K_\infty(x)

wbhart:
It's probably time to take a look at this code before I get too busy with test code and docs @fieker @thofma 

I shamelessly ripped off as much code from OrdLocalization.jl as possible. I'm not sure who has copyright of that file. Let me know if you want me to add someone's name.

The checked=false thing isn't that helpful here, and if I implemented what the interface calls for it would actually slow it down, since it is usually cheaper to check if the result is valid here before computing it (which is different to OrdLocalization). If the interface was changed to "**may** not be checked and an element of the rational function field not in the localization **may** be returned" then there would be no problem,

I wasn't sure of the difference between "in" (found in Base, exported) and "isin" (not). Perhaps you could advise me on that. If it is not meant to be exported, I can change it to "isin" or "icositsnotexported".

gcd(0, a) always returns 0 because the degree of 0 is -inf and is always lower than that of a. If you don't like this for whatever reason, let me know.

inv doesn't give the same error message as in OrdLocalization. It seemed a shame to waste the error message that you get for free from R(a, checked).


wbhart:
I added some tests, but there's a failure due to this:

```
julia> AbstractAlgebra.promote_rule == Hecke.promote_rule
false
```

Do you guys have your own promote_rule system? If so, there's no way the generic code in AbstractAlgebra can work with it.

fieker:
On Thu, May 27, 2021 at 12:37:18PM -0700, wbhart wrote:
> I added some tests, but there's a failure due to this:
> 
> ```
> julia> AbstractAlgebra.promote_rule == Hecke.promote_rule
> false
> ```
> 
> Do you guys have your own promote_rule system? If so, there's no way the generic code in AbstractAlgebra can work with it.
> 
We don't at least not as far as I know

src/Misc/DegreeLocalization.jl
```diff
@@ -0,0 +1,396 @@
...
+
+function in(a::Generic.Rat{T}, R::KInftyRing{T}) where T <: FieldElement
+   if parent(a) != function_field(R)
+      return false
+   end
+   return degree(numerator(a, false)) <= degree(denominator(a, false))
+end
+
+function deepcopy_internal(a::KInftyElem{T}, dict::IdDict) where T <: FieldElement
+   parent(a)(deepcopy(data(a)))
```
thofma, line 97: Probably should be passed `dict`?

thofma:
There was one bad promote rule in `Misc/OrdLocalization.jl`, which I am fixing in #300.

This is look good, thanks. Do you mind using 2 spaces? This is the convention, see https://github.com/thofma/Hecke.jl/blob/master/CONTRIBUTING.md.

wbhart:
Note dict is not passed in OrdLocalization and two spaces are only used for part of that file. I just copied what I could see as far as conventions.

wbhart:
However, I did use the maketype system for rand. I assume that is ok. If you want me to revert to a more traditional rand function, let me know.

I also used FieldElement instead of FieldElem, in case we ever want to move this elsewhere. Again let me know if this is not what you want.

There may be some other deviations I made from habit, but mainly I tried to look for examples in Hecke to follow.

src/Misc/DegreeLocalization.jl
wbhart: Are you sure? I'm having trouble finding any other examples of where we do so. I haven't the slightest idea what it does by the way.

wbhart:
Also, do you want this in the documentation? And if so, where? I could start a directory for function fields maybe?

I believe the next project is round two for function fields and then will be ideals I think, just to give you an idea where this is going.

thofma:
Thanks. I think we can wait with the documentation. Do you want to merge this first before moving on?

wbhart:
If you like.

thofma:
I agree that it would be a good idea to put this next into a dedicated `src/FunctionField` directory. Then the orders, ideals and round 2 can also go in there.

fieker:
On Fri, May 28, 2021 at 08:20:53AM -0700, wbhart wrote:
> Also, do you want this in the documentation? And if so, where? I could start a directory for function fields maybe?
> 
Id' say yes, please to both
> I believe the next project is round two for function fields and then will be ideals I think, just to give you an idea where this is going.
> 
Well.... you can also try to link against the integral closure in
Singular, the Round4, Montes (Bauch), ... but Round2 is easiest.

PR #316, opened 2021-06-14
Eval mod

fieker:
@thofma have a look, in Williams example, the time is dows to 4sec (from 86)


CarloSircana:
This is a good idea if the index of the equation order in the maximal order is large and suggests that the equation defining William's field is awful.

PR #329, opened 2021-07-05
Gcd local fields

CarloSircana:


thofma:
Why do we need `Nemo.precision`?

CarloSircana:
Because I use precision as a parameter in the function

PR #346, opened 2021-07-23
fix type inference problem

CarloSircana:


src/FieldFactory/relative_to_absolute.jl
```diff
@@ -29,6 +29,7 @@ function _relative_to_absoluteQQ(L::NfRelNS{nf_elem}, auts::Vector{NfRelNSToNfRe
 end
 
 function _relative_to_absolute(L::NfRelNS{nf_elem}, auts::Vector{NfRelNSToNfRelNSMor_nf_elem})
+  @vtime :Fields 4 maximal_order(L)
```
thofma, line 32: Is this on purpose added here?

CarloSircana: Well, I wanted to add it, even if not in this pull request.

thofma: OK. I was just checking that this was not a debugging artifact.

PR #367, opened 2021-08-03
Fix fields bug

CarloSircana:


test/FieldFactory/FieldFactory.jl
```diff
@@ -181,6 +181,19 @@
       d = fmpz(1252291517600545939502745293690906945712691746311040212121628914687318440182651069503694911322360563684969)
       flds = [Hecke.field_context(K)]
       @test isone(length(fields(72, 37, flds, d)))
+
+      f = x^24 + 12*x^23 + 12*x^22 - 372*x^21 - 1224*x^20 + 4192*x^19 + 22370*x^18 - 17518*x^17
+         - 198383*x^16 - 30608*x^15 + 1017900*x^14 + 582696*x^13 - 3257224*x^12 - 2385744*x^11 
+          + 6751956*x^10 + 4794056*x^9 - 9162405*x^8 - 4964804*x^7 + 7887120*x^6 + 2258008*x^5
```
thofma, line 187: Is this working? For me this is the wrong polynomial:
```julia
julia> function h(x)
         g = x
             + 2
         return g
       end
h (generic function with 1 method)

julia> h(x)
x
```

CarloSircana: You are right, fixed

PR #376, opened 2021-08-05
_minkowski_multiple: a fix, some tweaking, and tests

fingolfin:
`_minkowski_multiple(1)` is 2 not 1 (there is a 1x1 rational matrix
group of order 2). This also agrees with the formulas.

Call `bernoulli_cache` for a very minor potential performance boost.

Allow `_minkowski_multiple(QQ, n)` as alias for `_minkowski_multiple(n)`,
to make it easier to write code uniformly dealing with both number fields
and the rationals.

Remove unused `_multiple_of_finite_group_order_glnz`


thofma:
Thanks

PR #378, opened 2021-08-09
Fix can_solve_with_kernel

### LINKED ISSUE #377: can_solve_with_kernel returns incorrect results
In the ring case the final return value is essentially always incorrect, though I have managed to fix that in the version I am implementing for AbstractAlgebra.

However the field case frequently returns the incorrect kernel. Here is a specific example:

```
using Hecke
M = matrix(QQ, 2, 9, [37//19 4167//8 632//9 127//2 6268//95 -2487//13 -117//190 6793//380 -1231//3; 9//19 507//4 154//9 31//2 1526//95 -605//13 -29//190 1661//380 -100])
B = matrix(QQ, 2, 9, [36509513279//19710600 -333707923//556920 -12611630942//2463825 2052298517//3260400 135240137//889200 -468468844939//290990700 -235276429//125400 306806125153//28721160 416877067//518700; 17762102161//39421200 -192846298//1322685 -36830950333//29565900 1501627327//9781200 1828703//49400 -2595021407//6613425 -216977//475 2075621461//797810 76147888//389025])
flag, S, K = can_solve_with_kernel(M, B)
```

The resulting kernel is not correct.

Any hints as to why this might be would be helpful.

thofma:
Fixes #377.

Silly me, of course it works! I just copied the code from `nullspace` in AbstractAlgebra, since it does the identical thing.

I ran the tests from @wbhart's AA PR repeatedly and they passed.

### DIFF
--- a/src/Misc/Matrix.jl
+++ b/src/Misc/Matrix.jl
@@ -691,6 +691,9 @@ It returns a tuple $(n, M)$ where $M$ is a matrix whose rows generate
 the kernel of $a$ and $n$ is the rank of the kernel.
 """
 function left_kernel(x::fmpz_mat)
+  if nrows(x) == 0
+    return 0, zero(x, 0, 0)
+  end
   x1 = hnf(x')'
   H, U = hnf_with_transform(x1)
   i = 1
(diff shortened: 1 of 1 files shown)


wbhart:
Thanks. This is very helpful.

PR #380, opened 2021-08-10
Add example section to doc

thofma:


src/Map/NfOrd.jl
```diff
@@ -579,7 +579,12 @@ function extend(f::T, K::AnticNumberField) where T <: Union{NfOrdToFqNmodMor, Nf
     if l == 0
       return f(O(m*x, false))//f(O(m))
     else
-      return f(O(pia^l * m * x, false))//f(O(pia^l * m, false))
+      w = pia^l * m * x
+      fl, z = _check_elem_in_order(w, O)
+      if !fl
+        error("Element not p-integral")
+      end
+      return f(O(w, z, false))//f(O(pia^l * m, false))
```
thofma, line 587: @CarloSircana Does this make sense? `w in O` should be equivalent to being `p`-integral, right?

CarloSircana: I think it's not optimal. Instead of using _check_elem_in_order, we should just ask for the valuation. On the other hand, I would probably check the valuation of x immediately.

PR #388, opened 2021-08-30
Properly handle units in class_group_add_relation

thofma:


thofma:
@fieker Something is strange. I am trying to handle the case where we find units as relations, but it is hanging. In the following line
https://github.com/thofma/Hecke.jl/blob/232fb1228d54b8926a0ced207b02babde520a506/src/NumFieldOrd/NfOrd/Clgp/Rel_add.jl#L44
is `n` supposed to be the norm of `a`? I thought so, but putting in `@assert n == norm(a)` shows that this is not always true, for example for `f = x^8 - x^7 + x^5 - x^4 + x^3 - x + 1`.

PR #397, opened 2021-09-09
Add optional type argument to (free_)abelian_group

fingolfin:
... to make it possible to have different abelian group implementations available in Oscar.jl.

This is incomplete, and in particular has no tests, but I've had it sitting around for so long, I thought I should turn it into a PR so that either I feel pressured into finishing it, or someone else perhaps can help out :-) 

CC @thofma @ThomasBreuer

### DIFF
--- a/src/GrpAb/GrpAbFinGen.jl
+++ b/src/GrpAb/GrpAbFinGen.jl
@@ -62,12 +62,13 @@ isabelian(::GrpAbFinGen) = true
 ##############################################################################
 
 @doc Markdown.doc"""
-    abelian_group(M::fmpz_mat) -> GrpAbFinGen
+    abelian_group(::Type{T} = GrpAbFinGen, M::fmpz_mat) -> GrpAbFinGen
 
 Creates the abelian group with relation matrix `M`. That is, the group will
 have `ncols(M)` generators and each row of `M` describes one relation.
 """
-function abelian_group(M::fmpz_mat; name::String = "")
+abelian_group(M::fmpz_mat; name::String = "") = abelian_group(GrpAbFinGen, M, name=name)
+function abelian_group(::Type{GrpAbFinGen}, M::fmpz_mat; name::String = "")
    if issnf(M) && nrows(M) > 0  && ncols(M) > 0 && !isone(M[1, 1])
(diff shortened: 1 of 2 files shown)


src/GrpAb/GrpAbFinGen.jl
```diff
@@ -77,38 +78,21 @@ function abelian_group(M::fmpz_mat; name::String = "")
...
-  if name != ""
-    set_name!(G, name)
-  end
+  name == "" || set_name!(G, name)
   return G
 end
 
 @doc Markdown.doc"""
-    abelian_group(M::Matrix{fmpz}) -> GrpAbFinGen
+    abelian_group(M::Matrix{T}) where T <: Union{Integer, fmpz} -> GrpAbFinGen
```
fingolfin, line 86: OK, I just realized I got this wrong, it should be this (of course also elsewhere):
```suggestion
    abelian_group(M::Matrix{<:T}) where T <: Union{Integer, fmpz} -> GrpAbFinGen
```
or even use `<:AbstractMatrix{<:T}` 

Also, perhaps writing it like this would be more userfriendly?
```suggestion
    abelian_group(M::Matrix{fmpz}) -> GrpAbFinGen
    abelian_group(M::Matrix{<:Integer}) -> GrpAbFinGen
```
But then again, if we introduce `Nemo.IntegerUnion`, ew could just use that and write
```suggestion
    abelian_group(M::<AbstractMatrix{<:IntegerUnion}) -> GrpAbFinGen
```

thofma: I believe the docstrings are allowed to be sloppy so I would also allow
`Matrix{<:Union{Integer, Union}}` or `AbstractMatrix{<:Union{Integer, Union}}`. Not sure if this shared by everyone.

fingolfin: Err, yeah, I messed up some of these, but I hope the gist is clear ;-)

fingolfin: I've updated it

thofma: The first is also not wrong, since `T <: Union{S, ...}` also allows all `T` with `T <: S`.

PR #399, opened 2021-09-10
Trivial

fieker:


examples/Round2.jl
```diff
@@ -76,6 +76,7 @@ mutable struct Order <: AbstractAlgebra.Ring
       r.trans = T
       r.itrans = Ti
     end
+    @show map(representation_matrix, basis(r))
```
thofma, line 79:
```suggestion
```
You should remove this change locally.

fieker:
On Fri, Sep 10, 2021 at 02:17:01AM -0700, Tommy Hofmann wrote:
> 
> 
> 
> > @@ -76,6 +76,7 @@ mutable struct Order <: AbstractAlgebra.Ring
>        r.trans = T
>        r.itrans = Ti
>      end
> +    @show map(representation_matrix, basis(r))
> 
> ```suggestion
> ```
> You should remove this change locally.
indeed - but it only in the GenericcRound2, so should be harmless.
Or is this actually used? Which would be cool

PR #407, opened 2021-09-19
Upadate_ResidueField_UnramifiedLocalField

aslam7861:
With this function, automorphism_group(L::UnramifiedLocalField) also works

### DIFF
--- a/src/LocalField/LocalField.jl
+++ b/src/LocalField/LocalField.jl
@@ -439,11 +439,8 @@ function ResidueField(K::LocalField{ S, UnramifiedLocalField}) where {S <: Field
    kk = FiniteField(f)[1]
    bas = basis(K)
    u = gen(kk)
-   function proj(a:: Hecke.LocalFieldElem)
-      if valuation(a) < 0
-         error("The projection is not well defined!")
-      end
-      col = typeof(kk(1))[]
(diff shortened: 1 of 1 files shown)


thofma:
Thanks. Could you add an example which was not working before here: https://github.com/thofma/Hecke.jl/blob/master/test/LocalField/LocalField.jl?

aslam7861:
> Thanks. Could you add an example which was not working before here: https://github.com/thofma/Hecke.jl/blob/master/test/LocalField/LocalField.jl?


You're welcome!
Here is the example which was not working before:
 Qx,x = FlintQQ["x"]
 f = Qx([ 1, 8, -40, -46, 110, 71, -113, -43, 54, 11, -12, -1, 1 ])
 L = number_field(f)[1];
 P = prime_decomposition(maximal_order(L),7)[1][1];
 lp, mp = Hecke.generic_completion(L,P);
 Qy,y = PolynomialRing(lp,"y")
 f, mf = ResidueField(lp)
 N = Hecke.unramified_extension(y^3+mf\(gen(f)) +4 )[1]
 ResidueField(N) # "this was not working before"
automorphism_group(N) #"This was also not working"

thofma:
Good. Can you please add it to the file `https://github.com/thofma/Hecke.jl/blob/master/test/LocalField/LocalField.jl`?

PR #411, opened 2021-09-23
neq.jl

aslam7861:
norm equations over finite and local fields extensions:

[3] L = number_field(x^6-6*x^4+9*x^2+23)[1]
[4] P = prime_decomposition(maximal_order(L),23)[1][1]
[5] lp,mp = Hecke.generic_completion(L,P)
[6] a = uniformizer(lp)
[7] Qy,y = PolynomialRing(lp,"y")
[8] N = Hecke.unramified_extension(y^2+y+1)[1]
[9] #include("neq.jl")
[10] b=21*(1+uniformizer(lp))
[11] norm_equation(N,(b))
[12] norm_equation_unramified(N,b)
[13] norm_det(ans)*b^-1-1 ##norm_det is a norm function using matrix representation
[14] f,mf = ResidueField(lp);
[15] F,mF = ResidueField(N);
[16] norm_equation(F,f(21))
[17] norm(ans) == f(21)
[18] Hecke.percent_P()
[19] k= base_field(lp)
[20] b=(1+uniformizer(k)^3)
[21] norm_equation(lp,b)
[22] norm_det(ans)*b^-1-1
[23] Hecke.percent_P()

### DIFF
--- a/src/LocalField/neq.jl
+++ b/src/LocalField/neq.jl
@@ -1,0 +1,216 @@
+#export: degree_relative, random_elem, one_root, norm_equation  
+
+
+degree_relative(L::Hecke.LocalField,K::Hecke.LocalField) = Int(absolute_degree(L)//absolute_degree(K)
+)
+
+ degree_relative(L::Hecke.RelFinField,K:: Hecke.RelFinField)  = Int(absolute_degree(L)//absolute_degree(K))
+degree_relative(L::Hecke.RelFinField,K::FqNmodFiniteField)  = Int(absolute_degree(L)//absolute_degree(K))
+
+degree_relative(L::FqNmodFiniteField, K::Hecke.Nemo.GaloisField)  = Int(absolute_degree(L)//absolute_degree(K))
+
+
+##############################################
+#random element with small coefficients
+##############################################
+
+function random_elem(L::Hecke.LocalField)
+     b = basis(L)
+     n = degree(L)
+     r = [rand(1:5*n) for i in 1:n]
+return sum( [ r[i]*b[i] for i in 1:n])
+end
+
+
+function random_elem(L::FlintQadicField)
+     b = basis_qadic(L)
+     n = degree(L)
+     r = [rand(1:5*n) for i in 1:n]
+return sum( [ r[i]*b[i] for i in 1:n])
+end
+
+
+
+
+
+function one_root( f::fq_nmod_poly, F:: Hecke.RelFinField)
+   g = polynomial(F, [coeff(f,i) for i = 0:degree(f) ] )
+   fac = factor(g)
+   if length(fac)  == 1
+      return []
+   end
+   r = [x for x in fac ][1][1]
+   @assert degree(r) == 1
+return -coeff(r,0)
+end
+
+
+function one_root( f::gfp_poly, F:: FqNmodFiniteField)
+   g = polynomial(F, [coeff(f,i) for i = 0:degree(f) ] )
+   fac = factor(g)
+   if length(fac)  == 1
+      return []
+   end
+   r = [x for x in fac ][1][1]
+   @assert degree(r) == 1
+return -coeff(r,0)
+end
+
+
+
+function one_root( f::Hecke.AbstractAlgebra.Generic.Poly, F:: Hecke.RelFinField)
+   g = polynomial(F, [coeff(f,i) for i = 0:degree(f) ] )
+   fac = factor(g)
+   if length(fac)  == 1
+      return []
+   end
+   r = [x for x in fac ][1][1]
+   @assert degree(r) == 1
+return -coeff(r,0)
+end
+
+function one_root( f::gfp_poly, F:: Hecke.RelFinField)
+   g = polynomial(F, [coeff(f,i) for i = 0:degree(f) ] )
+   fac = factor(g)
+   if length(fac)  == 1
+      return []
+   end
+   r = [x for x in fac ][1][1]
+   @assert degree(r) == 1
+return -coeff(r,0)
+end
+
+
+#########################################################################
+# Alternative way of computing the norm due to some problem in resultant
+#
+##########################################################################
+
+function norm_det(a)
+   L = parent(a)
+   m = matrix([[coeff(a*x, i-1) for i in 1:degree(L)] for x in basis(L)])
+return det(m)
+end
+
+
+######################### norm equation over finite fields ##############
+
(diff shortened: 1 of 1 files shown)


src/LocalField/neq.jl
```diff
@@ -0,0 +1,216 @@
...
+
+##############################################
+#random element with small coefficients
+##############################################
+
+function random_elem(L::Hecke.LocalField)
+     b = basis(L)
+     n = degree(L)
+     r = [rand(1:5*n) for i in 1:n]
+return sum( [ r[i]*b[i] for i in 1:n])
```
fingolfin, line 21: This code is unevenly indented, fix that? I.e. this line is not indented at all, similarly elsewhere. Also, the lines before seem to use 5 spaces for indentation, which is highly unusual (I've seen 2, 3 and 4 used in various Oscar projects and elsewhere, but I think never ever 5, in the past 20 years :-)

src/LocalField/neq.jl
```diff
@@ -0,0 +1,216 @@
...
+
+
+
+
+
+function one_root( f::fq_nmod_poly, F:: Hecke.RelFinField)
+   g = polynomial(F, [coeff(f,i) for i = 0:degree(f) ] )
+   fac = factor(g)
+   if length(fac)  == 1
+      return []
```
fingolfin, line 40: This is not type stable, is it?

src/LocalField/neq.jl
```diff
@@ -0,0 +1,216 @@
...
+degree_relative(L::Hecke.RelFinField,K::FqNmodFiniteField)  = Int(absolute_degree(L)//absolute_degree(K))
+
+degree_relative(L::FqNmodFiniteField, K::Hecke.Nemo.GaloisField)  = Int(absolute_degree(L)//absolute_degree(K))
+
+
+##############################################
+#random element with small coefficients
+##############################################
+
+function random_elem(L::Hecke.LocalField)
```
fingolfin, line 17: Wouldn't it make more sense to turn this into a `rand` method?

src/LocalField/neq.jl
```diff
@@ -0,0 +1,216 @@
...
+     b = basis(L)
+     n = degree(L)
+     r = [rand(1:5*n) for i in 1:n]
```
fingolfin, line 20: Where is the "5' coming from? Seem like this is something arbitrary?

src/LocalField/neq.jl
```diff
@@ -0,0 +1,216 @@
...
+
+
+
+function one_root( f::fq_nmod_poly, F:: Hecke.RelFinField)
+   g = polynomial(F, [coeff(f,i) for i = 0:degree(f) ] )
+   fac = factor(g)
+   if length(fac)  == 1
+      return []
+   end
+   r = [x for x in fac ][1][1]
```
fingolfin, line 42: Huh? Perhaps like this?
```suggestion
   r = first(fac)[1]
```

src/LocalField/neq.jl
```diff
@@ -0,0 +1,216 @@
...
+     n = degree(L)
+     r = [rand(1:5*n) for i in 1:n]
+return sum( [ r[i]*b[i] for i in 1:n])
+end
+
+
+
+
+
+function one_root( f::fq_nmod_poly, F:: Hecke.RelFinField)
```
fingolfin, line 36:
```suggestion
function one_root(f::fq_nmod_poly, F::Hecke.RelFinField)
```

src/LocalField/neq.jl
```diff
@@ -0,0 +1,216 @@
...
+
+function one_root( f::gfp_poly, F:: FqNmodFiniteField)
+   g = polynomial(F, [coeff(f,i) for i = 0:degree(f) ] )
+   fac = factor(g)
+   if length(fac)  == 1
+      return []
+   end
+   r = [x for x in fac ][1][1]
+   @assert degree(r) == 1
+return -coeff(r,0)
```
fingolfin, line 56: Isn't this exactly the same code as in the method and before and after? Only the type of the arguments differs? Maybe write the function just once and use ` Union` type  for each of the two arguments?

src/LocalField/neq.jl
```diff
@@ -0,0 +1,216 @@
...
+    f = polynomial(k,vcat([b],[rand(k) for i = 1:n-1],[1] ) )
+    while !isirreducible(f)
+       f = polynomial(k,vcat([b],[rand(k) for i = 1:n-1],[1] ) )
+    end
+return one_root(f,F)
+end
+
+
+
+function norm_equation(F::Hecke.RelFinField, b::fq_nmod)
```
fingolfin, line 124: Again: aren't you duplicating the same code, just with different types for the arguments? If so, once again: merge them and use a union type for the arguments?

src/LocalField/neq.jl
```diff
@@ -0,0 +1,216 @@
...
+    k = parent(b)
+    n = degree_relative(F,k)
+    Qx,x = PolynomialRing(k,"x")
+    Qz,z = PolynomialRing(F,"z")
```
fingolfin, line 128: These two polynomial rings are not used, are they?
```suggestion
```

src/LocalField/neq.jl
```diff
@@ -0,0 +1,216 @@
...
+
+function norm_equation(R:: Hecke.LocalField, b::Union{qadic,Hecke.LocalFieldElem})
+   K = parent(b)
+   prec_b = precision(b)
+   @assert base_field(R) == K  #"since trace(a,_)" is not defiend"
+   f,mf = ResidueField(K);
+   F,mF = ResidueField(R);
+   ee = absolute_ramification_index(K)
+   if degree(R) == ramification_index(R) && mf(b) !=f(1)
+     return "To be implemented"
```
fingolfin, line 147:
```suggestion
     error("To be implemented")
```

fingolfin (changes requested):
There should be some basic tests for this, and some comments that explain what the functions do (even better would be proper docstrings, of course)

src/LocalField/neq.jl
aslam7861: Thanks for your nice suggestion.s I will change all  accordingly

src/LocalField/neq.jl
aslam7861: The field is with infinite elements. So I wanted to construct a random element with small coefficients. One can change 5 with any other number

thofma:
Thanks @aslam7861

PR #425, opened 2021-10-13
use setprecision without !

fieker:
needs eventually a serious re-thinking: padics do not support lowering
of precision inplace. Maybe that needs to change? Increasing precision
is fine


thofma:
After the merge, there is no change here. Did they end up in another PR that was already merged?

thofma:
@fieker?

fieker:
I think this was also part of other changes, so I am happy to close this. The underlying isse is still there - it is a matter of efficiency...

PR #440, opened 2021-11-05
degree 1 and 2 fields are difficult...

fieker:


thofma:
Which bug does this fix? Do you have the example?

fieker:
Tricky, an example would be
```K, a = cyclotomic_field(3)
L, b = cyclotomic_field(3)
Hecke.force_coerce_cyclo(L, a)
```
This possibly loops forever as `a.elem_length` is undefined, or crash or so. I can put a test in (trivial), but without the fix, the test may never terminate

fieker:
even more difficult. Aparently, the element needs to be constructed in a bad way.

thofma:
The don't bother adding the test.

fieker:
On Mon, Nov 08, 2021 at 12:34:52AM -0800, Tommy Hofmann wrote:
> The don't bother adding the test.

Now I am curious as to why and where this element is coming from.
But there is a vague chance it was memory corruption elsewhere.

thofma:
Merge at your own discretion.

PR #445, opened 2021-11-09
Fix norm for relative finite fields

thofma:
Apparently the order must be swapped? Can you confirm @fieker?
```Julia
julia> f, o = FiniteField(7, 2, "o");

julia> fx, x = f["x"];

julia> F, u = FiniteField(x^3 + o + 4, "u");

julia> c = 3*u + 5*o + 1;

julia> norm(c)
x.data = 3*x + 5*o + 1
defining_polynomial(parent(x)) = x^3 + o + 4
resultant(x.data, defining_polynomial(parent(x))) = 5*o
5*o

julia> c^(divexact(7^6 - 1, 7^2 - 1))
2*o
```


fieker:
Yep, it needs to be swapped

PR #448, opened 2021-11-11
improve direct_prod by not using hnf

fieker:


fieker:
just to explain: for any abelian group A the computation of the direct product of 100 copies of A takes time. A lot of time. Internally we need an hnf of the relations to proceed. This of course is trivial from the hnf of the factors, but as the matrix in total can be quite large, our hnf code collapsed...

PR #451, opened 2021-11-17
Make matrix(::Vector{Vector{}}) return the correct matrix

thofma:
Before:
```julia
julia> matrix(F, [[0, 0], [1, 0]])
[0   0]
[1   0]

julia> matrix([[F(0), F(0)], [F(1), F(0)]])
[0   1]
[0   0]
```
@fieker: This was only used in the conjugate code for local fields and some factorization code. I believe the GModule code in Oscar is also using it wrongly (this is how I found it).

### DIFF
--- a/src/LocalField/Conjugates.jl
+++ b/src/LocalField/Conjugates.jl
@@ -340,7 +340,7 @@ function special_gram(m::Vector{Vector{qadic}})
 end
 
 function special_gram(m::Vector{Vector{padic}})
-  n = matrix(m)
+  n = transpose(matrix(m))
   n = n'*n
   return [[n[i,j] for j=1:ncols(n)] for i = 1:nrows(n)]
 end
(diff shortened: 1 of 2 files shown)


thofma:
Probably used more often in Oscar ...

fieker:
On Wed, Nov 17, 2021 at 12:31:47PM -0800, Tommy Hofmann wrote:
> Before:
> ```julia
> julia&gt; matrix(F, [[0, 0], [1, 0]])
> [0   0]
> [1   0]
> 
> julia&gt; matrix([[F(0), F(0)], [F(1), F(0)]])
> [0   1]
> [0   0]
> ```
> @fieker: This was only used in the conjugate code for local fields, and there only at the end when computing the regulator as a determinant. So this still works. But I believe the GModule code in Oscar is also using it wrongly (this is how I found it).
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/thofma/Hecke.jl/pull/451
This is bad - however, what is the correct result?

PR #452, opened 2021-11-18
try to fix Gunter's problem

fieker:


thofma:
Added a comment and a test. Will merge as soon as everything is green.

PR #455, opened 2021-11-18
deal with trivial cases

fieker:


src/NumField/NfAbs/Simplify.jl
```diff
@@ -345,7 +345,7 @@ function _find_prime(v::Vector{fmpz_poly})
       d1 = lcm(Int[x for (x, v) in FS])
       d = lcm(d1, d)
     end
-    if d < total_deg^2
+    if d <= total_deg^2
```
thofma, line 348: Do you have an example where the code fails because of this? Or the result is suboptimal?

fieker:
On Thu, Nov 18, 2021 at 05:24:15AM -0800, Tommy Hofmann wrote:
> 
> 
> 
> > @@ -345,7 +345,7 @@ function _find_prime(v::Vector{fmpz_poly})
>        d1 = lcm(Int[x for (x, v) in FS])
>        d = lcm(d1, d)
>      end
> -    if d < total_deg^2
> +    if d <= total_deg^2
> 
> Do you have an example where the code fails because of this? Or the result is suboptimal?
for a field of total_deg 1 < can never be achieved. This is now also
blocked up the chain

PR #457, opened 2021-11-18
don't put zeros in sparse rows

fieker:


thofma:
It seems that the module code in Oscar relies on this ... https://github.com/thofma/Hecke.jl/runs/4249967799?check_suite_focus=true

thofma:
I am inclined to store the base ring in the row. What do you think?

fieker:
On Thu, Nov 18, 2021 at 05:17:37AM -0800, Tommy Hofmann wrote:
> I am inclined to store the base ring in the row. What do you think?
No string feelings either way. The rows were originally never intended
to be used externally, but now they are...

PR #459, opened 2021-11-18
Update log function

aslam7861:
needs to be removed since log(1)=0. example->
K = QadicField(7,4,10)[1];
Qx, x = PolynomialRing(K);
L = Hecke.eisenstein_extension(x^120-7)[1];
log(1+uniformizer(L)^25)  ##it should not be zero


aslam7861:
it needs more requirements.....

PR #465, opened 2021-11-24
RFS: systematically have GF return only the field

fieker:
and FiniteField always try to return the generator as well


fingolfin (approved):
I like it, as it goes a step towards resolving https://github.com/oscar-system/Oscar.jl/issues/590

PR #473, opened 2021-12-02
attempt to fix `force_coerce_cyclo`

ThomasBreuer:
- Deal with the case that either going up or down does not suffice.
  (For example, write a 3rd root of unity in the 9th cyclotomic field as an element of the 15th cyclotomic field.)

  Most of the old code is still there, it has just been rearranged because we have to coerce first down and then up in the general case.

- Added tests for the situations that can occur.

- Removed code for the special case `isodd(fa) && (2*fa) % fb == 0` because it was wrong in some cases (see the extended tests).

- Return `nothing` if no coercion is possible and `throw_error` is not `Val{true}`; up to now, `false` was returned in one case.

### DIFF
--- a/src/NumField/NfAbs/NfAbs.jl
+++ b/src/NumField/NfAbs/NfAbs.jl
@@ -1289,67 +1289,47 @@ function force_coerce_cyclo(a::AnticNumberField, b::nf_elem, throw_error::Type{V
   else
     b_length = b.elem_length
   end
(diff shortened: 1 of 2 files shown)


fieker (approved):

ThomasBreuer:
@thofma Thanks for merging, but this was perhaps a bit too early. (The coverage report came afterwards, and it would have been good to adjust the pull request to it.)

thofma:
Oops, sorry.

PR #480, opened 2021-12-09
Automorphism group generators of binary quadratic forms

simonbrandhorst:
Implements automorphism group generators of binary quadratic forms of square discriminant and removes an erroneous assert

### DIFF
--- a/test/QuadForm/QuadBin.jl
+++ b/test/QuadForm/QuadBin.jl
@@ -205,6 +205,15 @@
     gens = automorphism_group_generators(g)
     @test gens == [ZZ[-1 0;0 -1]]
 
+    g = binary_quadratic_form(1, 2, 0)
+    gens = automorphism_group_generators(g)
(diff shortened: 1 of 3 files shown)


test/QuadForm/QuadBin.jl
```diff
@@ -223,6 +236,8 @@
     @test isequivalent(f, g, proper = false)
     @test isequivalent(2 * f, 2 * f)
     @test !isequivalent(2 * f, 3 * f)
+    h = binary_quadratic_form(-2, 1, 0)
+    @test isreduced(Hecke.reduction(q))
```
thofma, line 240:
```suggestion
    @test isreduced(Hecke.reduction(h))
```

thofma:
Do you have examples where this failed? It seems that the branches you introduced are not hit during the tests. According to the annotation here: https://github.com/thofma/Hecke.jl/pull/480/files. But maybe the annotation is wrong?

PR #482, opened 2021-12-09
add a "native" type for matrices over number fields

fieker:
more a proof-of-concept, the idea is to avoid the GC
by hiding the entries in the matrix and having a finalizer for
the matrix only.

rref! and ref! is working and competitive
view, hcat, vcat: OK
conversion to and from the classical type: OK

missing:
 - native add, mul, ...
 - det

all(?) generic methods should work - but without any benefit

### DIFF
--- a/examples/NfMat.jl
+++ b/examples/NfMat.jl
@@ -1,0 +1,334 @@
+module NfMatModule
+
+using Hecke
+import Nemo
+
+struct nf_elem_raw
+  a::Clong
+  b::Clong
+  c::Clong
+  d::Clong
+  function nf_elem_raw()
+    new(0,1,0,0)  #the den is in pos 2 for deg == 1 or > 2
+                  #and in pos 4 for deg == 2
+  end
+  function nf_elem_raw(::Int)
+    new(0,0,0,1)  #the den is in pos 2 for deg == 1 or > 2
+                  #and in pos 4 for deg == 2
+  end
+end
+
+function Base.zero(::Type{nf_elem_raw})
+  return nf_elem_raw()
+end
+
+function Base.one(::Type{nf_elem_raw})
+  return nf_elem_raw(1)
+end
+
+mutable struct NfMatElem <: MatElem{nf_elem}
+  entries:: Vector{nf_elem_raw}
+  rows:: Vector{Int}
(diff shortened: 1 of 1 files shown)


examples/NfMat.jl
```diff
@@ -0,0 +1,334 @@
+module NfMatModule
+
+using Hecke
+import Nemo
+
+struct nf_elem_raw
+  a::Clong
+  b::Clong
+  c::Clong
+  d::Clong
```
tthsqe12, line 10: These definitely should be Int because Clong is not correct on windows.

tthsqe12: @thofma

fieker:
On Fri, Dec 10, 2021 at 06:51:34AM -0800, dan wrote:
> @tthsqe12 commented on this pull request.
> 
> 
> 
> > +  a::Clong
> +  b::Clong
> +  c::Clong
> +  d::Clong
> 
> These definitely should be Int because Clong is not correct on windows.
> 
I thought long is used in flint? and Int is routinely 32 bits only? But
be my guest: this needs to match in size the antic type

thofma:
What `ulong/slong` in flint is, matches with `Int/UInt` in julia.

fieker:
Do support windows? Oscar does not....

thofma:
Nemo and Hecke do, although I disabled the tests for Windows on Hecke a while ago because they were always starving on memory.

fieker:
On Sat, Dec 11, 2021 at 06:40:46AM -0800, Tommy Hofmann wrote:
> Nemo and Hecke do, although I disabled the tests for Windows on Hecke a while ago because they were always starving on memory.
> 
OK, Nemo is doing
  Ptr{Nothing}
  Int
  Int
  Int

C is doing a union of
  fmpz_t
  fmpz_t
and
  fmpz [3]
  fmpz_t
and
  fmpq_poly_elem_t
which is
  fmpz *
  fmpz_t
  slong
  slong

I figure on all reasonable architectures this is 4*8 bytes, including
Windows in non-32-bit mode. Are we supporting 32-bit as well?

tthsqe12:
4 Ints will have the right size on all currently supported archs.

PR #484, opened 2021-12-11
Reenable Windows CI

thofma:


thofma:
@tthsqe12 Any idea what to do about those errors on Windows?

tthsqe12:
jl_malloc returned null, so we are using too much memoty here, or something got corrupted.

PR #499, opened 2021-12-22
fix _solve_X_ker

simonbrandhorst:


thofma:
Do we have an example where the previous version failed?

simonbrandhorst:
you can call it with any input and it fails. Since it asks for an undefined variable `ker`. Though I do not have a good example right now. Translating some more code from sage. But that is in oscar since it calls some orthogonal groups.

simonbrandhorst:
and actually I forgot what the stuff is doing ;-). So this is kind of a no-brainer commit.

PR #510, opened 2022-01-04
Normal form for degenerate TorQuadMod

simonbrandhorst:


thofma:
Seems that the long tests are failing: https://github.com/thofma/Hecke.jl/runs/4704391381?check_suite_focus=true

You can do `Hecke.test_module("...", long = true)` to enable them.

PR #517, opened 2022-01-08
Fix isisometric_with_isometry

simonbrandhorst:


thofma:
I pushe some fixes for the isotropic vector computation. This is now working I believe and I added an example exhibiting the original bug. But it is not feasible to test it for random spaces of rank >= 3. It is just too slow.

I also fix some other bad things. The tests should be failing with
```
  Test threw exception
  Expression: (gg_deg + g) - gg_deg == g
  ArgumentError: the second form must be regular to apply Witt cancellation
  Stacktrace:
```
I didn't even know what we had `-` for (local) genera :)

Do you have an idea how to fix it?

simonbrandhorst:
Thank you. 
Remove the test. It is nonsense, since the (-) function is based on Witt cancellation which requires a regular form.

PR #527, opened 2022-01-13
Convert some uses of acccessors to use attributes instead

fingolfin:


thofma:
Thanks! Will merge once green.

PR #536, opened 2022-01-15
Answer to Issue #533 + fixes of some bug

StevellM:
I suggest an answer to #533 and also some fixes. I wanted to put into 2 different pull requests but I did a mess with my branches.


src/QuadForm/Quad/ZLattices.jl
```diff
@@ -491,16 +493,52 @@ function _root_lattice_A(n::Int)
   return z
 end
 
+function _root_lattice_D(n::Int)
+  n < 2 && error("Parameter ($n) for root lattices of type :D must be greater of equal to 2")
```
simonbrandhorst, line 497:
```suggestion
  n < 2 && error("Parameter ($n) for root lattices of type :D must be greater or equal to 2")
```

src/QuadForm/Quad/ZLattices.jl
```diff
@@ -491,16 +493,52 @@ function _root_lattice_A(n::Int)
...
+  if n == 2
+    G = matrix(ZZ, [2 0 ;0 2])
+  elseif n == 3
+    return _root_lattice_A(n)
+  else
+    G = zero_matrix(ZZ, n, n)
+    G[1,3] = G[3,1] = -1
+    for i in 1:n
+      G[i,i] = 2
+      if i in 2:n-1
```
simonbrandhorst, line 507:
```suggestion
      if 2 <= i <= n-1
```

test/QuadForm/Quad/ZLattices.jl
```diff
@@ -347,6 +341,39 @@ end
   @test ambient_space(N) === ambient_space(L)
   @test rank(N) == 0
   @test basis_matrix(invariant_lattice(L, identity_matrix(QQ, 2))) == basis_matrix(L)
+  
+  randlist = rand(2:70,10)
```
simonbrandhorst, line 345:
```suggestion
  randlist = rand(2:20,10)
```
There is no need to test so many or big lattices. If it works for `2:10` it works.

test/QuadForm/Quad/ZLattices.jl
```diff
@@ -347,6 +341,39 @@ end
...
+  L = [root_lattice(:D,i) for i in randlist]
+  @test any(l -> discriminant(l) == 4, L)
+  @test any(iseven, L)
+  @test any(l -> norm(l) == 2, L)
+
+  L = root_lattice(:D,3)
+  #Problem here: equality of lattices depends on 'strict' equality of their ambient
```
simonbrandhorst, line 352: It is intended that only lattices in the same ambient space compare. 
Otherwise `isisometric` can be used.

simonbrandhorst (changes requested):

simonbrandhorst:
@thofma Please confirm the changes on the associative algebra part.

src/AlgAssRelOrd/Ideal.jl
```diff
@@ -1432,11 +1432,13 @@ end
 # factor.
 # Always considers I as an ideal of its left order.
 function factor(I::AlgAssRelOrdIdl)
+  @assert !iszero(I)
   O = left_order(I)
   @assert ismaximal(O)
+  basis_pmatrix(I) != basis_pmatrix(1*O) || error("I must be proper")
```
thofma, line 1438:
```suggestion
  I != 1*O || error("I must be proper")
```

src/QuadForm/Quad/ZLattices.jl
StevellM: Now that we are moving equality of lattices, should I keep the definition of `D3` like this, being equal to `A3` or should I make standard like `D4`,`D5`, and so on ? In this case `D3` and `A3` will be isometric but not equal.

simonbrandhorst: It doesn't really matter I think. So you can keep it. There is also an elegant description of the `A,D,E` root lattices as a sublattice of the standard euclidean space of dimension `n+1`. But lets keep the ambient keyword for later :-).

PR #540, opened 2022-01-17
Fix and use some rank 2 code

thofma:


thofma:
Are some of your examples degenerated, that is, zero discriminant?

simonbrandhorst:
No, all non-degenerate.

simonbrandhorst:
perhaps one of square determinant would be nice. They are all non-squares.

thofma:
Yes, this is the case that was failing before. I will add one.

simonbrandhorst:
There you go. Added some.

simonbrandhorst:
square determinant and degenerate

thofma:
Nice, thanks.

PR #541, opened 2022-01-17
Fix dual and add lattice equality

thofma:
For lattice equality I am leaving a lot on the table performance wise, so please complain if it is too slow :)

### DIFF
--- a/src/QuadForm/Herm/Lattices.jl
+++ b/src/QuadForm/Herm/Lattices.jl
@@ -220,8 +220,7 @@ end
 ################################################################################
 
 function lattice_in_same_ambient_space(L::HermLat, m::T) where T
-  return hermitian_lattice(base_field(L), m,
-                           gram_ambient_space = gram_matrix(ambient_space(L)))
+  return lattice(ambient_space(L), m)
 end
 
 ################################################################################
@@ -321,14 +320,15 @@ end
 ################################################################################
 
 function dual(L::HermLat)
-  G, B = _gram_schmidt(gram_matrix_of_rational_span(L), involution(L))
+  G = gram_matrix_of_rational_span(L)
+  @req rank(G) == nrows(G) "Lattice must be non-degenerate"
+  B = matrix(pseudo_matrix(L))
   C = coefficient_ideals(L)
   Gi = inv(G)
   new_bmat = Gi * B
-  new_coeff =  eltype(C)[inv(induce_image(c, involution(L))) for c in C]
+  new_coeff =  eltype(C)[inv(involution(L)(c)) for c in C]
   pm = pseudo_matrix(new_bmat, new_coeff)
-  return hermitian_lattice(base_field(L), pm,
-                           gram_ambient_space = gram_matrix(ambient_space(L)))
+  return lattice(ambient_space(L), pm)
 end
 
 ################################################################################
--- a/test/QuadForm/Lattices.jl
+++ b/test/QuadForm/Lattices.jl
@@ -20,7 +20,7 @@
     @test ambient_space(Lover) === ambient_space(L)
     L = Lover
   end
-  @test Hecke.ismaximal_integral(L, p )[1]
+  @test Hecke.ismaximal_integral(L, p)[1]
 
   @test ambient_space(dual(L)) === ambient_space(L)
   @test ambient_space(Hecke.lattice_in_same_ambient_space(L,pseudo_matrix(L))) === ambient_space(L)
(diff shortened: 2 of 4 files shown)


StevellM:
For equality, we could also create another function, of this kind (I did on my own repository, I was trying to implement equality of AbsLat): 

``` 
@doc Markdown.doc"""
    pseudo_hnf(L::AbsLat) -> AbsLat

Given a lattice $L$ with pseudo-matrix $M$, return the pseudo-hermite normal 
form of $M$, and cache it as the pseudo-matrix of $L$.
"""
function pseudo_hnf(L::AbsLat)
  M = L.pmat
  new_mat = pseudo_hnf(M, :lowerleft)
  L.pmat = new_mat
  return new_mat
end
```
so once computed for `L`, we cache the pseudo-hermite normal form of its `pseudo_matrix` and so one could add a line in the test of equality like

```
if L.pmat == M.pmat
  return true
end
```
so that if their respective pseudo-matrix are already in pseudo hnf form, we avoid having to recompute them (for instance if we run a code where we need to compare several lattices to a same fixed one)

thofma:
Thanks! Unfortunately we do not have uniqueness for pseudo-Hermite normal forms, but the caching will still help. I will amend the PR.

StevellM:
Yes I remarked this, I could take the `pseudo_hnf` of the `pseudo_hnf` of a pseudo matrix and get another result. I guess your `_modules_equality` is here for this non-uniqueness ?

thofma:
Yes (it also assumes `pseudo_hnf(..., :lowerleft)` :)

simonbrandhorst:
@StevellM  note that caching is okay. But mutating an object is dangerous and in general discouraged. For instance some properties of the lattice may depend on the pseudomatrix, for instance the automorphism group generators can be given with respect to the given pseudo basismatrix. If the pseudo basis silently changes, that invalidates prior computations and leads to bugs.

thofma:
Yes, anybody can access fields of objects and change them. I am not sure how this relates to caching. Touching fields of objects is done at your own risk.

StevellM:
Yes that's true, I always forget not to touch to those access... How should we proceed then ? create a new access for this pseudo matrix in pseudo hnf, or make sure that the pseudo matrix cached at the creation of the lattice is already in pseudo hnf ?

thofma:
I will fix it tomorrow by storing it explicitly.

simonbrandhorst (approved):

simonbrandhorst:
> I will fix it tomorrow by storing it explicitly.
I suggest the interface  `pseudo_hnf(L::AbsLat)`

simonbrandhorst:
@StevellM `dual(L)` should require that `L` is non-degenerate (otherwise the dual is not a lattice). Please add an `@req` check.

src/QuadForm/Herm/Lattices.jl
```diff
@@ -321,14 +320,15 @@ end
 ################################################################################
 
 function dual(L::HermLat)
-  G, B = _gram_schmidt(gram_matrix_of_rational_span(L), involution(L))
+  G = gram_matrix_of_rational_span(L)
+  @req rank(G) == nrows(G) "Lattice must be non-degenerate"
```
simonbrandhorst, line 324:
```
 rank(L::AbsLat) -> Int

  Returns the rank of the underlying module of L.
```
Hence this is not the correct test. I propose introducing a (cached) method `isdegenerate(L::AbsLat)` which returns `rank(L) == rank(rational_span(L))`

thofma: I am testing whether the Gram matrix has full rank. Why is this not correct?

thofma:
Ah, annoying. Can we use `issubset`? There the ordering is the other way around. I know this is an annoying inconsistency.

src/QuadForm/Herm/Lattices.jl
simonbrandhorst: ah I missred the lhs for `rank(L)` so nvm it is correct

simonbrandhorst:
We can enable issubset as well.

thofma:
I will fix it later today.

test/QuadForm/Lattices.jl
```diff
@@ -110,14 +110,39 @@
   o = @inferred Hecke.automorphism_group_order(L)
   @test o == 1536
 
+  @test ambient_space(P*L) === ambient_space(L)
+
+  @test issublattice(L, P*L)
+  @test !issublattice(P * L, L)
+  @test issubset(P*L, L)
+  @test !issubset(L, P*L)
+  @test isintegral(L) == issublattice(dual(L), :)
```
thofma, line 119:
```suggestion
  @test isintegral(L) == issublattice(dual(L), L)
```

PR #543, opened 2022-01-17
Caching local genus while taking a representative + typos

StevellM:
For `HermLat`, if we take a representative of a local genus `g`, the test `representative(g) in g` returns `false` so I modified it to cache this information in the `symbols` dictionary. There were also few typos that I depicted. I didn't put any tests yet because I needed the `dual` and equality from other PR to provide a larger coverage.

### DIFF
--- a/src/QuadForm/Herm/Genus.jl
+++ b/src/QuadForm/Herm/Genus.jl
@@ -175,7 +175,7 @@ function discriminant(G::LocalGenusHerm)
     return d
   end
   E = base_field(G)
-  fl = islocal_norm(E, base_field(K)(-1), prime(G))
+  fl = islocal_norm(E, base_field(E)(-1), prime(G))
   if fl
     return d
   else
@@ -479,7 +479,13 @@ Given a local genus, return a Hermitian lattice contained in this genus.
 """
 function representative(G::LocalGenusHerm)
   E = G.E
-  return lattice(hermitian_space(E, gram_matrix(G)))
+  p = prime(G)
+  L = lattice(hermitian_space(E, gram_matrix(G)))
+  S = ideal_type(base_ring(base_ring(L)))
+  symbols = Dict{S, LocalGenusHerm{typeof(base_field(L)),S}}()
+  set_attribute!(L, :local_genera => symbols)
+  symbols[p] = G
+  return L
 end
 
 ################################################################################
@@ -1375,11 +1381,10 @@ function local_genera_hermitian(E, p, rank::Int, det_val::Int, max_scale::Int, i
 end
 
 @doc Markdown.doc"""
-    genera_hermitian(E::NumField, rank::Int, determinant::Int, max_scale = nothing) -> Vector{GenusHerm}
+    genera_hermitian(E::NumField, rank::Int, signatures::Dict{InfPlc, Int}, determinant::Int, max_scale = nothing) -> Vector{GenusHerm}
 
-Return all genera of Hermitian lattices over $E$ at $\mathfrak p$ with
-rank `rank`, scale valuation bounded by `max_scale` and determinant valuation
-equal to `det_val`.
+Return all genera of Hermitian lattices over $E$ with rank `rank`, signatures given by `signatures`,
+scale bounded by `max_scale` and determinant equal to `det`.
 """
 function genera_hermitian(E, rank, signatures, determinant; max_scale = nothing)
   K = base_field(E)
--- a/src/QuadForm/Herm/Genus.jl
+++ b/src/QuadForm/Herm/Genus.jl
@@ -1381,10 +1381,10 @@ function local_genera_hermitian(E, p, rank::Int, det_val::Int, max_scale::Int, i
 end
 
 @doc Markdown.doc"""
-    genera_hermitian(E::NumField, rank::Int, signatures::Dict{InfPlc, Int}, determinant::Int, max_scale = nothing) -> Vector{GenusHerm}
+    genera_hermitian(E::NumField, rank::Int, signatures::Dict{InfPlc, Int}, determinant::nf_elem, max_scale = nothing) -> Vector{GenusHerm}
 
 Return all genera of Hermitian lattices over $E$ with rank `rank`, signatures given by `signatures`,
-scale bounded by `max_scale` and determinant equal to `det`.
+scale bounded by `max_scale` and determinant equal to `determinant`.
 """
 function genera_hermitian(E, rank, signatures, determinant; max_scale = nothing)
   K = base_field(E)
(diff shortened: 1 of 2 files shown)


src/QuadForm/Herm/Genus.jl
```diff
@@ -175,7 +175,7 @@ function discriminant(G::LocalGenusHerm)
     return d
   end
   E = base_field(G)
-  fl = islocal_norm(E, base_field(K)(-1), prime(G))
+  fl = islocal_norm(E, base_field(E)(-1), prime(G))
```
thofma, line 178: Not sure, why is this necessary?

StevellM: Becquse `K` is not defined :)

thofma: Oops, I should stop for today.

src/QuadForm/Herm/Genus.jl
```diff
@@ -479,7 +479,13 @@ Given a local genus, return a Hermitian lattice contained in this genus.
...
 function representative(G::LocalGenusHerm)
   E = G.E
-  return lattice(hermitian_space(E, gram_matrix(G)))
+  p = prime(G)
+  L = lattice(hermitian_space(E, gram_matrix(G)))
+  S = ideal_type(base_ring(base_ring(L)))
+  symbols = Dict{S, LocalGenusHerm{typeof(base_field(L)),S}}()
+  set_attribute!(L, :local_genera => symbols)
+  symbols[p] = G
+  return L
```
simonbrandhorst, line 488: This does not make sense to me. 
What is it supposed to fix? `representative(g) in g` works for me. Can you give an example where it fails?

simonbrandhorst (changes requested):

src/QuadForm/Herm/Genus.jl
thofma: The lattice `L = representative(g)` will remember its local genus at `p`. `genus(L, p)` should then be instant.

StevellM:
```
julia> K,a = quadratic_field(2)
(Real quadratic field defined by x^2 - 2, sqrt(2))

julia> Kt, t = K["t"]
(Univariate Polynomial Ring in t over K, t)

julia> OK = maximal_order(K)
Maximal order of Real quadratic field defined by x^2 - 2 
with basis nf_elem[1, sqrt(2)]

julia> E,b = number_field(t^2-a, "b")
(Relative number field over with defining polynomial t^2 - sqrt(2)
 over Real quadratic field defined by x^2 - 2, b)

julia> p = prime_ideals_over(OK,5)[1]
<5, 5>
Norm: 25
Minimum: 5
principal generator 5
two normal wrt: 5

julia> g = genus(HermLat, E, p, [(0,1,1,0),(2,2,-1,-1)])
Local genus symbol (scale, rank, det) at <5, 5>:
(0, 1, +)(2, 2, -)

julia> L = representative(g)
Hermitian lattice of rank 3 and degree 3
over
Relative maximal order with pseudo-basis (1) * 1//1 * <1, 1>, (b) * 1//1 * <1, 1>

julia> L in g
false
```
The problem is when we do the check `L in g`, if `get_attribute(L, :local_genera)` doesn't have a key `p`, it will recreate the local genus `g` but as julia object, this new version of `g` is distinct from the one we have started.

StevellM: For `HermLat` the way the representative is constructed, it doesn't keep systematically in memory that `genus(L,p) == g`... 
```
julia> genus(L,p) == g
false
```

thofma: Is the equality of local genera missing?

StevellM: A priori no, at least the method `function ==(G1::LocalGenusHerm, G2::LocalGenusHerm)` exists in /src/QuadForm/Herm/Genus.jl from the line 698

thofma: Something does not look right here. In particular
```
julia> L in g
false
```

StevellM: This is coming mainly from the function `genus(L,p)` which returns the value at the key `p` of `get_attribute(L, :local_genera)` and if it doesn't exist, it calls `Hecke._genus(L,p)` which creates a new copy of `g`, but which is not the same as `g`

thofma: It should not matter, since we define == for local genera. It does not fall back to === (which tests if the objects are identical).

StevellM:
That might help
```
julia> det(genus(L,p))
1

julia> det(g)
-1
```

src/QuadForm/Herm/Genus.jl
StevellM: That might help 
```
julia> det(genus(L,p))
1

julia> det(g)
-1
```

simonbrandhorst: Yes. There is a bug somewhere.

thofma: In this case `p` is split in E, so the genus symbol can never have a `-` as the determinant class? It seems
```
g = genus(HermLat, E, p, [(0,1,1,0),(2,2,-1,-1)])
```
makes no sense (this is the same as `[(0,1,1),(2,2,-1)])` Can you confirm @simonbrandhorst?

thofma: Nevermind, I am not sure what I wrote makes sense. I think I found the problem.

thofma: Still confused. I am looking at Prop. 3.3.5 of http://www.math.rwth-aachen.de/~Markus.Kirschmer/papers/herm.pdf. Consider `(2, 2 -1)`, so rank 2 of scale 2. We are not ramified, so in case 2 and the lattice looks like <z^2, z^2> where z is a uniformizer of p (notation as in Stevells example). So the determinant is z^4, which is a local norm:
```
julia> Hecke.islocal_norm(E, elem_in_nf(uniformizer(p)^4), p)
true
```
So I don't think (2, 2, -1) is a genus symbol?

P.S.: In the unramified case, scale and rank always determine the detclass.

thofma: @StevellM Where did you find that genus symbol?

simonbrandhorst:
Tests fail. Please fix.

thofma:
I will put the changes in https://github.com/thofma/Hecke.jl/pull/551

simonbrandhorst:
meaning that this one can be closed?

thofma:
No, I want to add the commits to my branch, so this here should not be deleted yet.

thofma:
I have these things now in #551.

PR #547, opened 2022-01-19
function added to check membership of a vector in a Zlattice.

FiroozehAga:


src/QuadForm/ClosestVector.jl
```diff
@@ -272,10 +272,12 @@ end
...
-  Return all the integer vectors `x` of length n such that the inhomogeneous 
-  quadratic function `q_{QT}(x) := xQx + 2xL + c <= 0` corresponding to an n variabled
-  quadratic triple. If the optional argument equal=true, it returns
-  all vectors `x` such that `q_{QT}(x) = 0`. By default equal=false.
+    Return all the integer vectors `x` of length n such that the inhomogeneous 
+    quadratic function `q_{QT}(x) := xQx + 2xL + c <= 0` corresponding to an n variabled
+    quadratic triple. If the optional argument ``equal = true``, it return
+    all vectors `x` such that `q_{QT}(x) = 0`. By default ``equal = false``.
+    If the argument ``sorting = true``, then we get a a list of sorted vectors.
+    The Default value for ``sorting`` is set to ``false``.
```
thofma, line 280:
```suggestion
Return all the integer vectors `x` of length n such that the inhomogeneous 
quadratic function `q_{QT}(x) := xQx + 2xL + c <= 0` corresponding to an n variabled
quadratic triple. If the optional argument ``equal = true``, it return
all vectors `x` such that `q_{QT}(x) = 0`. By default ``equal = false``.
If the argument ``sorting = true``, then we get a a list of sorted vectors.
The Default value for ``sorting`` is set to ``false``.
```

thofma: No indentation for the prose text. Also for the others below.

src/QuadForm/Quad/ZLattices.jl
```diff
@@ -934,3 +934,26 @@ function invariant_lattice(L::ZLat, G::MatElem;
...
+    Base.in(v::Vector, L::ZLat) -> Bool
+
+
+  This function checks if the vector 'v' lies in the lattice 'L' or not.
+"""
+function Base.in(v::Vector, L::ZLat)
+  @assert size(v)[1]==rank(L) "The vector should have the same length as the rank of the lattice."
+  B = basis_matrix(L)
+  V = matrix(QQ, size(v)[1], 1, v)
+  if !isone(denominator(V)) || !can_solve(B, V)
```
thofma, line 954: `[1//2]` is in the lattice with basis matrix `[1//2;]` but has denominator not one.

src/QuadForm/Quad/ZLattices.jl
```diff
@@ -934,3 +934,26 @@ function invariant_lattice(L::ZLat, G::MatElem;
...
+################################################################################
+#
+#  Membership check
+#
+################################################################################
+
+@doc Markdown.doc"""
+    Base.in(v::Vector, L::ZLat) -> Bool
+
+
```
simonbrandhorst, line 947:
```suggestion
```

src/QuadForm/Quad/ZLattices.jl
FiroozehAga: That is true. I didn't think about the non-integer cases. What would you recommend? I can add an option to make it possible for users to just get the answer for the integral case. Or should I just remove the denominator check altogether?

src/QuadForm/ClosestVector.jl
```diff
@@ -348,9 +350,10 @@ end
...
   
   
-  Returns all vectors `x` in `L` such that `b(v-x,v-x) <= c`, where `b` is the bilinear form on `L`.
-  If the optional argument ``equal = true`` then it return all vectors `x` in `L` such that `b(v-x,v-x) = c`.
-  By default equal=false.
+    Return all vectors `x` in `L` such that `b(v-x,v-x) <= c`, where `b` is the bilinear form on `L`.
+    If the optional argument ``equal = true`` then it return all vectors `x` in `L` such that `b(v-x,v-x) = c`.
+    By default ``equal = false``. If the argument ``sorting = true``, then we get a a list of sorted vectors.
+    The Default value for ``sorting`` is set to ``false``.
 """
```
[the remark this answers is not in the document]
FiroozehAga: will fix it:)

src/QuadForm/Quad/ZLattices.jl
thofma: You can call `fl, w = can_solve_with_solution(B, V)`. Then `fl` must be true and `w` must be integral. If one of those is not true, then `v` is not in the lattice.

simonbrandhorst: I wonder if we want to cache the hnf of L? And then use `can_solve_left_reduced_triu`?

thofma: Who knows that this function is doing. I would not try to do it in this PR here.

src/QuadForm/Quad/ZLattices.jl
```diff
@@ -934,3 +934,26 @@ function invariant_lattice(L::ZLat, G::MatElem;
...
+  This function checks if the vector 'v' lies in the lattice 'L' or not.
+"""
+function Base.in(v::Vector, L::ZLat)
+  @assert size(v)[1]==degree(L) "The vector should have the same length as the degree of the lattice."
+  B = basis_matrix(L)
+  V = matrix(QQ, size(v)[1], 1, v)
+  fl, w = can_solve_with_solution(B, V)
+  if !fl || !isone(denominator(w))
+    return false
```
simonbrandhorst, line 955: all of this can be shortened to one line:
`return fl && isone(denominator(w))`

simonbrandhorst (approved):

PR #548, opened 2022-01-19
various tests and fixes for quadratic lattices

simonbrandhorst:


src/QuadForm/Quad/ZGenus.jl
```diff
@@ -869,14 +869,15 @@ function Base.:(==)(G1::ZpGenus, G2::ZpGenus)
   # This follows p.381 Chapter 15.7 Theorem 10 in Conway Sloane's book
   @req prime(G1) == prime(G2) ("Symbols must be over the same prime "
                                 *"to be comparable")
+  sym1 = [g for g in symbol(G1) if g[2] != 0]
```
thofma, line 872: Is it even allowed to have `g[2] == 0`? I assume this is the rank?

simonbrandhorst: You are right, usually not. But that was how I implemented `genus(zero_matrix(ZZ,0,0))` one could give it the empty symbol list instead. Would need change elsewhere though.

thofma: In the hermitian case we have an empty list. Ut don't worry if it is too much work.

PR #549, opened 2022-01-20
Ambient space

simonbrandhorst:


test/QuadForm/Lattices.jl
```diff
@@ -11,17 +11,21 @@
...
+  L = quadratic_lattice(K,B, gram=D)
+  @test dim(ambient_space(L)) == 3
+  @test sprint(show, L) isa String
+  @test gram_matrix(ambient_space(L)) == D
 
+  B1 = pseudo_matrix(diagonal_matrix([a,a,a]))
+  @test_throws ArgumentError quadratic_lattice(K, B1, gram=D)
 
   fl = false
   while !fl
```
thofma, line 31: This loop is now running forever apparently?

simonbrandhorst: let's see if L is integral the loop must terminate or there is a bug ... however if L is not integral, I can see it happen.

PR #553, opened 2022-01-21
Sparse rref and kernel

joschmitt:
Here is a rref and kernel for sparse matrices. Sometimes they are faster than the regular ones...
I'm not sure what to return for the kernel: if we want to be consistent with the dense kernel, it would have to be a (sparse) matrix with the basis in the columns. But our sparse matrices are row based, so this would be pretty annoying to work with. So far I just return a vector of sparse rows. Or should one return a sparse matrix where the ROWS are the basis for the kernel (and document this)?
(This is also the reason why there is no `left_kernel` so far...)


thofma:
Thanks! @fieker Any good idea about the return type?

fieker:
On Fri, Jan 21, 2022 at 03:48:07AM -0800, Tommy Hofmann wrote:
> Thanks! @fieker Any good idea about the return type?
I think the kernel should be a dense matrix - as this is what it mainly
is.

joschmitt:
> I think the kernel should be a dense matrix - as this is what it mainly is.

Good point, thanks!

joschmitt:
Maybe don't merge this quite yet. I might have minor adjustments once I started using it.

joschmitt:
This is good to go now.

thofma:
The modules in Oscar use the sparse matrices, so the errors could be related?

joschmitt:
> The modules in Oscar use the sparse matrices, so the errors could be related?

Oh, you're right. I'll have a look.

joschmitt:
I think the problem was a missing deepcopy for `MPolyElem_dec`, see https://github.com/oscar-system/Oscar.jl/pull/1081 . Now the tests work for me locally.

thofma:
@fieker Do you want to have a look? Otherwise I will merge soon.

PR #556, opened 2022-01-21
Added involution for quadratic extensions

StevellM:
Up to now, we get the involution of a quadratic extension only via a `HermLat` or a `HermSpace`. I made available directly for any quadratic extensions and uses this function to create the involutions associated to `HermLat` and `HermSpace`. It seems to be more intuitive to get the involution directly from the field.

### DIFF
--- a/src/Map/automorphisms.jl
+++ b/src/Map/automorphisms.jl
@@ -185,7 +185,7 @@ function set_automorphisms(K::NfAbsNS, auts::Vector{NfAbsNSToNfAbsNS})
 end
 
 function involution(K::Union{NfRel, AnticNumberField})
-  @assert degree(K) == 2
+  @req degree(K) == 2 "Number field must have degree 2 over its base field"
   a = gen(K)
   A = automorphisms(K)
   if A[1](a) == a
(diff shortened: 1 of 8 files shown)


test/Map/NumField.jl
```diff
@@ -37,6 +40,7 @@
   f = @inferred hom(k, K, a^3)
   @test_throws ErrorException hom(k, K, a)
   @test f(b) == a^3
+  @test_throws AssertionError s = involution(K)
```
thofma, line 43:
```suggestion
  @test_throws ArgumentError s = involution(K)
```

src/Map/automorphisms.jl
```diff
@@ -184,6 +184,16 @@ function set_automorphisms(K::NfAbsNS, auts::Vector{NfAbsNSToNfAbsNS})
   return nothing
 end
 
+function involution(K::Union{NfRel, AnticNumberField})
+  @assert degree(K) == 2
```
thofma, line 188:
```suggestion
  @req degree(K) == 2 "Number field must have degree 2 over its base field"
```

test/Map/NumField.jl
StevellM: Is it the same ?

thofma: No, but `@req` throws an `ArgumentError`.

PR #559, opened 2022-01-21
shorten ZGenus test

simonbrandhorst:
Let's see if the coverage is okay.


thofma:
Looking at https://app.codecov.io/gh/thofma/Hecke.jl/compare/559/overview this seems to be OK?

simonbrandhorst:
The overview at least is not helpful, in the last commit I disabled the test altogether and there coverage of ZGenus dropped by 12%. But the long test finished in 72 instead of 85 minutes. So I will shorten it further ... 5 minutes should be enough,

thofma:
Feel free to merge at your own discretion.

PR #560, opened 2022-01-23
Fix: isupper_triangular

YueRen:
isupper_triangular should check matrix M row i only up to min(i-1,ncols(M)), not max(i-1, ncols(M)) which would be until the end of the row for small i. (bug found while debbuging pAdicSolver with Avi)


thofma:
Thanks Yue!

PR #563, opened 2022-01-24
Try to make the (parallel) tests faster

thofma:


thofma:
I am replacing the big job "test/QuadForm.jl" but whatever "test/QuadForm/*.jl" yields. If all the long things are in one file, this won't help too much.

CC: @simonbrandhorst

thofma:
Quadform/Quad.jl starts first and finished last 😄

I am afraid thus means that we have to move some tests around to make it faster.

thofma:
OK, 54 minutes for the long test.

PR #564, opened 2022-01-24
Fix definitness + tests for HermSpace

StevellM:
There is still a `_isdefinite(V::AbsSpace)`, I don't know whether or not it has to stay or if we can remove it regarding that we have the other functions below... seems to be a function created before `istotally_positive` and `istotally_negative`. Yet, this first one return -1, 0 or 1, instead of boolean, so maybe it has a purpose somewhere. I don't know.

### DIFF
--- a/src/QuadForm/Herm/Spaces.jl
+++ b/src/QuadForm/Herm/Spaces.jl
@@ -44,7 +44,7 @@ function hermitian_space(E::NumField, gram::MatElem; cached::Bool = true)
 
   involutionV = involution(E)
 
-  gramc == transpose(matrix(involutionV.(gramc))) || throw(ArgumentError("$gram must be hermitian"))
+  @req gramc == transpose(map_entries(involutionV, gramc)) "$gram must be hermitian"
 
   K = base_field(E)
 
--- a/src/QuadForm/Herm/Spaces.jl
+++ b/src/QuadForm/Herm/Spaces.jl
@@ -287,7 +287,7 @@ end
 Return whether $U$ is represented by $V$ locally at $\mathfrak p$.
 """
 function islocally_represented_by(U::HermSpace, V::HermSpace, p)
-  if rank(U) < rank(V)
+  if rank(U) > rank(V)
     return false
   elseif rank(U) == rank(V)
     return isisometric(U, V, p)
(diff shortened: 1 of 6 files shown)


src/QuadForm/Herm/Spaces.jl
```diff
@@ -45,6 +44,8 @@ function hermitian_space(E::NumField, gram::MatElem; cached::Bool = true)
 
   involutionV = involution(E)
 
+  gramc == transpose(matrix(involutionV.(gramc))) || throw(ArgumentError("$gram must be hermitian"))
```
thofma, line 47:
```suggestion
  @req gramc == transpose(map_entries(involutionV, gramc)) "$gram must be hermitian"
```

src/QuadForm/Spaces.jl
```diff
@@ -355,9 +355,13 @@ isisometric(L::AbsSpace, M::AbsSpace, p)
...
 function _isdefinite(V::AbsSpace)
-  K = fixed_field(V)
-  R = maximal_order(K)
-  if !istotally_real(K) || (ishermitian(V) && !istotally_complex(K))
+  E = base_ring(V)
+  if isquadratic(V)
+    K = E
+  else
+    K = base_field(E)
+  end
```
thofma, line 363: Is `K = fixed_field(V)` not doing the right thing?

StevellM:
Ok now I know why this internal `_isdefinite` had a purpose, and that's why the tests are failing, I will put it back how it was (the file for the mass functions)

thofma:
What is the purpose?

StevellM:
In the file `~/QuadForm/Herm/Mass.jl`, there is this internal `_local_factor_generic(L,p)` for which we precisely need to know if it is indefinite, positive definite or negative definite because at some point, there is the rescaling of a uniformiser according to the definiteness of the lattice, and for this it needs to be an element of the base field, not only a boolean.

thofma:
Thanks!

src/QuadForm/Spaces.jl
thofma, line 363:
```suggestion
  fixed_field(V)
```

thofma, line 363:
```suggestion
  K = fixed_field(V)
```

StevellM:
Should I commit also the same change for the two next functions ?

thofma:
If you don't mind, that would be great.

PR #566, opened 2022-01-27
Enable more tests

thofma:


thofma:
@fingolfin Do you see the failure on nightly? That is why I don't like using internals 😄

fingolfin:
I wouldn't touch anything in `Core.Compiler` with a 10 foot pole. Which is why I never used `Core.Compiler.return_type`. But `Base.return_types` seems OK?

thofma:
Don't understand why it would be more OK. Is also not part of the official API and not documented. It has exactly the same "problems".

PR #570, opened 2022-01-29
Revise how maximal orders are stored/accessed

fingolfin:


thofma:
You are very brave to touch these parts of the code base :) I guess this needs a new AA release?

src/NumFieldOrd/NfRelOrd/NfRelOrd.jl
```diff
@@ -527,18 +519,11 @@ end
...
 
 function MaximalOrder(L::NumField)
-  try
-    O = _get_maximal_order_of_nf_rel(L)
-    return O::order_type(L)
-  catch e
-    if !isa(e, AccessorNotSetError)
-      rethrow(e)
-    end
+  return get_attribute!(K, :maximal_order) do
```
fingolfin, line 522:
```suggestion
  return get_attribute!(L, :maximal_order) do
```

fingolfin:
Seems to pass tests now (but I've squashed and rebased once more, so let's see)

fingolfin:
It would be good to know which of the modified code actually is exercised by the test suite. For Oscar.jl, we see that inline thanks to Codecov reporting it there. I believe this is a matter of "installing" `https://github.com/marketplace/codecov` for this repository

thofma:
We have inline comments too.

thofma:
I think they got removed because the job was rerun?

thofma:
I can take it from here. I saw some other minor problems, but they were there before.

fingolfin:
Thanks @thofma, go ahead then :-).

And I do see Codecov comments now. Maybe last night I looked "too early" before Codecov was done processing the reports?

src/RCF/conductor.jl
```diff
@@ -942,14 +942,11 @@ the algorithm is applied to the normal closure of $K$ (without computing it).
 """
 function maximal_abelian_subfield(K::NfRel{nf_elem}; of_closure::Bool = false)
   zk = maximal_order(base_field(K))
-  d = ideal(zk, discriminant(K))
-  try
-    ZK = _get_maximal_order_of_nf_rel(K)
+  if has_attribute(K, :maximal_order)
+    ZK = get_attribute(K, :maximal_order)
```
fingolfin, line 946: Perhaps should be done like this instead (I wrote part of this patch before I discovered `ismaximal_order_known`)
```suggestion
  if ismaximal_order_known(K)
    ZK = maximal_order(K)
```

src/Hecke.jl
```diff
@@ -275,12 +265,16 @@ include("Deprecations.jl")
 #
 ################################################################################
 
+function ismaximal_order_known(K::AnticNumberField)
+  return has_attribute(K, :maximal_order)
+end
+
 function _get_maximal_order(K::AnticNumberField)
```
fingolfin, line 272: I think nothing is left that is using `_get_maximal_order` (instead code just calls `maximal_order`), so perhaps it can be removed. Alternatively, one could change code to use `_get_maximal_order` again, to emphasize that precomputed data is accessed

thofma: will think about it and adjust it, thanks.

thofma:
I think once a new commit(?) is pushed, the old comments are immediately removed. We take the coverage from the "Run long tests" which usually finishes last.

src/NumFieldOrd/NfRelOrd/NfRelOrd.jl
```diff
@@ -564,21 +549,14 @@ function maximal_order_via_simple(m::NumFieldMor{<:NfRel, <:NfRelNS})
...
+    L = codomain(m)
+    OL = maximal_order(L)
+    B = absolute_basis(OL, L)
+    OK = Order(K, [ m\b for b in B ], check = false, isbasis = true)
+    OK.ismaximal = 1
+    return OK
   end
```
fingolfin, line 559: maybe should also add a type assertion here on the return type?

src/NumField/NfAbs/NonSimple.jl
```diff
@@ -46,16 +46,8 @@ export issimple, simple_extension
...
-function _get_maximal_order(K::NfAbsNS)
-  if isdefined(K, :O)
-    return K.O
-  else
-    throw(AccessorNotSetError())
-  end
-end
-
-function _set_maximal_order(K::NfAbsNS, O::NfAbsOrd{NfAbsNS, NfAbsNSElem})
-  K.O = O
```
fingolfin, line 58: Actually... this patch should then also remove member `O` from struct `NfAbsNS` ! I wanted to do that, but it seems I forgot :-(.

src/NumFieldOrd/NfRelOrd/NfRelOrd.jl
```diff
@@ -564,21 +549,14 @@ function maximal_order_via_simple(m::NumFieldMor{<:NfRel, <:NfRelNS})
...
-  _set_maximal_order(K, OK)
-  return OK
+  return get_attribute!(K, :maximal_order) do
+    L = codomain(m)
+    OL = maximal_order(L)
+    B = absolute_basis(OL, L)
+    OK = Order(K, [ m\b for b in B ], check = false, isbasis = true)
+    OK.ismaximal = 1
+    return OK
+  end::order_type(OK)
```
fingolfin, line 559: Does this really work? Isn't `OK` out of scope here?

fingolfin: Maybe 
```suggestion
  end::order_type(K)
```
But there were bo test failures, so either a test should be added or I am misunderstanding something badly

thofma: Yes, there was a test failure, see https://github.com/thofma/Hecke.jl/runs/5009676182?check_suite_focus=true

PR #571, opened 2022-01-30
Updated functions in ``QuadForm/ClosestVector.jl`` for type stability

FiroozehAga:


src/QuadForm/ClosestVector.jl
```diff
@@ -1,65 +1,55 @@
...
 @doc Markdown.doc"""
-    convert_type(Q::MatrixElem, K::MatrixElem, d::RingElement) -> ZLat, Vector{RingElement}, RingElement
+    convert_type(Q::fmpq_mat, K::Vector{fmpq}, d::fmpq) -> ZLat, Vector{fmpq}, fmpq
 
-  Converts a quadratic triple QT = [Q, K, d] to the input values required for closest vector problem (CVP). 
+Converts a quadratic triple QT = [Q, K, d] to the input values required for closest vector problem (CVP). 
 """
-function convert_type(G::MatrixElem, K::MatrixElem, d::RingElement)
-    @assert typeof(G) == typeof(K) "Input matrices should be of the same type"
+function convert_type(G::fmpq_mat, K::Vector{fmpq}, d::fmpq)
```
fingolfin, line 8: This achieves type stability by sacrificing generality. If that's what you want... but it's not strictly necessary , there are other ways.

actually you also change here and elsewhere some arguments from (single row/column) matrices to Vector...

src/QuadForm/ClosestVector.jl
```diff
@@ -69,20 +59,26 @@ function pojective_quadratic_triple(Q::MatrixElem, L::MatrixElem, c::RingElement
...
     for j in 1:size(pL[1],1)
         pt1 = transpose(pL[2][j])
         pQT = pL[1][j] - pL[4][j]^-1 * (pL[2][j] * pt1)
-        pLL = pL[3][j] - ((pL[5][j] * pL[4][j]^-1)[1] * pL[2][j])
-        c1 = c1 - (pL[5][j]^2 * pL[4][j]^-1)[1] #rational number
+        pLL = matrix(pL[3][j]) - ((pL[5][j] * pL[4][j]^-1) * pL[2][j])
+        PLL = typeof(L)(undef, size(pLL)[1])
+        for j2 in 1:size(pLL)[1]
+            PLL[j2] = pLL[j2]
+        end
```
simonbrandhorst, line 75: This looks like you are copying `pLL` and give the copy the name`PLL`. Why? Before you did not copy.

src/QuadForm/ClosestVector.jl
```diff
@@ -69,20 +59,26 @@ function pojective_quadratic_triple(Q::MatrixElem, L::MatrixElem, c::RingElement
         push!(pp1, p1)
         push!(LL1, L1)
         push!(r1, Q[n+1-i, n+1-i])
-        push!(m1, L[n+1-i,1:1])
+        push!(m1, L[n+1-i,1:1][1])
         pL =[Lqqt, pp1, LL1, r1, m1]
     end
     #------------------------------------------------
+    c1 = c; pQT = []; ppQT = []; pLL = []; 
+    ppLL = []; cc1 = []; Lqt = [];
```
simonbrandhorst, line 67: Note that these arrays are quite inefficient since they are of type `Vector{any}`.

src/QuadForm/ClosestVector.jl
FiroozehAga: I do that to convert the type of the list from ```fmpq_mat``` to ```Vector{fmpq}``` So that the output tuple type is consistent with the input parameters. I can leave the result as ```fmpq_mat```. But I can also change the input parameter ```Vector{fmpq}``` to a 1xn ```fmpq_mat```. Then the output would also be consistent with the input types. 
What would you recommend?

src/QuadForm/ClosestVector.jl
FiroozehAga: I'll keep it in mind. Will improve the code gradually:)

src/QuadForm/ClosestVector.jl
```diff
@@ -69,20 +59,26 @@ function pojective_quadratic_triple(Q::MatrixElem, L::MatrixElem, c::RingElement
         push!(pp1, p1)
         push!(LL1, L1)
         push!(r1, Q[n+1-i, n+1-i])
-        push!(m1, L[n+1-i,1:1])
+        push!(m1, L[n+1-i,1:1][1])
         pL =[Lqqt, pp1, LL1, r1, m1]
```
fingolfin, line 63: Move this line out of the loop? And perhaps turn it into a tuple -- or get rid of it entirely?

src/QuadForm/ClosestVector.jl
```diff
@@ -69,20 +59,26 @@ function pojective_quadratic_triple(Q::MatrixElem, L::MatrixElem, c::RingElement
...
     end
     #------------------------------------------------
+    c1 = c; pQT = []; ppQT = []; pLL = [];
```
fingolfin, line 66:
```suggestion
    c1 = c
    ppQT = []
```
No point in initializing variables that you the overwrite with other values.

src/QuadForm/ClosestVector.jl
```diff
@@ -69,20 +59,26 @@ function pojective_quadratic_triple(Q::MatrixElem, L::MatrixElem, c::RingElement
...
         push!(LL1, L1)
         push!(r1, Q[n+1-i, n+1-i])
-        push!(m1, L[n+1-i,1:1])
+        push!(m1, L[n+1-i,1:1][1])
         pL =[Lqqt, pp1, LL1, r1, m1]
     end
     #------------------------------------------------
+    c1 = c; pQT = []; ppQT = []; pLL = []; 
+    ppLL = []; cc1 = []; Lqt = [];
     for j in 1:size(pL[1],1)
```
fingolfin, line 68: So `pL[1] === Lqqt` by definition, so why not use this? What is the point of `pL`

fieker:
On Mon, Jan 31, 2022 at 10:19:54PM -0800, Max Horn wrote:
> @fingolfin commented on this pull request.
> 
> 
> 
> >          pL =[Lqqt, pp1, LL1, r1, m1]
>      end
>      #------------------------------------------------
> +    c1 = c; pQT = []; ppQT = []; pLL = []; 
> 
> ```suggestion
>     c1 = c
>     ppQT = []
> ```
> No point in initializing variables that you the overwrite with other values.
Wrong comment here? But there is an imporant point sometimes in
intialising: scoping...

src/QuadForm/ClosestVector.jl
fieker: I would recommend to think this carefully through: A `Vector{fmpq}` is a julia-construct that we interpret as either row or column depending on context. However, the memory layout is different. For fmpq it's not too bad, but for fmpz the performance on say addition of Vector{fmpz} and fmpz_mat could easily be a magnitude, due to different memory layout...
For more interesting operations this can get more and more interesting. 
What do you want to do?

fingolfin:
> Wrong comment here? But there is an imporant point sometimes in intialising: scoping...

No, it was in the intended place (I removed two initializations; this is easier to see in the web interface than via email). And yes, scoping can be a reason, but here the variables in question (`pQT` and `pLL`) are internal to the loop, so initializing them to a value (of a different type as well) before the loop is not helpful.

Of course I only *think* that this is what is happening; it is rather easy to get confused when there are so many similar variable names like `p1`, `pL`, `pLL`, `ppLL`, `PLL`.... so maybe that's another recommendation to @FiroozehAga: to consider using more self-explanatory variable names :-)

src/QuadForm/ClosestVector.jl
FiroozehAga: Thank you for your comment. That is good to know. I will keep this in consideration for future code. For the moment, this file constitutes of pre-procedures required for the enumeration of closest vectors. Most of the functions are not optimized and will change over time. In practice, the user should only call one of the ```closest_vectors()``` functions at the end of this file and the rest would eventually be private functions (hidden away unless someone needs them). The one that takes ZLat (as an input parameter) is the optimized one at the moment.

simonbrandhorst: Do you actually need rational matrices, i.e. are you performing a divison anywhere?
Then `fmpz` is the better choice or perhaps even `Int` if you can bound the ellipsoid and any other integers appearing.

FiroozehAga:
Thank you for your comments. I will update the functions accordingly. Just wanted to mention that the reason the functions might look a little strange at the moment (including variable names and/or other choices in the code) is mostly that it is a prototype since I am reproducing someone else's constructions. The code would change over time especially in the optimization phase of my work. Also, at the time I wrote the initial functions, certain functionality was not available in Hecke so I had to improvise and use some hacks to make the code work. But all of that will gradually change as we achieve more stability:)

src/QuadForm/ClosestVector.jl
FiroozehAga: That is just me being lazy and I didn't want to remember the names of different lists in the function. But I can change it.

src/QuadForm/ClosestVector.jl
FiroozehAga: Hello, what would be another way of doing this? I am not very familiar with other methods. Could you please guide me?

PR #572, opened 2022-01-30
Remove _get_fac_elem_mon_of_nf

fingolfin:
This `FacElemMon{AnticNumberField}` constructor is a bit weird. It seems to override the default caching strategy (where `FacElemMonDict` is used to store the cached object) by one where the data is instead stored as an attribute of the given `AnticNumberField`. But it still *also* checks if the required object already exists in `AnticNumberField`...

But the key used is the ring itself, so how could `haskey(FacElemMonDict, R)` ever return `true`? I am tempted to just remove that check

Independently, I wonder: why bother overriding the cache at all in this special case? My guess: because the global `FacElemMonDict` really is a leak. But then perhaps this special case should be extended to all rings that support attribute storage (which we nowadays can check using `_is_attribute_storing_type`. Moreover, perhaps `FacElemMonDict` (and all its relatives) should be changed to use `AbstractAlgebra.CacheDictType`

### DIFF
--- a/src/Hecke.jl
+++ b/src/Hecke.jl
@@ -204,11 +204,6 @@ function __init__()
   global _get_maximal_order_of_nf_rel = t[1]
   global _set_maximal_order_of_nf_rel = t[2]
 
-  t = create_accessors(AnticNumberField, FacElemMon{AnticNumberField}, get_handle())
-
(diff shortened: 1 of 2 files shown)


thofma:
Yes, there is something too complicated going on here. Will investigate in the future.

PR #574, opened 2022-01-30
Adjust to Nemo changes

thofma:


src/Misc/FiniteField.jl
```diff
@@ -12,11 +12,6 @@ function FlintFiniteField(p::fmpz; cached::Bool = true)
   return k, k(1)
 end
 
-function FlintFiniteField(p::Int, k::Int; cached::Bool = true)
-  @assert isprime(p)
-  return FlintFiniteField(p, k, "o", cached = cached)
-end
-
 GF(p::Integer, k::Int, s::AbstractString="o"; cached::Bool = true) = FlintFiniteField(p, k, s, cached = cached)[1]
 GF(p::fmpz, k::Int, s::AbstractString="o"; cached::Bool = true) = FlintFiniteField(p, k, s, cached = cached)[1]
```
fingolfin, line 16: I was wondering if those can't also go, but they still diverge from their Nemo counterparts:
- Here `s` can be specified, but not in Nemo
- Nemo version call `isprime` resp. `isprobable_prime`
- Nemo only supports `Int` and `UInt` but not arbitrary `Integer`... But actually it is not clear to me how that would work here, given that there is no `FlintFiniteField` taking any `Integer`, such as a `BigInt`?! So perhaps  `GF(p::Integer, ...)` and also the `FlintFiniteField(p::Integer ...)` method at the top of the file should go?

If nothing else, it would be good to do this:
```suggestion
GF(p::Integer, k::Int, s::Union{AbstractString,Symbol}="o"; cached::Bool = true) = FlintFiniteField(p, k, s, cached = cached)[1]
GF(p::fmpz, k::Int, s::Union{AbstractString,Symbol}="o"; cached::Bool = true) = FlintFiniteField(p, k, s, cached = cached)[1]
```

src/Misc/FiniteField.jl
```diff
@@ -12,11 +12,6 @@ function FlintFiniteField(p::fmpz; cached::Bool = true)
   return k, k(1)
```
fingolfin, line 12: Can't `FlintFiniteField(p::fmpz, ...` also go? Flint has that now, too, with signature `FlintFiniteField(char::fmpz, deg::Int, s::Union{AbstractString,Symbol} = :o; cached = true)`.

This leaves the method accepting any `Integer`. For that it'd be great it also had argument `s::Union{AbstractString,Symbol}="o"` to match its "sibling methods".

thofma: No, Nemo does not have them as far as I can see:
```
julia> FiniteField(2)
ERROR: MethodError: no method matching FlintFiniteField(::Int64)
Closest candidates are:
  FlintFiniteField(::Int64, ::Int64) at ~/.julia/dev/Nemo/src/flint/fq_nmod.jl:664
  FlintFiniteField(::Int64, ::Int64, ::Union{AbstractString, Symbol}; cached) at ~/.julia/dev/Nemo/src/flint/fq_nmod.jl:664
  FlintFiniteField(::Union{fmpz_mod_poly, gfp_fmpz_poly}) at ~/.julia/dev/Nemo/src/flint/fq.jl:633
```

fingolfin: My mistake, I somehow completely missed the omitted degree argument

src/Misc/FiniteField.jl
```diff
@@ -12,13 +12,8 @@ function FlintFiniteField(p::fmpz; cached::Bool = true)
...
 
-function FlintFiniteField(p::Int, k::Int; cached::Bool = true)
-  @assert isprime(p)
-  return FlintFiniteField(p, k, "o", cached = cached)
-end
-
-GF(p::Integer, k::Int, s::AbstractString="o"; cached::Bool = true) = FlintFiniteField(p, k, s, cached = cached)[1]
-GF(p::fmpz, k::Int, s::AbstractString="o"; cached::Bool = true) = FlintFiniteField(p, k, s, cached = cached)[1]
+GF(p::Integer, k::Int, s::Union{AbstractString,Symbol}="o"; cached::Bool = true) = FlintFiniteField(p, k, s, cached = cached)[1]
+GF(p::fmpz, k::Int, s::Union{AbstractString,Symbol}="o"; cached::Bool = true) = FlintFiniteField(p, k, s, cached = cached)[1]
```
fingolfin, line 16: The very first thing `FlintFiniteField` does is to convert `s` into a symbol, so might as well start with a symbol as default argument?
```suggestion
GF(p::Integer, k::Int, s::Union{AbstractString,Symbol}=:o; cached::Bool = true) = FlintFiniteField(p, k, s, cached = cached)[1]
GF(p::fmpz, k::Int, s::Union{AbstractString,Symbol}=:o; cached::Bool = true) = FlintFiniteField(p, k, s, cached = cached)[1]
```

src/Misc/FiniteField.jl
fingolfin: Hrmmm.... the "Mark as resolved" button does not show up for this conversation. GitHub bug?

PR #576, opened 2022-02-02
Replace subgroup by sub

fingolfin:
Completely untested, but `subgroup` does not seem to be defined anywhere? Perhaps I missed something, or perhaps these functions are simply not covered by CI?

Motivated by https://github.com/oscar-system/Oscar.jl/issues/667﻿


thofma:
I think I used it in some project a while ago. I am sure I added it for a reason, but unfortunately without tests it seems.

fingolfin:
I guess you renamed those in https://github.com/thofma/Hecke.jl/pull/396 and then perhaps due to some time overlap between PRs these additional uses crept in?

PR #577, opened 2022-02-02
Fix `genus_representatives`

StevellM:
I have fixed `genus_representatives` for hermitian lattices. I had to modify again the code for `genus_generators` regarding loops that were not tested and broken. I found a nice example in a paper of M. Kirschmer to improve the test coverage of the functions. I tidied up a bit and put everything related to neighbours methods and genus representatives in the proper file `GenusRep.jl` and the functions related to orthogonal sums of local genera into the file `Genus.jl`. Unfortunately we can't see precisely the changes I have made because I forgot to do commits step by step... However most of the fixes are coming from undefined functions or incompatibility of some arguments: I made the current code more "Hecke-type" to make it run in any cases.

There is an extra function called `relcallback` that is temporary until an isometry test for indefinite lattices is implemented.


src/QuadForm/Herm/GenusRep.jl
```diff
@@ -48,86 +8,171 @@ end
...
+  return LL
+end
+
+function stdcallback(list, L)
+  keep = all(LL -> !isisometric(LL,L)[1], list)
+  return keep, true
+end
+
+#TODO : implement isometry check for indefinite lattice in the same space
+function relcallback(list,L)
```
thofma, line 155: Did we always have this? I wonder what the "rel" stands for?

src/QuadForm/Herm/GenusRep.jl
```diff
@@ -284,75 +320,43 @@ function _neighbours(L, P, result, max, callback = stdcallback, use_auto = true)
...
+    iterated_neighbours(L:HermLat, P::NfRelOrdIdl; use_auto = false, max = inf,
+				                   callback = false,
+						   missing_mass = Ref{fmpq}(zero(fmpq)))
+
+Return a set of representatives of `N(L,P)` (see [Kir16, Definition 5.2.6]). At most
+`max` representatives are returned.
 
+The use of the automorphism group of `L` can be enabled by `use_auto = true`. The
+use of the callback function can be enabled by `callback = stdcallback`. By defaut,
+the use of the mass in disabled.
```
thofma, line 349:
```suggestion
the use of the mass is disabled.
```

src/QuadForm/Herm/GenusRep.jl
StevellM: No, only for `neighbours_with_ppower` because it is called outside of the loop running for definite lattices. So in particular, there is this test running even for indefinite lattices.  M. Kirschmer seems to avoid this by not allowing the use of the automorphism group but even though we do the same, asking for an isometry test will use this group... Here the "rel" stands for "relative", that's a way I found to differentiate it from `stdcallback` for which I guess "std" stands for "standard".

StevellM: Ah and maybe I misunderstood: we never had this function, I added it waiting to fix isometry check for indefinite lattices

src/QuadForm/Herm/GenusRep.jl
```diff
@@ -48,86 +8,171 @@ end
 
 ################################################################################
 #
-#  Genus representatives for hermitian lattices
+#  Neighbours methods
 #
 ################################################################################
 
+function _all_row_span(M)
```
thofma, line 15: It appears this function is not called. Maybe we can add an example where this is used?

But I am also happy to improve the coverage later if you want to first get this in.

StevellM: For now yes, it is used in a specific loop of `_neighbours` that is not covered because it needs some specific examples I guess... Then we can wait before merging so that I can try to find/construct a nice example that will target this specific loop.

src/QuadForm/Herm/GenusRep.jl
thofma: OK

simonbrandhorst: Using the automorphism group does not make sense in the indefinite case. There we do not need to iterate over all the p-neighbors up to isometry. Instead we just visit all spinor genera.

src/QuadForm/Herm/GenusRep.jl
```diff
@@ -48,86 +8,171 @@ end
...
 
 function _neighbours(L, P, result, max, callback = stdcallback, use_auto = true)
-  ok, scale = ismodular(L, P);
+  ok, scale = ismodular(L, P)
   @req ok "The lattice must be locally modular"
   R = base_ring(L)
   K = nf(R)
   a = involution(L)
+
+  if !isdefinite(L) && use_auto
```
simonbrandhorst, line 171: I would rather raise an error instead of silently overwriting the user input.

src/QuadForm/Herm/GenusRep.jl
StevellM: In this context it is not directly used, but the short vectors and Cholesky decomposition to check if two lattices are isometric... because when adding a new candidates as a neighbour, we need to check if it was not already in the list of candidates already computed

src/QuadForm/Herm/GenusRep.jl
thofma: One problem with that is that if we follow through with this everywhere, we could not change the default `use_auto` to `true` in `iterated_neighbors`? (Because then for indefinite lattices it would be per default an error). I would rather document it in the form: "is `use_auto` is `true`, then the function will use automorphisms in the definite case. In the indefinite case this keyword has no effect."

StevellM: Yes, you are right.. Here I did this because, by default, `use_auto = true`, so I just adapted the automatic output to the indefinite case. Maybe it will more natural to set `use_auto = false` and then turn this into an input check... should I try to make this change ?

src/QuadForm/Herm/GenusRep.jl
```diff
@@ -393,4 +395,314 @@ function iterated_neighbours(L::HermLat, P; use_auto = false, max = inf,
   return result
 end
 
+@doc Markdown.doc"""
+    neighbours_with_ppower(L::HermLat, P::NfRelOrdIdl, e::Integer, use_auto = true)
+
+Return a representative of an isometry class in the special genus of `L`,
```
simonbrandhorst, line 401: There is no guarantee that the output lattice lies in the same special genus as `L`.
Rather the way that it is used is exactly to produce a lattice in the genus which is in a different special genus.
That is the purpose of Markus Algorithm 4.7.

It rather outputs a chain of p-neighbors where the next one is different from the previous one.
So that it is indeed a chain and not a loop.

Return a chain of `P` neighbors `of length `e`. 
`L=L_1, L_2 ... , L_e` such that L_{i-1} != L_{i+1} for `i = 2, \dots , e-1`.

src/QuadForm/Herm/GenusRep.jl
StevellM: I didn't see Tommy's comment before posting mine. I have actually followed this and change the documentations by explicitly saying that `use_auto` won't have any effect if the lattice is indefinite, and the boolean value of this entry is considered only in the definite case.

PR #590, opened 2022-02-11
rename exclude. It conflicts with exclude

fieker:
If using Revise when changing tests this is neccessary


test/runtests.jl
```diff
@@ -30,6 +30,7 @@ end
 # Is GAP there?
 _with_gap = false
 
+#=
```
thofma, line 33: Warum auskommentiert?

PR #592, opened 2022-02-11
admit also types instead of functions in `MapFromFunc`

ThomasBreuer:
An example where this extension turns out to be useful is shown in the comments in oscar-system/Oscar.jl/pull/1067.

### DIFF
--- a/src/Map/MapType.jl
+++ b/src/Map/MapType.jl
@@ -41,8 +41,8 @@ mutable struct MapCache{D, C, De, Ce}
   pr::Dict{Ce, De}
   prStat::Dict{Ce, Int}
 
-  old_im::Base.Callable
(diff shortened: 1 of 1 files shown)


thofma:
There are many other things which are callable (@fingolfin could tell you more about this :)).

I don't think we lose anything by leaving it untyped. It is runtime dispatch anyway.

ThomasBreuer:
I do not understand the code coverage problems: The report says that one line has been added, but where?
How can the code coverage decrease when all changes make the code lines just shorter?

thofma:
Because the code as well as the tests are randomized in part. So not all branches will in every run be covered. Already a simple `x = rand(F); while !iszero(x); ...; x = rand(F); end` will introduce these fluctuations.

PR #593, opened 2022-02-11
Make subfield return integral primitive elements

thofma:


fieker (commented):
The hom has to be changed as well...

thofma:
Right, obviously also needs a better test.

fieker:
The test would be fine, but check is set to false in creating the hom....

PR #594, opened 2022-02-11
Overload compose instead of *

thofma:


simonbrandhorst:
how about overloading `\circ`? That is used universally among mathematicians as far as I know.

simonbrandhorst:
Ah well, guess it's only ASCII now.

thofma:
There is a generic fallback from `*` to `compose`. I am just switching it up here to implement `compose` so that both `compose` and `*` will work.

PR #595, opened 2022-02-14
Replace subgroup by sub

fingolfin:
Huh, apparently I did not push the latest version of this branch for PR #576 :-(


StevellM:
A comment not regarding the changes: both of the functions have a field `K` as an argument but don't use it... is it superfluous ? Or maybe should we add a test like `K == domain(mG)` ?

thofma:
Eh, yes @StevellM there is something to be improved here :)

PR #597, opened 2022-02-14
Remove the two last accessors

fingolfin:


src/NumFieldOrd/NfOrd/Clgp.jl
```diff
@@ -463,11 +462,11 @@ obtained via `[ f(U[1+i]) for i in 1:unit_group_rank(O) ]`.
 All elements will be returned in factored form.
 """
 function unit_group_fac_elem(O::NfOrd; method::Int = 3, unit_method::Int = 1, use_aut::Bool = false, GRH::Bool = true, redo::Bool = false)
-  U = _get_UnitGrpCtx_of_order(O, false)
+  U = get_attribute(O, :UnitGrpCtx)
```
fingolfin, line 465: I debated adding type annotations here and elsewhere:
```suggestion
  U = get_attribute(O, :UnitGrpCtx):: UnitGrpCtx
```
but the current code doesn't have these either, plus this still wouldn't be a *concrete* type. 

But looking at it again now, it seems all concrete instances are of type `UnitGrpCtx{FacElem{nf_elem, AnticNumberField}}` anyway?!

thofma: Yes, indeed. It seems we did not use the parameter in the end, haha.

thofma:
Thanks for your heroic effort

PR #598, opened 2022-02-15
change on degree of AbsLat

StevellM:


test/QuadForm/Lattices.jl
```diff
@@ -47,6 +47,7 @@
   # test lazy creation of the ambient space.
   L = quadratic_lattice(K,Bp, gram=D)
   @test dim(ambient_space(L)) == 3
+  @test degree(L) == ncols(L.pmat.matrix)
```
thofma, line 50: I don't think the tests should test against internal stuff.  Why doesn't it just compare with `3`?

StevellM: Because in this case since `L` has a defined ambient space the degree uses the dimension of this ambient space. I just wanted to ensure that the second definition, in the case the ambient space is not defined, makes sense when we have the ambient space.

PR #601, opened 2022-02-15
Avoid deprecated 4-arg MatrixSpace method

fingolfin:


fingolfin:
Tests failed because Nemo was still providing some 4-arg versions of MatrixSpace (and *only* 4-arg, not the new-style 3-arg methods). So this requires https://github.com/Nemocas/Nemo.jl/pull/1283 to be merged and released

PR #602, opened 2022-02-16
Some fixes and new tests

StevellM:


thofma:
I don't understand dropping the `inf`? It is type unstable on purpose. If the return type is a union of few elements, the performance penalty is negligible. The `valuation` function will become type unstable in the same way in the future.

StevellM:
I had to because there was a bug with a test, see the previous commit... At some point there was a valuation compared with a normic defect, but valuations are of type `Int64` while in my changes normic defects were `IntExt`. So while inferring the  type of the normic defect, Julia couldn't compare both values because they didn't have the same type...

thofma:
This sounds like there is an underlying bug which should be fixed instead. By design `IntExt` should compare fine with ordinary `Int`, see https://github.com/thofma/Hecke.jl/blob/master/src/Misc/Infinity.jl. What was the error exactly?

src/QuadForm/Herm/LocallyIsometricSublattice.jl
```diff
@@ -147,47 +151,44 @@ function _locally_isometric_sublattice_odd_ramified(M, L, p, P, absolute_map)
...
 
   k, h = ResidueField(order(p), p)
   hext = extend(h, K)
   for j in length(C)-1:-1:1
     c = C[j]
     if all(!(scale(c, i) in [0, 1]) for i in 1:length(c))
-      @assert scale(C[1], 1) - valuation(scale(LL), P) >= 0
       s = div(scale(C[1], 1) - valuation(scale(LL), P), 2)
-      LL = lattice(ambient_space(LL), P^s * pseudo_matrix(LL))
+      LL = lattice(ambient_space(LL), _module_scale_ideal(P^s, pseudo_matrix(LL)))
```
thofma, line 166: Why?

src/QuadForm/Herm/LocallyIsometricSublattice.jl
```diff
@@ -147,47 +151,44 @@ function _locally_isometric_sublattice_odd_ramified(M, L, p, P, absolute_map)
...
       break
     end
     B, G, S = jordan_decomposition(LL, p)
     r = findfirst(i -> scale(c, i) == 1, 1:length(c))
     if r !== nothing
       r = rank(c, r)
       i = findfirst(j -> j == 1, S)
-      @assert i !== nothing
```
thofma, line 171: Why?

StevellM:
In the file `~test/QuadForm/Lattices.jl` in the testset `Misc`, there is this test 
``` 
@test valuation(u-1, p) == @inferred normic_defect(E, u, p)
```
so when I fixed `normic_defect` by changing `inf` into `return inf` (I guess this was a mistake and return was just missing), the test I mentioned fails. That's why all the tests failed in my previous commit. And it came from the fact that it could compare the type `Int64` of the valuation with the inferred type `IntExt` of the normic defect.

src/QuadForm/Herm/LocallyIsometricSublattice.jl
StevellM: This works only for fractional ideals, so for now it was easier than turning `P^s` into a fractional ideal, or create a new method for integral ideals

src/QuadForm/Herm/LocallyIsometricSublattice.jl
StevellM: I made it run over a bunch of examples and it seems that the codes never had any problem of assertion so I assumed that these were debug lines, and I just deleted them to lighten the codes. But I could put them back is you think we should keep them :)

src/QuadForm/Herm/LocallyIsometricSublattice.jl
thofma: Ah, right, thanks. I remember putting `fractional_ideal()` everywhere :(

src/QuadForm/Herm/LocallyIsometricSublattice.jl
thofma: Those are super cheap so I would be in favour of just keeping them.

StevellM:
Sure no problem. I will put back the assertion tests and modify those again.

src/QuadForm/Herm/LocallyIsometricSublattice.jl
thofma: Yes, that would be a good idea.

src/QuadForm/Herm/Genus.jl
```diff
@@ -1452,17 +1452,11 @@ function representative(G::GenusHerm)
   P = _non_norm_primes(G.LGS)
   E = base_field(G)
   V = hermitian_space(E, _hermitian_form_with_invariants(base_field(G), rank(G), P, G.signatures))
-  @vprint :Lattice 1 "Finding maximal integral lattice\n"
```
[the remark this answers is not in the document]
StevellM: Ah yes, I didn't how those things work and while trying to read through the function I took them off because it seemed to be really heavy in the code and not convenient. I forgot to put them back in my last commit...

PR #605, opened 2022-02-17
Make MapFromFunc safer and add documentation

thofma:


thofma:
@fieker The tests will fail since there are some dodgy things in the multivariate polynomial factorization code. But it is a bit difficult for me to figure out. Can you have a look?

fieker:
On Thu, Feb 17, 2022 at 01:05:08PM -0800, Tommy Hofmann wrote:
> @fieker The tests will fail since there are some dodgy things in the multivariate polynomial factorization code. But it is a bit difficult for me to figure out. Can you have a look?
I will - looks interesting though

thofma:
https://github.com/thofma/Hecke.jl/blob/c8bc8b6f07ba68bbc46c76651c40ec2a03b3f69f/src/NumField/NfAbs/MPolyAbsFact.jl#L453
The second argument is already an element of `S = F[[t]]`. Before my change, I think this was doing `setprecision(S(y), 1)` for an element `y` in `F`. So is this supposed to just be `setprecision(inv(...), 1)`.

PR #606, opened 2022-02-17
Sparse matrix bug fix

a-kulkarn:
There was a bug in swap_cols!. I also added a size function.

Quick question. Is there a reason this doesn't subtype MatElem?

### DIFF
--- a/src/Sparse/Matrix.jl
+++ b/src/Sparse/Matrix.jl
@@ -52,6 +52,7 @@ function nnz(A::SMat)
 end
 
 size(A::SMat) = (nrows(A), ncols(A))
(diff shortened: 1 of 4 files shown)


fieker:
On Thu, Feb 17, 2022 at 02:02:08PM -0800, a-kulkarn wrote:
> There was a bug in swap_cols!. I also added a size function.
Thanks.
> 
> Quick question. Is there a reason this doesn&#39;t subtype MatElem?
Do deliberate decision - however, we do not support the full
interface... e.g. you cannot do a[i,j] = 2. A different reasoning might
be that using the matrix interface is inefficient:
 - you do not want to iterate over size
 - products of SMat's are not supported (and would dense usually)
I'm happy to put it under MatElem, but what are you're applications?

PR #607, opened 2022-02-22
Fix `mass` for `HermLat` and allow `use_auto`

StevellM:
I have fixed the code for the mass of hermitian lattices and I have implemented the use of the automorphism group of definite lattices when it could be used. I have moved some tests to centralise as much as possible into the `~/test/QuadForm/Herm` directory. There were also few fixes to do in `~/Herm/Spaces.jl` that I have done. There is also more documentation where I could provide some.

### DIFF
--- a/src/QuadForm/Herm/Mass.jl
+++ b/src/QuadForm/Herm/Mass.jl
@@ -18,7 +18,6 @@ function _local_factor_dyadic(L::HermLat, p)
   @assert valscale >= 0
   val = div(valscale, 2)
   if !iszero(val)
-    println("line 21")
     s = elem_in_nf(uniformizer(p))
     L = rescale(L, s^(-val))
   end
@@ -44,7 +43,6 @@ function _local_factor_dyadic(L::HermLat, p)
   k1 = div(m1, 2)
 
   if isodd(m)
-    println("line 46")
     return fmpq(1, 2) * _gauss(k, k1, q^2)
   end
 
@@ -61,50 +59,39 @@ function _local_factor_dyadic(L::HermLat, p)
     h1 = m1 == 0 || G[end][4] == 2 * div(e + 1, 2)
     # L_p = H(0)^k0 \perp H(1)^k1
     if h0 && h1
-      println("line 64")
       t = iseven(e) ? k1 : k0
       return (q^t + 1) * lf
     end
 
     if !b
       if iseven(e)
-        println("line 71")
         return q^(m * (f2 - l) - k1) * (q^m1 - 1) * lf
       else
-        println("line 74")
         return q^(m * (f2 - l) + k0) * (q^m1 - 1) * lf
       end
     else
       if iseven(e)
-        println("line 79")
         return q^(m * (f2 - 1 - l) + k1) * (q^m0 - 1) * lf
       else
-        println("line 82")
         return q^(m * (f2 - l ) - k0)    * (q^m0 - 1) * lf
       end
     end
   else # non-hyperbolic case
     if isodd(e)
       if b && l == div(e - 1, 2)
-        println("line 89")
         return (q^k0 - 1) * lf
       elseif b
-        println("line 92")
         return q^(m * (f2 - l) - k0) * (q^m0 - 1) * lf
       else
-        println("line 95")
         return q^(m * (f2 - l) + k0) * (q^m1 - 1) * lf
       end
     end
 
     if b
-      println("line 101")
       return q^(m * (f2 - 1 - l) + k1) * (q^m0 - 1) * lf
     elseif l == div(e, 2) # e is even
-      println("line 104")
-      return q^(k1 - 1) * lf
+      return (q^k1 - 1) * lf
     else
-      println("line 107")
       return q^(m * (f2 - l) - k1) * (q^m1 - 1) * lf
     end
   end
@@ -132,14 +119,11 @@ function _local_factor_maximal(L::HermLat, p)
   if !islocal_norm(E, K(disc), p)
     q = norm(p)
     if ram
-      println("line 135")
       return fmpq(q^m - 1, 2*q + 2)
     else
-      println("line 138")
       return fmpq(q^m - (-1)^m, q + 1)
     end
   end
-  println("line 142")
   return fmpq(1)
 end
 
@@ -213,7 +197,14 @@ end
 #
 ################################################################################
 
+@doc Markdown.doc"""
+    local_factor(L::HermLat, p) -> fmpq
+
+Given a definite hermitian lattice $L$ and a bad prime $p$, return the local density
+of $L$ at $p$.
+"""
 function local_factor(L::HermLat, p)
+  @req isdefinite(L) "Lattice must be definite"
   S = base_ring(L)
   E = nf(S)
   K = base_field(E)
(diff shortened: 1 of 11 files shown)


src/QuadForm/Herm/Mass.jl
```diff
@@ -149,7 +149,7 @@ function _local_factor_generic(L::HermLat, p)
   ss = elem_in_nf(uniformizer(p))^(-val)
 
   if def
-    R = base_ring(base_ring(L))
+    R = maximal_order(K)
```
thofma, line 152: These two are not necessarily the same objects.

src/QuadForm/Herm/Mass.jl
```diff
@@ -199,14 +197,24 @@ end
...
 function local_factor(L::HermLat, p)
+  @req isdefinite(L) "Lattice must be definite"
   S = base_ring(L)
   E = nf(S)
   K = base_field(E)
   q = norm(p)
   lp = prime_decomposition(S, p)
   ram = length(lp) == 1 && lp[1][2] == 2
-  if ram && isdyadic(p)
+  if ram && iseven(q)
```
thofma, line 214: If you write it like this, can you put `# isdyadic(p)` as a comment? This is easier to parse.

src/QuadForm/Herm/Spaces.jl
```diff
@@ -5,26 +5,26 @@
...
-Create the Hermitian space over `K` with Gram matrix equal to the identity
-matrix. The matrix `G` must be square and Hermitian with respect to the non-trivial
-automorphism of `K`. The number field `K` must be a quadratic extension, that
-is, `degree(K) == 2` must hold.
+Create the Hermitian space over `E` with Gram matrix equals to `gram`. The matrix `gram` 
+must be square and Hermitian with respect to the non-trivial automorphism of `E`. 
+The number field `E` must be a quadratic extension, that is, `degree(E) == 2` must hold.
 """
 function hermitian_space(E::NumField, gram::MatElem; cached::Bool = true)
+  @req (E isa NfRel && degree(E) == 2) "E must be a relative number field of degree 2"
```
thofma, line 27: No, this is not necessary.
```julia
julia> K, = quadratic_field(-1)
(Imaginary quadratic field defined by x^2 + 1, sqrt(-1))

julia> V = hermitian_space(K, 2)
Hermitian space over
Imaginary quadratic field defined by x^2 + 1
with Gram matrix
[1 0; 0 1]

julia> V = hermitian_space(K, K[1 2; 2 1])
Hermitian space over
Imaginary quadratic field defined by x^2 + 1
with Gram matrix
[1 2; 2 1]
```

src/QuadForm/Herm/Spaces.jl
```diff
@@ -245,25 +254,30 @@ end
...
-    end
-  else
-    if iseven(div(rk, 2))
-      return false
-    else
-      return !islocal_norm(E, K(-1), p)
-    end
-  end
-end
+# function _islocally_hyperbolic_hermitian_detclass(rk, d, E, K, p)
```
thofma, line 257: Can you remove it if it is not used anymore?

src/QuadForm/Lattices.jl
```diff
@@ -1181,34 +1181,49 @@ end
 
 function maximal_sublattices(L::AbsLat, p; use_auto::Bool = false,
                                            callback = false, max = inf)
-  @req base_ring(L) == order(p) "asdsd"
```
thofma, line 1184: Is there a problem with "asdsd"? :D

src/QuadForm/Herm/Mass.jl
```diff
@@ -199,14 +197,24 @@ end
...
+    if ismaximal(L,p)[1]
+      return _local_factor_maximal(L, p)
```
thofma, line 216: Seems to be the only line that is not covered by the tests.

thofma (commented):
Nice work, thanks.

src/QuadForm/Herm/Mass.jl
StevellM: I thought they were, like in the cases in which I tried to test this.. I could put it back, I just wanted to make the definition `R` "lighter".

src/QuadForm/Herm/Mass.jl
StevellM: Sure, I will add it!

src/QuadForm/Herm/Spaces.jl
StevellM: Oh ok, my guess here was that we really want that the fixed field is of `NumField` type, so that we have to consider `QQ` has the trivial number field. So here we only need the degree to be 2 ?

src/QuadForm/Lattices.jl
StevellM: Not at all actually, I just put a clearer error message.. what is the meaning of "asdsd" by the way ? :)

src/QuadForm/Herm/Mass.jl
StevellM: The only new line yes, but more lines are not covered because the database and all the other examples I could find don't cover all the possible cases unfortunately... I added it in case it will be needed at some point.

StevellM:
Regarding Cho's recipe mentioned at some point in the code: I found the article related to this, but for now, for me, it seems a bit intricate to decrypt. Is it worth to try to implement it for hermitian lattices, since it is done for quadratic lattices, or the codes are enough for now ?

src/QuadForm/Herm/Mass.jl
thofma: I would prefer keeping the previous version.

src/QuadForm/Herm/Spaces.jl
thofma: Yes, I think we should restrict to degree `2` and this will be enough.

src/QuadForm/Lattices.jl
thofma: Just joking. I was too lazy to put a proper error message there.

src/QuadForm/Herm/Mass.jl
thofma: Ah, ok, makes sense.

thofma:
I mostly got inspiration from Markus' code. I guess @simonbrandhorst has to decide if there is a need for it :)

src/QuadForm/Lattices.jl
```diff
@@ -1181,34 +1181,49 @@ end
...
   R = base_ring(L)
   K = nf(R)
   k, h = ResidueField(R, p)
   hext = extend(h, K)
+  use_auto = isdefinite(L) ? use_auto : false
 
   if use_auto
-    throw(NotImplemented())
+    G = automorphism_group_generators(L)
+    Binv = inv(B)
```
simonbrandhorst, line 1196: are you sure that `B` is always square? For instance if the lattice is not full.
perhaps you rather need `can_solve_with_solution`

StevellM: This is what I wanted to do at first, but the outputs were wrong. So I chose to follow something already implemented for neighbours methods. I could try to manage it and see how to adapt it properly here to deal with the non-full cases

simonbrandhorst: If it is too much of a mess, set it as a `@req` and add a todo.

StevellM: I guess we should not allow the use of the automorphism group for non full lattices since it cannot compute the generators of the automorphism group

StevellM: No sorry I said nothing, we can make it by disallowing the `ambient_representation`.

PR #608, opened 2022-02-23
fixed ring for lattices

StevellM:
and a small fix on a function (it is not tested since I haven't found any example to cover it, yet).

### DIFF
--- a/test/QuadForm/Lattices.jl
+++ b/test/QuadForm/Lattices.jl
@@ -162,7 +162,7 @@
   @test L == dual(L)
   R = @inferred fixed_ring(L)
   @test R === base_ring(base_ring(L))
-  @test R.basis_nf == [K(1), a]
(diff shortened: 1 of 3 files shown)


test/QuadForm/Lattices.jl
```diff
@@ -160,6 +160,14 @@
   V = hermitian_space(E, gram_matrix(root_lattice(:E, 8)))
   L = lattice(V, pseudo_matrix(identity_matrix(E, 8)))
   @test L == dual(L)
+  R = @inferred fixed_ring(L)
+  @test R === base_ring(base_ring(L))
+  @test R.basis_nf == [K(1), a]
```
thofma, line 165:
```suggestion
  @test ismaximal(R)
```

thofma: (There is no guarantee how it will look like)

Issue #557, opened 2022-01-21
Meta lattice issue


thofma:
- [ ] Cache a pseudo HNF and improve equality work for number field lattices (#541, @thofma)
- [ ] For creation of lattices without ambient space we want to allow *either* provide basis (pseudo)-matrix (then the ambient space will have the identity as gram matrix) or provide a Gram matrix (via `gram = `) (then the lattice is free and the provided matrix is the Gram matrix of the ambient space). We retire the `gram_ambient_space` keyword. (@thofma)
- [ ] Make the quadratic (hermitian) lattice data base use the artifact system (@thofma)
- [ ] Test and document the databases (requires previous step).
- [ ] Move `LocMultGrpModSquMap` out of Misc to its "proper" place (@thofma).

StevellM:
For lattice constructors, in both hermitian and quadratic cases, an idea will be to allow the following constructors:

1.  If the ambient space `V` is fixed by the user, then one can:

-   Construct a lattice in this ambient space with a given pseudo matrix: `lattice(V::AbsSpace, B::PMat)`; 

- Construct a lattice in this ambient space with a given basis matrix: `lattice(V::AbsSpace, basis::MatElem)`;

-  Construct a lattice in this ambient space with a given set of generators: `lattice(V::AbsSpace, gens::Vector{Vector{}})`;

- Construct the trivial lattice in this ambient space: `lattice(V::AbsSpace)`.

2. If the ambient space is not specified, then one should mentioned the base algebra `F`, with the possibility to specify the gram matrix `gram` for the ambient space (which will be the identity if not specified, i.e. `gram = nothing`). This specification is necessary for the free lattice. Then one can:

-  Construct a hermitian/quadratic lattice over `F` with a given pseudo matrix: `hermitian/quadratic_lattice(F::Field, B::PMat ; gram = nothing)`;

- Construct a hermitian/quadratic lattice over `F` with a given basis matrix: `hermitian/quadratic_lattice(F::Field, basis::MatElem ; gram = nothing)`;

- Construct a hermitian/quadratic lattice over `F` with a given set of generators: `hermitian/quadratic_lattice(F::Field, gens::Vector{Vector{}} ; gram = nothing);

- Construct the free hermitian/quadratic lattice over `F` whose ambient space admits `gram` as gram matrix (in this case, the argument `gram` needs to be specified): `hermitian/quadratic_lattice(F::Field ; gram::MatElem)`.

simonbrandhorst:
[...]

PR #609, opened 2022-02-25
Change constructors for abstract lattices according to #557

StevellM:


StevellM:
With these new constructors, some `HermLat` and `QuadLat` functions in the respective `~/Types.jl` are not used anymore. Should we keep them in case it is needed one day ? My feeling is like they are not necessary now, and they seem not to be type stable, for what I have seen while creating the few tests I have implemented. What do yout think @thofma @simonbrandhorst ?

src/QuadForm/Herm/Lattices.jl
```diff
@@ -16,136 +16,103 @@ end
 #
 ################################################################################
 
+function lattice(V::HermSpace, B::PMat)
+  E = base_ring(V)
+  @req rank(matrix(B)) == min(nrows(B), ncols(B)) "B must be of full rank"
```
simonbrandhorst, line 21: This seems quite expensive. I think we should allow to turn this check of by using
`lattice(::HermSpace, B::PMat, check::Bool=True)`
Any other suggestion @thofma ?

src/QuadForm/Herm/Lattices.jl
```diff
@@ -16,136 +16,103 @@ end
...
+  @req nf(base_ring(B)) == E "Incompatible arguments: B must be defined over E"
+  @req ncols(B) == dim(V) "Incompatible arguments: the number of columns of B must be equal to the dimension of V" 
+  gram = gram_matrix(V, matrix(B))
```
simonbrandhorst, line 24: This is too expensive to be called during the construction. 
Also I seem to remember that L.gram stays undefined if we have an ambient space.

src/QuadForm/Herm/Lattices.jl
```diff
@@ -16,136 +16,103 @@ end
...
+
+Given a list of vectors `gens` and a number field `E` of degree 2, return the hermitian
+lattice spanned by the elements of `gens` inside the hermitian space over $E$ with 
+gram matrix $M$.
+
+If $M$ is not supplied, the gram matrix of the ambient space will be the identity
+matrix over $E$ of size the length of the elements of `gens`.
+"""
+function hermitian_lattice(E::NumField, gens::Vector ; gram = nothing) 
+  if length(gens) == 0
```
simonbrandhorst, line 82: This part does not seem to make sense. If the generators are of length zero, then the gram matrix must be a 0x0 matrix.

src/QuadForm/Herm/Lattices.jl
StevellM: Yes I was not sure about this point, it was there on the previous codes so I kept it in case

src/QuadForm/Herm/Lattices.jl
StevellM: This is the way used to create the zero lattice. And `gram` in these constructors always refers to the gram matrix of the ambient space, so one could create the zero lattice in a non trivial space, no ?

src/QuadForm/Herm/Lattices.jl
```diff
@@ -16,136 +16,103 @@ end
...
-  K = base_ring(V)
-  gram = gram_matrix(V, matrix(B))
-  z = HermLat{typeof(K), typeof(base_field(K)), typeof(gram), typeof(B), morphism_type(typeof(K))}()
-  z.pmat = B
-  z.gram = gram
-  z.base_algebra = base_ring(V)
-  z.involution = involution(V)
-  z.space = V
-  return z
+function hermitian_lattice(E::NumField ; gram::T) where T <: MatElem
```
simonbrandhorst, line 109:
```suggestion
function hermitian_lattice(E::NumField ; gram::MatElem)
```

src/QuadForm/IO.jl
```diff
@@ -23,24 +23,24 @@ function to_hecke(io::IO, L::QuadLat; target = "L", skip_field = false)
     pol = string(f)
     pol = replace(pol, string(var(parent(f))) => "x")
     println(io, "f = ", pol, ";")
-    println(io, "K, a = number_field(f)")
+    println(io, "K, a = number_field(f, \"a\", cached = false)")
```
simonbrandhorst, line 26: why change to `cached = false`?

src/QuadForm/Lattices.jl
```diff
@@ -333,6 +333,70 @@ function generators(L::AbsLat; minimal::Bool = false)
...
+###############################################################################
+#
+# Constructors
+#
+###############################################################################
+
+@doc Markdown.doc"""
+    lattice(V::AbsSpace, B::PMat) -> AbsLat
+
+Given an abstract space $V$ and a pseudo-matrix $B$, return the abstract lattice
```
simonbrandhorst, line 345:
```suggestion
Given an ambient space $V$ and a pseudo-matrix $B$, return the lattice
```

src/QuadForm/IO.jl
StevellM: I wanted to make the functions for quadratic and hermitian lattices consistent regarding each other, so I added this here as it is done for hermitian lattices

src/QuadForm/Lattices.jl
```diff
@@ -333,6 +333,70 @@ function generators(L::AbsLat; minimal::Bool = false)
...
+spanned by the pseudo-matrix $B$ inside $V$. If $V$ is hermitian (resp. quadratic)
+then the output is a hermitian (resp. quadratic) lattice.
+"""
+lattice(V::AbsSpace, B::PMat)
+
+@doc Markdown.doc"""
+    lattice(V::AbsSpace, basis::MatElem) -> AbsLat
+
+Given an abstract space $V$ and a matrix $basis$, return the abstract lattice
```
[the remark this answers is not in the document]
simonbrandhorst: Note that `$\Latex$` renders as latex in the documentation. When you refer to mathematics use latex. But when you refer to variable names, better use backticks.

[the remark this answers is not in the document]
StevellM: Good to know, thanks! I will revise all the documentations.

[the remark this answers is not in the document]
simonbrandhorst: Similar remarks apply below.

src/QuadForm/Lattices.jl
```diff
@@ -333,6 +333,70 @@ function generators(L::AbsLat; minimal::Bool = false)
...
+  pm = sub(pm, i:nrows(pm), 1:ncols(pm))
+  L = lattice(V, pm)
+  L.generators = gens
+  return L
+end
+
+@doc Markdown.doc"""
+    lattice(V::AbsSpace) -> AbsLat
+
+Given an abstract space $V$, return the abstract lattice with trivial basis
```
simonbrandhorst, line 394:
```suggestion
Given an abstract space $V$, return the abstract lattice with the standard basis
```

src/QuadForm/Herm/Lattices.jl
thofma: Eh, we can disable it with a check yes. Makes sense.

thofma:
If there is a accepting `V::AbsSpace` or `L::AbsLat`, how should `V` and `L` be referred to? I understand why @StevellM chose "abstract". I named the types in this way because they are the abstract types, but not sure we should refer to the mathematical objects in this way. I think I used to call them just "space" and "lattice". For the right definition of "hermitian" these are all hermitian, but this is too confusing I think :) Any thoughts?

src/QuadForm/Herm/Lattices.jl
simonbrandhorst: Of course, you are right.

simonbrandhorst:
I would call `AbsSpace` the ambient space and `L` a lattice in `V`.

StevellM:
The constructor for `ZLat`, I had conflicts with the functions when adding the check test for the rank of the matrices, because there are no such tests for Z-lattices. So I had to separate the constructors, and make sure that the creation for `ZLat` will never call the other ones. So I separate the case of `QuadSpace` over `QQ` and the other cases. For now that is the bast way I have found to solve this problem...

thofma:
Hm, I see, thanks for the explanation. I think it would be a bit more clearer to just give the signature
```
function lattice(V::QuadSpace{FlintRationalField, fmpq_mat}, B::MatElem; isbasis::Bool = true)
```
a `check` keyword argument (which is not used).

For consistency the constructors should all have the same behaviour. There is something else that probably needs to be adjusted in the future, but adding the `check` here should be fine for now.

thofma:
I have tested it locally, if you want I can just push it.

StevellM:
If it is not a problem for you, yes it could be nice :)

thofma:
Done :)

PR #614, opened 2022-03-04
Fix common_eigenspaces

joschmitt:
I needed common eigenspaces and noticed that Hecke has them (well hidden). Now they work again (`intersect_spaces` -> `_intersect_spaces`).
I decided that the default side for an eigenspace should be `:right` to be consistent with `kernel` (which is also an eigenspace after all).

### DIFF
--- a/src/AlgAssAbsOrd/Conjugacy/Husert.jl
+++ b/src/AlgAssAbsOrd/Conjugacy/Husert.jl
@@ -33,17 +33,17 @@ function _issimilar_husert_generic(A, B)
   k = 1
   for i in 1:s
     K, _ = NumberField(mus[i], "a", cached = false)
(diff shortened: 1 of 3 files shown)


thofma:
This is Carlo land, so you can do whatever you want as long as the tests pass

joschmitt:
Out of curiosity: how does this coverage get computed? My "diff coverage" is 100%, but overall it decreases?

thofma:
See the discussion here: https://github.com/thofma/Hecke.jl/pull/592

thofma:
@fieker Any comment since you looked at similar problems?

fieker:
but the right/left is a catastrophy. We never implemented the default left stuff

PR #617, opened 2022-03-08
Implement `fmpz` unit ranges

mgkurtz:
I like ranges 😊 Especially for test and demonstration purposes. So, I was puzzled, when the current stub produced wrong results (for step ranges with `step ≠ 0`) and `MethodError`s, so I fixed this for some the tests I came up with.

Though I did not want to copy half of Julia’s range implementation. So exotic use cases may still throw some `MethodError` but at least hopefully no wrong results. Also sadly `fmpz(1) .* (fmpz(1):fmpz(2)) isa StepRangeLen{fmpz, fmpz, fmpz, BigInt}` resulting in mixed `fmpz` and `BigInt` comparisons on further use. This could be fixed, but it’s probably not worth the effort, regarding that no one else seemed to miss ranges at all 😄

I guess, `fmpz` ranges could also be migrated to Nemo instead?

### DIFF
--- a/src/Misc/Integer.jl
+++ b/src/Misc/Integer.jl
@@ -187,9 +187,14 @@ end
 log(a::fmpz) = log(BigInt(a))
 log(a::fmpq) = log(numerator(a)) - log(denominator(a))
 
-function one(::Type{Nemo.fmpz})
-  return fmpz(1)
-end
+one(::Type{fmpz}) = fmpz(1)
+one(::fmpz) = fmpz(1)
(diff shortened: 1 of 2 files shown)


thofma:
Nice, thanks. It is fine if this lives in Hecke for the moment. There seems to be a failing test though.

mgkurtz:
Fine. Thanks for welcoming this feature.
Now the tests should hopefully also work with Julia 1.6. The `range` implementation in Julia has changed a little since then (to the better).

PR #625, opened 2022-03-09
cleanup of codes and docs

StevellM:
This is supposed to end the improvement of the test coverage for hermitian spaces/lattices and do a little preparation for the html documentation (c.f. #557 ). 
I have done some cleanups on some codes in the `~/QuadForm/Herm/` repository, and I started to develop more the documentation and making it more consistent. I focused mainly on what I want to treat in the html documentation at first (i.e. hermitian spaces/lattices) so the changes are mostly regarding functions about hermitian world. Nonetheless, I have modified also few things in `~/QuadForm/` and `~/QuadForm/Quad/` where I have found something to change. 
There is a final test for genus representatives of hermitian lattices supposed to catch a some part of the code not covered previously.
I am not strict on the documentation, I have tried to do as precised and clear possible but this can still be modified, of course.


src/QuadForm/Herm/Lattices.jl
```diff
@@ -704,10 +708,11 @@ function ismaximal_integral(L::HermLat)
 end
 
 @doc Markdown.doc"""
-    ismaximal(L::HermLat, p) -> Bool, HermLat
+    ismaximal(L::HermLat, p::NfOrdIdl) -> Bool, HermLat
 
-Decide if the norm of $L_p$ is integral and if the lattice is maximal. If the
-first condition holds but the second does not, then a proper overlattice with
+Given a hermtiian lattice `L` over $E/K$ and a prime ideal `p` of $O_K$, check
```
thofma, line 713:
```suggestion
Given a hermitian lattice `L` over $E/K$ and a prime ideal `p` of $O_K$, check 
```

thofma:
Thanks. This looks good!

StevellM:
Great. I have seen other typos I will have to correct. However, I don't understand that the tests fail: while running the `Hecke.test_module` function I had the same problems, but on the REPL the tests are ok.

thofma:
The tests are run with `Hecke.assertions(true)`.

PR #627, opened 2022-03-25
minor fixes for galois group over number field

fieker:
- inplace setprec! for qadics (or polys over qadic) was no longer
   working. (HenselCtxQadic)
   Unnotices as this is not used (any more) in factoring
   but is used in galois
 - princial_subfields in relative extension was broken

### DIFF
--- a/src/NumField/NfAbs/Elem.jl
+++ b/src/NumField/NfAbs/Elem.jl
@@ -845,9 +845,9 @@ function _height(a::nf_elem)
   return h
 end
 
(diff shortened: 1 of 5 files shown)


thofma:
Do you have the examples for the broken things?

fieker:
On Fri, Mar 25, 2022 at 07:35:56AM -0700, Tommy Hofmann wrote:
> Do you have the examples for the broken things?
Amy relative extension would have computed subfields, but not relative
ones. I will have to fix my PR anyhow

PR #630, opened 2022-03-28
Fix Hecke.test_module(..., false)

### LINKED ISSUE #629: test_module(.., false) broken
since rng will not be defined....

thofma:
Closes #629


fieker:
Great - it seems to work! Thanks!

PR #634, opened 2022-03-30
add nbits(::Integer) - some cases were missing

fieker:


thofma:
There seems to be a problem

src/Misc/Integer.jl
```diff
@@ -546,25 +546,14 @@ end
...
-function nbits(a::Int)
-  a==0 && return 0
-  return floor(Int, (log(abs(a))/log(2)))+1
-end
-
-function nbits(a::UInt)
-  a==0 && return 0
-  return floor(Int, (log(a)/log(2)))+1
+function nbits(a::Integer)
+  return ndigits(a, 2)
```
rfourquet, line 556:
```suggestion
  return ndigits(a, base=2)
```

fieker:
On Wed, Mar 30, 2022 at 04:27:21AM -0700, Rafael Fourquet wrote:
> @rfourquet commented on this pull request.
> 
> 
> 
> >  
>    Returns the number of bits necessary to represent $a$.
>  """
> -function nbits(a::Int)
> -  a==0 && return 0
> -  return floor(Int, (log(abs(a))/log(2)))+1
> -end
> -
> -function nbits(a::UInt)
> -  a==0 && return 0
> -  return floor(Int, (log(a)/log(2)))+1
> +function nbits(a::Integer)
> +  return ndigits(a, 2)
> 
> ```suggestion
>   return ndigits(a, base=2)
> ```
> 
indeed....

PR #635, opened 2022-03-31
Changes on html doc for hermitian forms

StevellM:
I haven't been able to create locally an experimental version of the webpage, so I am not sure about the final result. Yet, I have tried to use what have been done on the Oscar doc and some of the Hecke doc.
I have put many examples, for almost all the functions presented in the documentation. I am not strict on that: if it is too much I could remove/change some of them.

### DIFF
--- a/docs/Build.jl
+++ b/docs/Build.jl
@@ -24,6 +24,11 @@ pages = [
                        "orders/ideals.md",
                        "orders/frac_ideals.md"
                      ],
+	 "Quadratic and hermitian forms" => [ "quad_forms/introduction.md",
+					      "quad_forms/basics.md",
+					      "quad_forms/lattices.md",
+					      "quad_forms/genusherm.md"
+					    ],
          "abelian/introduction.md",
          "class_fields/intro.md",
          "Misc" =>  ["FacElem.md",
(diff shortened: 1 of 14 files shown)


thofma:
Does `Hecke.init_doc(); Hecke.build_doc(;format = :html)` work for you locally? I have a fix for the references in #637.

StevellM:
Yes it works, but this didn't open the browser, it just created some file I could read but from vim... So I don't have the proper display of anything which Latex-like or julia-like, just some brut code.

thofma:
Hm, it should open automatically, but it depends a bit on the system. Alternatively you can navigate to `Hecke/docs/site/index.html` and open it using your preferred browser. Does that work?

StevellM:
Ok finally I found out: constructing the doc automatically opens the browser... except if I work on a server of the uni (on which I did all the work while my laptop was being fixed). I tried on my laptop and it opens in the browser automatically, so I will be able to check everything there now!

thofma:
Eh, yes! There is not much we could do regarding displaying if you work remotely. If the server has a browser installed, you could open that remotely, but this is usually not a great experience.

StevellM:
Yes indeed, I have to stop relying on this. Well now I see everything that is not going well, so I will do many changes. However, I have seen (and I did the same accordingly) that we could do a 
```
```repl 2
blabla # hide
```.
```
and this would eventually hide the line before the "# hide", but this didn't work. Any idea ?

thofma:
I forgot to ask, but is this ready to go?

StevellM:
Yes on my side everything is ok.

Issue #639, opened 2022-04-05
Relative residue field vs. residue field of relative extensions


StevellM:
As far as I see, if we are given a relative extension of number fields, creating the residue field of the top field will create a field not over the residue field of the bottom field, but over some regular finite field. For instance:
```
julia> Qx, x = QQ["x"]
(Univariate Polynomial Ring in x over Rational Field, x)

julia> Qx, x = QQ["x"];

julia> K,a = number_field(x^4+x+1, "a");

julia> Kt, t = K["t"];

julia> E, b = number_field(t^2-a*t+1,"b");

julia> OK = maximal_order(K);

julia> OE = maximal_order(E);

julia> p = prime_decomposition(OK, 11)[1][1];

julia> P = prime_decomposition(OE,p)[1][1];

julia> FE, _ = ResidueField(OE, P);

julia> FK, _ = ResidueField(OK, p);

julia> FE
Finite field of degree 6 over F_11

julia> FK
Finite field of degree 3 over F_11

julia> defining_polynomial(FE)
x^6 + 3*x^4 + 4*x^3 + 6*x^2 + 7*x + 2
```
It would be interesting to have the version as a `RelFiniteField` since the type exists and has some functionalities. However, it doesn't seem to be used. I guess I have an idea on how to produce it, but I only know the defining polynomial if `p` is not an index divisor in `OE` : 

```
julia> isindex_divisor(OE, p)
false

julia> P.non_index_div_poly
y^2 + 10*$*y + 1
```
but in the general case I don't see how to produce it. The goal would be to use this in order to provide completion of an arbitrary tower of extensions of number fields (and thus this will be needed to create the unramified parts at every step).

thofma:
Hm, I thought Carlo used this in one form or the other for the local fields.

@joschmitt did you ever need the relative extensions of the residue fields for the algebras?

thofma:
P.S.: The easiest way that always works is just computing the embedding `FK -> FE` (via the maps named `_`) and compute the relative extension from there.

joschmitt:
> @joschmitt did you ever need the relative extensions of the residue fields for the algebras?

I'm not sure I understand the question. I don't think there was `RelFiniteField` "back then". I probably did the stuff in `Map/FqPolyRing.jl` which is an earlier (?) version of the relative extensions.
The code that's used for the quotients in the relative number fields should mainly be in `Map/NfRelOrd.jl`.

StevellM:
[...]

PR #643, opened 2022-04-08
add `relative_residue_field` for `NfRel`

StevellM:
This follows #639. I have adapted some similar codes written for absolute residue field. I had to modify some other methods, like the `(pre)image(::Map{C, D},a)` because for the map to residue fields, there are more than 2 parameters. I could only produce examples in tower of 3 extensions at most.

### DIFF
--- a/src/Map/Map.jl
+++ b/src/Map/Map.jl
@@ -25,8 +25,12 @@ end
 
 function preimage(M::Map{D, C}, a) where {D, C}
   if isdefined(M.header, :preimage)
-    p = M.header.preimage(a)::elem_type(D)
-    @assert parent(p) === domain(M)
+    if isdefined(M.header, :domain)
+      p = M.header.preimage(a)::elem_type(M.header.domain)
+      @assert parent(p) === domain(M)
+    else
+      error("No domain known")
+    end
     return p
   end
   error("No preimage function known")
@@ -35,7 +39,11 @@ end
 function image(M::Map{D, C}, a) where {D, C}
   if isdefined(M, :header)
     if isdefined(M.header, :image)
-      return M.header.image(a)::elem_type(C)
+      if isdefined(M.header, :codomain)
+        return M.header.image(a)::elem_type(M.header.codomain)
+      else
+        error("No codomain known")
+      end
     else
       error("No image function known")
     end
--- a/src/Map/NfRelOrd.jl
+++ b/src/Map/NfRelOrd.jl
@@ -160,14 +160,14 @@ function RelOrdQuoMap(O::T1, Q::RelOrdQuoRing{T1, T2, T3}) where { T1, T2, T3 }
 end
 
 
-mutable struct NfRelOrdToRelFinFieldMor{T, S, U, W} <: Map{NfRelOrd{T, S, U}, RelFinField, HeckeMap, NfRelOrdToRelFinFieldMor}
-  header::MapHeader{NfRelOrd{T, S, U}, RelFinField{W}}
+mutable struct NfRelOrdToRelFinFieldMor{S, T} <: Map{S, RelFinField{T}, HeckeMap, NfRelOrdToRelFinFieldMor}
+  header::MapHeader{S, RelFinField{T}}
   poly_of_the_field
-  P::NfRelOrdIdl{T, S, U}
+  P
   map_subfield::Union{NfOrdToFqMor, NfRelOrdToRelFinFieldMor}
 
-    function NfRelOrdToRelFinFieldMor{T, S, U, W}(O::NfRelOrd{T, S, U}, P::NfRelOrdIdl{T, S, U}, mapsub::NfOrdToFqMor) where {T, S, U, W <: fq}
-    z = new{T, S, U, W}()
+    function NfRelOrdToRelFinFieldMor{S, T}(O::S, P, mapsub::NfOrdToFqMor) where {S, T <: fq}
+    z = new{S, T}()
     z.P = P
     z.map_subfield = mapsub
     p = minimum(P, copy = false)
@@ -259,8 +259,8 @@ mutable struct NfRelOrdToRelFinFieldMor{T, S, U, W} <: Map{NfRelOrd{T, S, U}, Re
     return z
   end
 
-  function NfRelOrdToRelFinFieldMor{T, S, U, W}(O::NfRelOrd{T, S, U}, P::NfRelOrdIdl{T, S, U}, mapsub::NfRelOrdToRelFinFieldMor) where {T, S, U, W <: RelFinFieldElem}
-    z = new{T, S, U, W}()
+  function NfRelOrdToRelFinFieldMor{S, T}(O::S, P, mapsub::NfRelOrdToRelFinFieldMor) where {S, T <: RelFinFieldElem}
+    z = new{S, T}()
     z.P = P
     z.map_subfield = mapsub
     p = minimum(P, copy = false)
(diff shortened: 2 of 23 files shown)


src/NumFieldOrd/NfOrd/ResidueField.jl
```diff
@@ -210,3 +210,45 @@ function ResidueFieldSmallDegree1(O::NfOrd, P::NfOrdIdl)
...
+
+The map from `O` to the residue field is stored in `E` under the attribute 
+`:rel_residue_field_map`.
+"""
+function relative_residue_field(O::NfRelOrd{S, T, U}, P::NfRelOrdIdl{S, T, U}) where {S, T, U}
+  @req ismaximal(O) "O must be maximal"
+  @req order(P) === O "P must be an ideal of O"
+  E = nf(O)
+  K = base_field(E)
+  projK = get_attribute(K, :rel_residue_field_map)
```
thofma, line 232: I am not sure I understand why we want to store it on the base field of the number field. What about storing it on `P`? I

src/Map/NfRelOrd.jl
```diff
@@ -158,3 +158,251 @@ function RelOrdQuoMap(O::T1, Q::RelOrdQuoRing{T1, T2, T3}) where { T1, T2, T3 }
   S = elem_type(O)
   return RelOrdQuoMap{T1, T2, T3, S}(O, Q)
 end
+
+
+mutable struct NfRelOrdToRelFinFieldMor{T, S, U, W} <: Map{NfRelOrd{T, S, U}, RelFinField, HeckeMap, NfRelOrdToRelFinFieldMor}
```
thofma, line 163:
```suggestion
mutable struct NfRelOrdToRelFinFieldMor{T, S, U, W} <: Map{NfRelOrd{T, S, U}, RelFinField{W}, HeckeMap, NfRelOrdToRelFinFieldMor}
```

src/NumFieldOrd/NfOrd/ResidueField.jl
```diff
@@ -210,3 +210,45 @@ function ResidueFieldSmallDegree1(O::NfOrd, P::NfOrdIdl)
...
+    if !(K isa Hecke.NfRel)
+      _, projK = ResidueField(OK, p)
+      set_attribute!(K, :rel_residue_field_map, projK)
+    else
+      _, projK = relative_residue_field(OK, p)
+      set_attribute!(K, :rel_residue_field_map, projK)
+    end
+  end
+  FK = codomain(projK)
+  types = collect(typeof(O).parameters)
```
thofma, line 245: Hm, this is very unidiomatic julia.

src/NumFieldOrd/NfOrd/ResidueField.jl
StevellM: For now I made this choice because I wanted to use it for the completions of tower of extensions. So the idea would be to get the relative residue field of the top field, and get access to the other maps  by checking the successive base fields. But I am not close to any change/improvement :)

thofma:
Just a remark. There are not enough parameters to properly type all the fields. I would probably just use one parameter for the type of the domain and another for the domain. Then one would not have to do the `.parameters` shenanigans.

src/NumFieldOrd/NfOrd/ResidueField.jl
thofma: Yes, but the result here will just be wrong if `relative_residue_field(O, P)` is called for different `P`s? Or am I missing something.

src/NumFieldOrd/NfOrd/ResidueField.jl
StevellM: Yes I didn't know how to manage in the general case. Mainly, I had a lot of problems of coercions while writing the code, and I am not familiar with everything related to type conversion/promotion... So if you have any idea on how one can get rid of this, I could modify this.

src/NumFieldOrd/NfOrd/ResidueField.jl
StevellM: Ah yes indeed, I haven't thought about this... that's a very good point. So yes, it should be more reasonable to store it on the ideal.

StevellM:
Ok good, sounds to be a good compromise. And I could use the `ideal_type` method also for the type of the stored ideal.

StevellM:
Little problem with rebasing, all the changes requested are in "apply suggestions". Sorry for that.

src/NumFieldOrd/NfOrd/ResidueField.jl
```diff
@@ -210,3 +210,44 @@ function ResidueFieldSmallDegree1(O::NfOrd, P::NfOrdIdl)
...
+
+Given a maximal order `O` in a relative number field $E/K$ and a prime ideal 
+`P` of `O`, return the residue field $O/P$ seen as an extension of the (relative) 
+residue field of a maximal order in `K` at $minimum(P)$.
+
+Note that if `K` is a relative number field, the latter will also be seen as a 
+relative residue field.
+
+The map from `O` to the residue field is stored in `E` under the attribute 
+`:rel_residue_field_map`.
```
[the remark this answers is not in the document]
StevellM: Good to know, I will keep it in mind!

src/Map/Map.jl
```diff
@@ -25,8 +25,12 @@ end
 
 function preimage(M::Map{D, C}, a) where {D, C}
   if isdefined(M.header, :preimage)
-    p = M.header.preimage(a)::elem_type(D)
-    @assert parent(p) === domain(M)
+    if isdefined(M.header, :domain)
+      p = M.header.preimage(a)::elem_type(M.header.domain)
```
thofma, line 29: Can you elaborate why this change is necessary? It also leads to failing tests in Oscar.

Since we assert `parent(p) === domain(M)`, what is wrong with doing `::elem_type(D)`?

StevellM: I made the same changes as for `image`... the problem that was occurring is that for some elements, the type `Map` is provided with more than 2 parameters, so when we ask for `elem_type(C)` for `Map{D, C}`, this was crashing because `C` was a tuple of types, not only a type by itself. So I had to change it to call directly the (co)domain, and not the parameters which was not always a type (in my case)... Then I adapted a previous code which was using a type `Map` with 4 parameters, but I can restrict it to 2 and avoid the last 2 parameters if it really breaks everything..

thofma: This looks very wrong. For which `Map{D, C}` is `C` is not (a supertype) of the type of the codomain?

thofma: (The map interface requires that `D` or `C` should be (supertypes of) the type of domain/codomain. If a map implementation is violating this, then of course everything stops working. So it would be good to know which map implementations violate the interface.)

StevellM: At least the one I have added, and the one I have adapted. So in the file in which I have added the `NfRelOrdToRelFinFieldMor`. I can make a new PR where I change this and put back the old versions of `image` and `preimage` if you want

StevellM: Or I am mistaking with `MapHeader`... but as far as I see, in every files of `Map.jl`, every structures have as a supertype `Map` with 4 parameters.

thofma: (I am currently running the tests with the old version to see what goes boom)

StevellM: I can't reproduce the bug that made me make those changes... I may have fixed it without noticing..

StevellM: So I have a fix: I put back the old definitions of `image` and `preimage` and my codes/tests running (at least on my laptop)

thofma: Yes, `Map` has four parameters, but the first two have the meaning as I described above. If one writes `Map{C, D}`, then this is equivalent to `Map{C, D, <: Any, <: Any}`.

PR #645, opened 2022-04-12
Rename `isisomorphic` -> `is_isomorphic`, add `is_isomorphic_with_map`

fingolfin:
The function `is_isomorphic` just returns a `Bool`, while `is_isomorphic_with_map` also returns a map.

This brings us in line with what we do in Oscar now (modulo the underscore after `is`, but that's temporary). Also, before this PR, some `isisomorphic` methods had one return value, some had two, which was quite confusing.

Breaking change.

Contains PR #682 which ideally should be merged first.

### DIFF
--- a/docs/src/number_fields/fields.md
+++ b/docs/src/number_fields/fields.md
@@ -116,6 +116,7 @@ simplified_simple_extension(::NonSimpleNumField)
 
 ```@docs
 is_isomorphic(::SimpleNumField, ::SimpleNumField)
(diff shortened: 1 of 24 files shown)


thofma:
I merged #682 first, was this correct? I am bit confused because of the conflicts.

fingolfin:
Yeah I had to rebase this due to the other PR having been squashed. No problem :-)

fingolfin:
This now passes tests (well, except for the Julia nightly test on macOS which is close to running two hours now, but whatever is wrong there this time (*sigh*) should not be related to this PR.

PR #653, opened 2022-04-19
Adding basic functionality for elliptic curves. Generalized division …

JHanselman:
…polynomials, rank computations, isogenies, height functions


src/EllCrv/EllCrv.jl
```diff
@@ -712,38 +703,125 @@ end
 #  Division polynomials
 #
 ################################################################################
-# TODO:
-# - Adjust to Sage/Magma
-# - Implement for long form, see Mazur, Tate, "p-adic sigma function"
-
+"""
+Computes the n-th division polynomial of an elliptic curve defined over a field k following Mazur and Tate. The two-torsion multiplicity flag determines the output.
```
thofma, line 707: The docstrings need some fixes, see https://docs.julialang.org/en/v1/manual/documentation/.

1. Should be imperative ("compute" instead of "computes"). I know we are inconsistent, but we are trying to adjust :)
2. The first line should look like
```
    division_polynomial(E::EllCrv, n::Int, two_torsion_multiplicity::Int, x, y) -> Poly
```
3. Lines should not be longer than 80 characters.
```

File: src/EllCrv/EllCrv.jl [hunk 2]
```diff
@@ -712,38 +703,125 @@ end
...
-# - Adjust to Sage/Magma
-# - Implement for long form, see Mazur, Tate, "p-adic sigma function"
-
+"""
+Computes the n-th division polynomial of an elliptic curve defined over a field k following Mazur and Tate. The two-torsion multiplicity flag determines the output. 
+When it is 0 the n-th division polynomial (as a univariate polynomial) without the 2-torsion factor is returned. (i.e. the two-torsion factor is 1).
+When it is 1 the n-th division polynomial (as a multivariate polynomial) is returned (i.e. the two-torsion factor is 2*y + a1*x + a3).
+When it is 2 the n-th division polynomial is returned as a univariate polynomial. (i.e. the two-torsion factor is 4*x^3+b2*x^2+2*b4*x+b6 = (2*y + a1*x + a3)^2).
+"""
+function division_polynomial(E::EllCrv, n::Int, two_torsion_multiplicity::Int, x = PolynomialRing(base_field(E),"x")[2], y = PolynomialRing(parent(x),"y")[2])
```
thofma, line 712: It is rather unidiomatic to make the return type depend on some parameter. It results in rather slow code in every function that calls `division_polynomial`. It would be better if the return type would always be the same, i.e., always univariate or always multivariate. Is there a way to make this possible? I noticed that Magma returns more than one object, but all of them are univariate.

src/EllCrv/Heights.jl
```diff
@@ -0,0 +1,244 @@
...
+#  Naive Height 
+#
+################################################################################
+
+function naive_height(P::EllCrvPt{fmpq}, prec = 100)
+  R = Nemo.RealField(2*prec)
+  x = P.coordx
+  p = numerator(x)
+  q = denominator(x)
+  return log(R(max(abs(p),abs(q))))
```
thofma, line 49: Shouldn't we check wether the result has radius less than `2^-prec`?

src/EllCrv/Heights.jl
```diff
@@ -0,0 +1,244 @@
...
+    else
+      mu = mu+f*log(abs(zw))
+      t = w/zw
+      beta = 1-beta
+    end
+  end
+  if isfinite(mu) & (radius(mu)<R(10)^(-d))
+    # Algorithm is only precise up to d decimals
+    add_error!(mu,R(10)^(-d))
+    return(mu)
```
thofma, line 198:
```suggestion
    return mu
```

src/EllCrv/Isogeny.jl
```diff
@@ -0,0 +1,226 @@
...
+  if (0 != degree(psi_G))
+    psi_quo = numerator((psi//psi_G))
+      if (degree(psi_quo)!=0)
+          throw(DomainError(psi, "In order to use Kohel's algorithm, a kernel of even degree needs to be contained in the two torsion."))
+      end
+    phi, omega, v, w, n, d = even_kernel_polynomial(E,psi)
+  else
+    phi, omega, v, w, n, d = odd_kernel_polynomial(E,psi)
+  end
+ Ry = parent(omega)
```
thofma, line 84: Indentation

thofma (commented):
I have only looked at the elliptic curve things. This looks good. I left some minor comments here and there.

What I did not find where tests for the new functions. It would be great if we could add tests, see `Hecke/test/...` where to put them (see also http://www.thofma.com/Hecke.jl/dev/dev/test/).

thofma:
P.S.: I did not mark all places with indentation problems (there were too many and it is tedious to do in the web interface).

src/EllCrv/EllCrv.jl
JHanselman: I was also wondering about what the best way would be to do this. The thing you usual call a division polynomial is multivariate. But the only time the y comes in, is when n is even because of the 2-torsion factor. (I think the multivariate version is also currently used by the algorithms the students implemented)

But there are situations where you want the other two possibilities: univariate without 2-torsion or univariate with square of two-torsion. These are the two things Magma returns in DivisionPolynomial(E,n).

In Magma the way to get the usual division polynomial is to multiply the one without the two-torsion factor with the multivariate 2-torsion factor that gets returned by TwoTorsionPolynomial(E), but I don't really like that either. (I believe the thing that should be returned is the one you find on wikipedia. A user shouldn't have to figure out it needs to be multiplied by something else).

The way I implemented it right now is the way it was done in Sage. 

Maybe I should split it into two functions? division_polynomial_uv (for univariate) and division_polynomial for the usual one?

I was also wondering if the split for long and short weierstrass form is even that much faster. (The formulas are essentially the same, the short version simply already assumes a_1,..., a3 are 0)

thofma: Yes, I agree with what you say and I like the idea of having `division_polynomial_univariate` (we nowadays try not to use too many non-standard abbreviations in function names).

Regarding the split, I have no opinion on that one. Your call :)

src/EllCrv/Heights.jl
```diff
@@ -0,0 +1,244 @@
...
+        M = min(valuation(B, p), N//2)
+        L = M*(M - N)//N
+      end
+    elseif ( C == 0 || ( C != 0 && B != 0 && valuation(C, p) >= 3*valuation(B, p)))
+      L = -2*valuation(B, p)//3
+    else
+      L = -valuation(C, p)//4
+    end
+    result = L*log(R(p))
+    if radius(result)<R(10)^(-d)
```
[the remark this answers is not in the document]
JHanselman: Ok, thanks!

JHanselman:
> I have only looked at the elliptic curve things. This looks good. I left some minor comments here and there.
> 
> What I did not find where tests for the new functions. It would be great if we could add tests, see `Hecke/test/...` where to put them (see also http://www.thofma.com/Hecke.jl/dev/dev/test/).

The old tests need to be changed as well. I know everything still works, but the constructors got changed, which breaks a whole bunch of tests.

I'm not quite sure how to write good test files though. Especially for 2_torsion_rank where the output can still seem correct even though internally all kinds of things could've gone wrong during the computation of the Selmer group. 

How many tests should I roughly have per functionality?

thofma:
Ideally, every function is called and every branch triggered. With the branches it can be a bit difficult, so I would say try to run all functions once.

Once the tests run successfully here, there will be an overview of which lines where called and which not.

JHanselman:
> I sometimes do computations with elliptic surfaces. They can be seen as elliptic curves over 1-dimensional function fields e.g. `Q[t]` or `GF(p)[t]`. Does any of the elliptic curve code work for function fields too? If there is some generic code, I would be grateful for some tests with elliptic curves over function fields to make sure it runs.

I think that in principle everything in EllCrv.jl and Isogeny.jl should work for elliptic curves over function fields. I could add some tests for that. But that's also only really basic stuff. I do plan to generalize some of the existing functions for minimal_models, etc., that only work over QQ right now to number fields. I'm guessing that'll probably translate more or less automatically to function fields.

src/EllCrv/EllCrv.jl
```diff
@@ -712,50 +776,128 @@ end
...
-
-# computes the n-th division polynomial psi_n in ZZ[x,y] for a given elliptic curve E over ZZ
-function division_polynomialE(E::EllCrv, n::Int, x = nothing, y = nothing)
+"""
+    division_polynomial(E::EllCrv, n::Int, two_torsion_multiplicity::Int, x, y)
+      -> Poly
+
+Compute the n-th univariate division polynomial of an elliptic curve defined 
+over a field k following Mazur and Tate. When x and or y are given the output is 
+automatically evaluated using the given values.
```
thofma, line 785:
```suggestion
    division_polynomial_univariate(E::EllCrv, n::Int, [x]) -> Poly

Compute the n-th univariate division polynomial of an elliptic curve defined 
over a field k following Mazur and Tate. By default the result is a univariate polynomial over the base ring of `E`.
When `x` is given, the output is evaluated using the given value for `x`.
```

thofma, line 785:
```suggestion
    division_polynomial_univariate(E::EllCrv, n::Int, [x]) -> Poly

Compute the n-th univariate division polynomial of an elliptic curve defined 
over a field k following Mazur and Tate. By default the result is a univariate polynomial over the base ring of `E`.
When `x` is given, the output is evaluated using the given value for `x`.
```

src/EllCrv/EllCrv.jl
```diff
@@ -712,50 +776,128 @@ end
...
+        twotorsfactor = one(R)
+      end
+  end
+  return twotorsfactor*poly, poly, twotorsfactor 
+end
 
-  A = numerator(E.coeff[1])
-  B = numerator(E.coeff[2])
+"""
+division_polynomial(E::EllCrv, n::Int, x, y) -> Poly
```
thofma, line 818:
```suggestion
    division_polynomial(E::EllCrv, n::Int, x, y) -> Poly
```

thofma, line 818:
```suggestion
    division_polynomial(E::EllCrv, n::Int, x, y) -> Poly
```

thofma:
I pushed some fixes and this is look ok. If you are happy I will merge this.

JHanselman:
> I pushed some fixes and this is look ok. If you are happy with it I will merge this.

I'm fine with it. I only wanted to add some more tests to increase coverage, but I can't look at the Codecov report anymore.

PR #662, opened 2022-04-29
Some improvments for the Mordell-Weil group things

thofma:
- Speedup the computation in the hot loop using inplace operations
- Avoid rational arithmetic (everything is integral)

@JHanselman Can you have a look and confirm that this makes sense?

P.S.: For `E1` the time goes down from 15s to 0.4s on my machine.


JHanselman:
> * Speedup the computation in the hot loop using inplace operations
> 
>     * Avoid rational arithmetic (everything is integral)
> 
> 
> @JHanselman Can you have a look and confirm that this makes sense?
> 
> P.S.: For `E1` the time goes down from 15s to 0.4s on my machine.

It looks good to me!

PR #668, opened 2022-05-02
Minor fixes for orders and ideals

thofma:


src/NumField/NfAbs/NfAbs.jl
```diff
@@ -1174,7 +1174,7 @@ function find_all_super(A::NumField, filter::Function = x->true)
   ls = length(s)
   filter!(x->x.value !== nothing, s)
   if length(s) < ls #pruning old superfields
-    set_attribute!(A, :sub_of)
+    set_attribute!(A, :sub_of => s)
```
thofma, line 1177: @fieker This is correct, right? I do not have an example for it, so would be good if you could confirm.

PR #669, opened 2022-05-02
Fix short[test]_vectors for Zlattice

thofma:


src/QuadForm/Morphism.jl
```diff
@@ -604,31 +604,35 @@ end
     short_vectors(L, ub) -> Vector{Tuple{Vector{Int}, fmpq}}
 
 Returns all tuples `(v, n)` such that `v G v^t = n <= ub`, where `G` is the
-Gram matrix of `L` and v is non-zero.
```
simonbrandhorst, line 607: why do you remove this? We do not list the zero vector even though it fits the eqn.

thofma: Some git problem on my part.

PR #671, opened 2022-05-02
catch segmentation fault for short_vectors of the trivial lattice

simonbrandhorst:


thofma:
#669 ?

thofma:
Magma throws an error, so I am doing the same.

simonbrandhorst:
Oh right. Didn't notice. 
I would like to allow the trivial lattice. It is a useful corner case  to me. Then I do not need catch it manually later in my code.

PR #680, opened 2022-05-05
add shortcut hyperbolic plane

StevellM:


src/QuadForm/Quad/ZLattices.jl
```diff
@@ -564,6 +564,41 @@ function _root_lattice_E(n::Int)
...
+#
+#  Hyperbolic plane
+#
+################################################################################
+
+@doc Markdown.doc"""
+    Zlattice(R::Symbol, n::Union{Int64, fmpz} = 1) -> Zlat
+
+Given `R = :H` or `R = :U`, return the hyperbolic plane with intersection form
+scaled by `n`.
```
thofma, line 577:
```suggestion
scaled by `n`, that is, with scale `n`.
```
I had to look up what intersection form" is :)

src/QuadForm/Quad/ZLattices.jl
```diff
@@ -564,6 +564,41 @@ function _root_lattice_E(n::Int)
...
+################################################################################
+#
+#  Hyperbolic plane
+#
+################################################################################
+
+@doc Markdown.doc"""
+    Zlattice(R::Symbol, n::Union{Int64, fmpz} = 1) -> Zlat
+
+Given `R = :H` or `R = :U`, return the hyperbolic plane with intersection form
```
fingolfin, line 576: Looking at this, I find myself quite puzled:
- I've never heard of a lattice being referred to as a "plane", hyperbolic or not. Is this a typical notation somewhere (and where? reference?). Definitely where I come from.
- I think giving a bit more information what is meant by that would be useful
- Why is the argument called `R`?
- And why can one use both `R = :H` and `R = :U`? And what does `U` stand for?
- why is this a `Zlattice` method, as opposed to its own (perhaps more easily discovered) method like `hyperbolic_plane_lattice`?

src/QuadForm/Quad/ZLattices.jl
StevellM: Yes sorry, since I called it "hyperbolic plane", I used intersection form instead of gram matrix, mainly because I didn't know how to name this matrix, neither how to print it. That's also why I put examples in the docstring to make it clear. :)

src/QuadForm/Quad/ZLattices.jl
StevellM: - It was a choice of naming here. It is the unique even unimodular hyperbolic Z-lattice of rank 2 (up to isometry), and you can find it called "hyperbolic plane" . This choice was mainly because I didn't know a conventional/canonical name for the gram matrix , neither how to make a "nice" diplay of it.  That's also why I put the examples. I could change the docstring by refering to it as said above (if this makes more sense);
- same as above
- I just used the same idea as for `root_lattice` where the argument of type `Symbol` is called `R`. No deep meaning there.
- There is not a canonical symbol: some people use `H`, others use `U`. :) 
- it seemed to be a good compromise, in the same way we work with root lattices: they are usually represented with a symbol, so it would have made sense to add a method using the symbol(s) (even though it is not always consistent). Of course, if not too long as a name, I can add a method `hyperbolic_plane_lattice(ZZ, n)` for instance, with the appropriate docstring.

thofma: > * I've never heard of a lattice being referred to as a "plane", hyperbolic or not. Is this a typical notation somewhere (and where? reference?). Definitely where I come from.

This is standard terminology when working with integral lattices. A quick google search confirms that Stevell did not make this up.

PR #682, opened 2022-05-06
Adust to is->is_ renaming in AA and Nemo

fingolfin:
- Update to new Nemo and AbstractAlgebra
- Apply AA mass rename
- Adjust to renaming
- Apply Nemo mass rename

This does not yet attempt to rename further things -- I simply thought I should get this part over, as I was able to do it quickly (thanks to the scripts). Also, it needs Nemo 0.31.0 which is not yet released.


fingolfin:
@thofma I think it would make sense to merge this now already, so that the next version of Hecke definitely works with the new AA / Nemo. If we get the other renaming done before, even better. (I've also updated PR #645 and rebased it on this PR here)

PR #686, opened 2022-05-06
Support for number fields, expanded isogenies, isomorphisms

JHanselman:
Code hasn't been tested rigorously yet and tests+docstrings still need to be written.
Just for homomorphism from group to automorphisms.


src/EllCrv/Misc.jl
```diff
@@ -211,6 +237,33 @@ function nrootscubic(b, c, d, p)
...
+      return FlintZZ(1)
+    else
+      return FlintZZ(3) #one double and one single root
+    end
+  else
+    return FlintZZ(3)
+  end
+end
+
 function smod(a::T, b::S) where {T, S}
```
fieker, line 267: This exists mostly as mod_sym or mod_sym!

src/EllCrv/Misc.jl
```diff
@@ -219,6 +272,29 @@ function smod(a::T, b::S) where {T, S}
   return z
 end
 
+function normal_elem(K::T, L::T) where T<:FinField
```
fieker, line 275: As we had to look it up: please provide a comment/ doc string to say what a normal_elem actually is

src/EllCrv/Models.jl
```diff
@@ -0,0 +1,278 @@
...
+"""
+function short_weierstrass_model(E::EllCrv)
+  return _short_weierstrass_model(E)
+end
+
+function _short_weierstrass_model(E::EllCrv{T}) where T
+  R = base_field(E)
+  p = characteristic(R)
+
+  if (p == 2) || (p == 3)
```
fieker, line 63: I can wrtite down a short equatino in char 3 I think - I cannot transform all of them, but they do exist. In char 2 I fear that a short equation cannot be smooth...

src/EllCrv/Models.jl
```diff
@@ -0,0 +1,278 @@
...
+    ynew = divexact(R.coordy, mue^3)
+    S = E([xnew, ynew])
+    return S
+  end
+
+  return E_int::EllCrv{fmpq}, trafo_int, trafo_rat
+end
+
+@doc Markdown.doc"""
+    integral_model(E::EllCrv{fmpq}) -> (F::EllCrv{nf_elem}, function, function)
```
fieker, line 254: the input here is stated as fmpq, the text allows also nf_elem. Further are the 2nd and 3rd return value really julia functios or are they maps? Ie. can I ask for domain/ codomain....

src/EllCrv/Models.jl
JHanselman: The code for short Weierstrass form was written by students and I simply moved it around. I wrote simplified_model to transform curves of characteristic 2 and 3 into the Weierstrass models that are usually used there. I should probably change that error message and take a look at what Magma does when you try to compute a short Weierstrass form in characteristic 3.

src/EllCrv/Models.jl
JHanselman: They are probably EllCrvIsos, which are Maps. I probably copy-pasted the old docstring for integral_model and forgot to change it.

src/EllCrv/Misc.jl
JHanselman: I'll add it.

fieker:
On Mon, May 09, 2022 at 02:03:54AM -0700, JHanselman wrote:
> @JHanselman commented on this pull request.
> 
> 
> 
> > +      (EE::EllCrv, function(EllCrvPt), function(EllCrvPt))
> +
> +Transform a curve given in long Weierstrass form over QQ to short Weierstrass
> +form. Return short form and both transformations for points on the curve;
> +first transformation from E (long form) to EE (short form), 
> +second transformation is the inverse of this map.
> +"""
> +function short_weierstrass_model(E::EllCrv)
> +  return _short_weierstrass_model(E)
> +end
> +
> +function _short_weierstrass_model(E::EllCrv{T}) where T
> +  R = base_field(E)
> +  p = characteristic(R)
> +
> +  if (p == 2) || (p == 3)
> 
> The code for short Weierstrass form was written by students and I simply moved it around. I wrote simplified_model to transform curves of characteristic 2 and 3 into the Weierstrass models that are usually used there. I should probably change that error message and take a look at what Magma does when you try to compute a short Weierstrass form in characteristic 3.
> 
The point was mainly about the possiblilty that you already HAVE a short
WS form...

JHanselman:
> On Mon, May 09, 2022 at 02:03:54AM -0700, JHanselman wrote: @JHanselman commented on this pull request. > + (EE::EllCrv, function(EllCrvPt), function(EllCrvPt)) + +Transform a curve given in long Weierstrass form over QQ to short Weierstrass +form. Return short form and both transformations for points on the curve; +first transformation from E (long form) to EE (short form), +second transformation is the inverse of this map. +""" +function short_weierstrass_model(E::EllCrv) + return _short_weierstrass_model(E) +end + +function _short_weierstrass_model(E::EllCrv{T}) where T + R = base_field(E) + p = characteristic(R) + + if (p == 2) || (p == 3) The code for short Weierstrass form was written by students and I simply moved it around. I wrote simplified_model to transform curves of characteristic 2 and 3 into the Weierstrass models that are usually used there. I should probably change that error message and take a look at what Magma does when you try to compute a short Weierstrass form in characteristic 3.
> The point was mainly about the possiblilty that you already HAVE a short WS form...
> […](#)
> -- Reply to this email directly or view it on GitHub: [#686 (comment)](https://github.com/thofma/Hecke.jl/pull/686#discussion_r867791851) You are receiving this because you commented. Message ID: ***@***.***>

I know. But this can only happen when char = 3 and j=0, i.e. a case that is currently covered in simplified_model. I only wanted to check how Magma handles the case where a supersingular curve with char 3 is given in non-short Weierstrass form. short_weierstrass_model should probably not give an error message in this case and simply perform the transformation. And only give an error message in the other cases.

src/EllCrv/Torsion.jl
```diff
@@ -564,7 +571,7 @@ function torsion_structure(E::EllCrv{nf_elem})
     structure = [fmpz(k1), fmpz(k2)]
     gens = [T1, T2]
   end
-  return [structure, gens]
+  return Vector{}[structure, gens]
```
thofma, line 574:
```suggestion
  return structure, gens
```
You can just return a tuple.

PR #690, opened 2022-05-10
Rename isFOO -> is_FOO

fingolfin:
Completely untested, just wanted to push this before I have to leave, will fix it up later.


docs/src/abelian/introduction.md
```diff
@@ -57,13 +57,13 @@ ngens(A::GrpAbFinGen)
 nrels(G::GrpAbFinGen)
 rels(A::GrpAbFinGen)
 isfinite(A::GrpAbFinGen)
-isinfinite(A::GrpAbFinGen)
+is_infinite(A::GrpAbFinGen)
 rank(A::GrpAbFinGen)
 order(A::GrpAbFinGen)
 exponent(A::GrpAbFinGen)
 istrivial(A::GrpAbFinGen)
```
fingolfin, line 64: Note: as a byproduct of how my renaming scripts work, in this PR, `ishermitian ` and `istrivial` were left untouched, I just added `@alias is_hermitian ishermitian` and `@alias is_trivial istrivial`, because they are imported from `LinearAlgebra`. resp. `GroupsCore`.

But I guess it would be nicer to change this here, so that users consistently see the `is_` variants in the manual. But then these docstrings must exist, suggesting that we also change it in the code. But *then* we must be careful were to declare those `@alias`es.

Also note that the same affects a few more function from AA and Nemo:
```julia
# AA:
@alias is_one isone
@alias is_zero iszero

# Nemo:
@alias is_equal isequal
@alias is_finite isfinite
@alias is_inf isinf
@alias is_integer isinteger
@alias is_less isless
@alias is_real isreal
```

So, how do we want to deal with all of these? Shall I adjust the manual and code to use the `is_` variants? For all? What about `is_one` / `is_zero`, those seem common enough to possibly cause confusion on their own.

fingolfin: CC @thofma @fieker 

Whichever way we go, probably a good idea to first merge this PR (once CI passes)

fingolfin: From my POV all of these could just say as they are, with perhaps the exception of `isfinite` -- because that is in various places listed next to e.g. `is_trivial`, which is a bit jarring. So listing it instead as `is_finite` might be nicer. But the user should still be able to learn that this is an alias for `isfinite`. Hrm.

thofma: I have no strong opinion here.

thofma: Should this be discussed somewhere else?

PR #701, opened 2022-05-19
IndexCalculus

BD-00:
IndexCalculus for DLP over safeprime fields.
Wiedemann version for more general fields in progress


src/Sparse/IndexCalculus.jl
```diff
@@ -0,0 +1,283 @@
...
+
+function primitive_elem(F::FinField,first::Bool) 
+    p = length(F)
+    Fact = prime_divisors(fmpz(p-1))
+    while true # alpha exists
+        for y in F
+            bool = true
+            if !first y = rand(F) end
+            if isprime(lift(y))
+                if !(one(F) in [y^(div(fmpz(p-1),i)) for i in Fact])
```
fieker, line 19: `if any(i->isone(y^divexact(fmpz(p-1),i)), Fact)`

src/Sparse/Matrix.jl
```diff
@@ -952,6 +952,162 @@ function Base.insert!(A::SMat{T}, i::Integer, B::SRow{T}) where T
...
+  end
+  return A
+end
+
+@doc Markdown.doc"""
+  delete_zero_rows!(A::SMat{T}, s=1)
+
+Deletes zero rows of $A$ in place.
+"""
+function delete_zero_rows!(A::SMat{T}, s=1) where T #where s denotes the first column where we wanna start
```
fieker, line 994: the removal starts/ stops at ROW s, not column s

src/Sparse/Matrix.jl
```diff
@@ -952,6 +952,162 @@ function Base.insert!(A::SMat{T}, i::Integer, B::SRow{T}) where T
...
+end
+
+@doc Markdown.doc"""
+  empty_col!(A::SMat{T}, TA::SMat{T}, j::Int) -> SMat{T}
+
+Deletes non-zero entries in $j$-th column of $A$ in place.
+"""
+function empty_col!(A::SMat{T}, TA::SMat{T}, j::Int) where T #only deletes entries in column j, output same size as input
+  for row in TA[j].pos 
+      i = findfirst(isequal(j), A[row].pos)
```
fieker, line 1010: `i` can be nothing (if the column does not occur in this row), then this will crash (and the nnz count is wrong)

src/Sparse/Matrix.jl
```diff
@@ -952,6 +952,162 @@ function Base.insert!(A::SMat{T}, i::Integer, B::SRow{T}) where T
...
+#
+#  Row/Col operations in matrix
+#
+################################################################################
+@doc Markdown.doc"""
+  function scale_col!(A::SMat{T}, j::Int, c::T) -> SMat{T}
+
+Returns $A$ after scaling j-th column in place using transposed matrix.
+"""
+function scale_col!(A, TA, j, c) #A[_j]->c*A[_,j]
```
fieker, line 1039: docstring does not match signature

src/Sparse/Matrix.jl
```diff
@@ -952,6 +952,162 @@ function Base.insert!(A::SMat{T}, i::Integer, B::SRow{T}) where T
...
+    add_scaled_col!(A::SMat{T}, i::Int, j::Int, c::T) -> SMat{T}
+
+As add_scaled_row!(A::SMat{T}, i::Int, j::Int, c::T) but with columns of $A$.
+"""
+function add_scaled_col!(A::SMat{T}, i::Int, j::Int, c::T) where T 
+  @assert c != 0
+  @assert 1 <= i <= ncols(A) && 1 <= j <= ncols(A)  
+  for r in A.rows
+    if i in r.pos
+      i_i = findfirst(isequal(i), r.pos) #changed
```
fieker, line 1069: will return nothing if i is not in r.pos. Which iis fine. But the assignement in the next line will fail

src/Sparse/Preprocessing.jl
```diff
@@ -0,0 +1,220 @@
...
+
+
+
+###############################################################################
+#
+#   AUXILIARY STUFF
+#
+###############################################################################
+#=
+#added to Matrix.jl
```
fieker, line 108: .. then maybe remove it from here?

src/Sparse/Wiedemann.jl
```diff
@@ -0,0 +1,164 @@
+@doc Markdown.doc"""
```
fieker, line 1: I'd also add a top-level function like `some_nullspace` taking a SMat(A) and calling `wiedemann(A, transpose(A))`

src/Sparse/Matrix.jl
```diff
@@ -952,6 +952,162 @@ function Base.insert!(A::SMat{T}, i::Integer, B::SRow{T}) where T
...
+  end
+  return A
+end
+
+@doc Markdown.doc"""
+    add_scaled_col!(A::SMat{T}, TA::SMat{T}, i::Int, j::Int, c::T) -> SMat{T}
+
+As add_scaled_row!(A::SMat{T}, i::Int, j::Int, c::T) but with columns of $A$.
+"""
+function add_scaled_col!(A::SMat{T}, TA::SMat{T}, i::Int, j::Int, c::T) where T  #A[_j]->c*A[_,i]+A[_j]
```
joschmitt, line 1092: Afterwards `TA` will in general not be the transpose of `A` anymore. Is this intended?

src/Sparse/Matrix.jl
joschmitt: Also, one can always use `searchsortedfirst` here since the list of columns is (hopefully) sorted, so this is a bit faster :)

src/Sparse/Wiedemann.jl
```diff
@@ -0,0 +1,164 @@
...
+    #Wiedemann sequence
+    # solve A^tAx = A^tAr = y => x -r in kernel(A^tA) to avoid finding zero vec
+	y = Hecke.mul!(storing_m,TA, Hecke.mul!(storing_n,A,r))
+	seq[1] = dot(randlin,c,zero!(z)) #randlin*c 		
+	for i = 2:2*n  #Wiedemann sequence 2m?
+        c =  Hecke.mul!(c,TA, Hecke.mul!(storing_n,A,c))
+		seq[i] = dot(randlin,c) 
+	end
+
+	done,f = Hecke_berlekamp_massey(seq)
```
joschmitt, line 27: I guess this should be `Hecke.berlekamp_massey`? And since this is Hecke, `berlekamp_massey` should be enough.

src/Sparse/Wiedemann.jl
```diff
@@ -0,0 +1,164 @@
...
+    #Wiedemann sequence
+    # solve A^tAx = y2 => x -y in kernel(A^tA) to avoid finding zero vec
+	y = multi!(storing_m,TA, multi!(storing_n,A,r))
+	seq[1] = dot(randlin,c,zero!(z)) #randlin*c 		
+	for i = 2:2*n  #Wiedemann sequence
+        c =  multi!(c,TA, multi!(storing_n,A,c))
+		seq[i] = dot(randlin,c) 
+	end
+    ##########################################################################################################################################
+	done,f = Hecke_berlekamp_massey(seq)
```
joschmitt, line 68: As above.

src/Sparse/Wiedemann.jl
joschmitt: Sorry. I just saw the function.

thofma:
Nice, thanks.

It would be great if those `@label/@goto` would be rewritten. It seems they can just be replaced by simple `while` loops or `if` clauses.

src/Sparse/Matrix.jl
BD-00: Since the assignment to i_i happens within the if-part, this shouldn't be called if i not in r.pos, right?
In this case we just jump to the next row

fieker:
On Fri, May 20, 2022 at 04:36:22AM -0700, Berenike Dieterle wrote:
> @BD-00 commented on this pull request.
> 
> 
> 
> > +  add_scaled_row!(A[i], A[j], c)
> +  A.nnz = A.nnz + length(A[j])
> +  return A
> +end
> +
> ***@***.*** Markdown.doc"""
> +    add_scaled_col!(A::SMat{T}, i::Int, j::Int, c::T) -> SMat{T}
> +
> +As add_scaled_row!(A::SMat{T}, i::Int, j::Int, c::T) but with columns of $A$.
> +"""
> +function add_scaled_col!(A::SMat{T}, i::Int, j::Int, c::T) where T 
> +  @assert c != 0
> +  @assert 1 <= i <= ncols(A) && 1 <= j <= ncols(A)  
> +  for r in A.rows
> +    if i in r.pos
> +      i_i = findfirst(isequal(i), r.pos) #changed
> 
> Since the assignment to i_i happens within the if-part, this shouldn't be called if i not in r.pos, right?
Correct, sorry

src/Sparse/Matrix.jl
BD-00: Yes. In the preprocessing function where I use this TA is changed afterwards.
Unfortunately preprocessing for fmpz isn't working at the moment, I'll change the aux function, after finding the error.

src/Sparse/Wiedemann.jl
BD-00: Like this?

src/Sparse/IndexCalculus.jl
```diff
@@ -0,0 +1,358 @@
...
+F = GF(p)
+a = rand(F)
+while a == zero(F)
+    a = rand(F)
+end
+g = rand(1:p-1)
+b = a^g
+
+log_, K = disc_log(a, b)
+a^log_ == b
```
thofma, line 358: Good idea to add a test. But this should not be under `src/` but under `testing`. There is some explanation here: http://www.thofma.com/Hecke.jl/dev/dev/test/.

test/Sparse.jl
```diff
@@ -5,4 +5,5 @@
   include("Sparse/HNF.jl")
   include("Sparse/Rref.jl")
   include("Sparse/Solve.jl")
+  include("Sparse/DiscLog.jl")
```
thofma, line 8:
```suggestion
  include("Sparse/DiscLog.jl")
  include("Sparse/IndexCalculus.jl")
```

src/Sparse/DiscLog.jl
```diff
@@ -0,0 +1,336 @@
...
+	end
+    return set_attribute!(F, :qlimit=>qlimit, :climit=>climit, :ratio=>ratio, :inc=>inc, :RelMat=>A, :FB_array=>FB, :factorbase=>FBs, :fb_length=>l)
+end
+
+function log_dict2(F::T, A, TA )where T<:Union{Nemo.GaloisField, Nemo.GaloisFmpzField}
+    p = get_attribute(F, :p)
+    cnt = 0
+    if !wiedemann(A, TA)[1]
+        @warn "wiedemann failed"
+            return F
```
thofma, line 121:
```suggestion
            return F
```
 indentation

src/Sparse/DiscLog.jl
```diff
@@ -0,0 +1,336 @@
...
+end
+
+
+#compare time of IdxCalc and IdxCalc2
+F = GF(cryptoprime(13))
+a = primitive_elem(F, false)
+b = F(101)
+@time 2*3
+@time g1 = IdxCalc(a,b)
+@time g2 = IdxCalc2(a,b)
```
thofma, line 218: Should not be here.

src/Sparse/DiscLog.jl
```diff
@@ -0,0 +1,336 @@
...
+    #Preprocessing
+    RR = ResidueRing(ZZ, rest)#falsches p ?
+    A = change_base_ring(RR,get_attribute(F,:RelMat))
+    TA = transpose(A)
+    A, TA = sp_prepro(A, TA, get_attribute(F, :fb_length))
+    #Wiedemann + dict with logs of FB
+    cnt = 0
+    if !wiedemann(A, TA)[1]
+        @warn "wiedemann failed"
+            return F
```
thofma, line 254: indentation

src/Sparse/IndexCalculus.jl
```diff
@@ -0,0 +1,293 @@
...
+            end
+            rel += relinc
+        end
+    end
+    #increase Sieve 
+    if nrows(A)/length(FB) < ratio
+		qlimit += inc[1]
+		climit += inc[2]
+		return sieve(F,(qlimit, climit, ratio, inc))
+	end
```
thofma, line 158:
```suggestion
		qlimit += inc[1]
		climit += inc[2]
		return sieve(F,(qlimit, climit, ratio, inc))
	end
```
indentation

src/Sparse/IndexCalculus.jl
```diff
@@ -0,0 +1,293 @@
...
+    log_dict(F::Nemo.Galois(Fmpz)Field, A::SMat, TA::SMat) -> Nemo.Galois(Fmpz)Field
+
+Given a field $F$ with attributes from sieve, logs of factorbase are computed and added to $F$.
+"""
+function log_dict(F::T, A, TA )where T<:Union{Nemo.GaloisField, Nemo.GaloisFmpzField}
+    p = get_attribute(F, :p)
+    cnt = 0
+    if !wiedemann(A, TA)[1]
+        @warn "wiedemann failed"
+            return F
```
thofma, line 178: indentation

src/Sparse/Matrix.jl
```diff
@@ -952,6 +952,172 @@ function Base.insert!(A::SMat{T}, i::Integer, B::SRow{T}) where T
...
+end
+
+@doc Markdown.doc"""
+  delete_zero_rows!(A::SMat{T}, s=1)
+
+Deletes zero rows of $A$ in place.
+"""
+function delete_zero_rows!(A::SMat{T}, s=1) where T #where s denotes the first row where we wanna start
+  for i=A.r:-1:s
+      if A[i].pos == []
```
thofma, line 996:
```suggestion
      if length(A[i].pos) == 0
```
Or `isempty(A[i].pos)`.

src/Sparse/Wiedemann.jl
```diff
@@ -0,0 +1,235 @@
+some_nullspace(A::SMat) = wiedemann(A::SMat, transpose(A)::SMat)
+
+#(p-1)/2 prime 
+@doc Markdown.doc"""
+    wiedemann(A::SMat{gfp_elem}, TA::SMat{gfp_elem}) ->Vector{gfp_elem}
+Computes ker($A$).
+"""
+function wiedemann(A::SMat{gfp_elem}, TA::SMat{gfp_elem}) #N::Int64
```
thofma, line 8: The indentation in this function looks off.

src/Sparse/Wiedemann.jl
```diff
@@ -0,0 +1,235 @@
...
+	#f(TA*A)'c
+    v = Hecke.evaluate(reducedf,TA,A,y).*a
+    return true, (v-r)
+end
+
+@doc Markdown.doc"""
+    wiedemann(A::SMat{gfp_fmpz_elem}, TA::SMat{gfp_fmpz_elem}) ->Vector{gfp_fmpz_elem}
+Computes ker($A$).
+"""
+function wiedemann(A::SMat{gfp_fmpz_elem}, TA::SMat{gfp_fmpz_elem}) #N::fmpz
```
thofma, line 49: Indentation in this function is off.

src/Sparse/Wiedemann.jl
```diff
@@ -0,0 +1,235 @@
...
+	@info "wiedemann 2 used"
+    RR = base_ring(A)
+	N = modulus(RR)
+	T= elem_type(RR)
+	#A = change_base_ring(RR, A)::SMat{T}
+	(n,m) = nrows(A),ncols(A);
```
thofma, line 55:
```suggestion
	(n,m) = nrows(A),ncols(A)
```

src/Sparse/Wiedemann.jl
```diff
@@ -0,0 +1,235 @@
...
+	end
+	constpartoff = coeff(f,0)
+	a = -inv(constpartoff)
+	reducedf = divexact(f-constpartoff,gen(parent(f)))
+    #f(TA*A)'c
+    v = Hecke.evaluate(reducedf,TA,A,y).*a
+    return true, (v-r)
+end
+
+function wiedemann(A::SMat{T}, TA::SMat{T}) where T<:Union{nmod, fmpz_mod}
```
thofma, line 87: Indentation is wrong.

src/Sparse/Wiedemann.jl
```diff
@@ -0,0 +1,235 @@
...
+	@info "wiedemann 3 used"
+    RR = base_ring(A)
+	N = modulus(RR)
+	#T= elem_type(RR)
+	#A = change_base_ring(RR, A)::SMat{T}
+	(n,m) = nrows(A),ncols(A);
```
thofma, line 93:
```suggestion
	(n,m) = nrows(A),ncols(A)
```

src/Sparse/Wiedemann.jl
thofma, line 49: Is this not the same function as above?

src/Sparse/Wiedemann.jl
```diff
@@ -0,0 +1,235 @@
...
+	C = collect(coefficients(f))
+	n = length(C)
+	s =  Hecke.mul!(s,TA, Hecke.mul!(storing_n,A,c)).*C[end]+c.*C[end-1]
+	for i = n-2:-1:1
+		s =  Hecke.mul!(s,TA, Hecke.mul!(storing_n,A,s)) + c.*C[i]
+	end
+	return s
+end
+
+function Hecke.evaluate(f,TA,A::SMat{T},c) where T <:Union{gfp_fmpz_elem, fmpz_mod}
```
thofma, line 184: Indentation. What is the difference to the function above?

src/Sparse/Wiedemann.jl
```diff
@@ -0,0 +1,235 @@
...
+	end
+	constpartoff = coeff(f,0)
+	a = -inv(constpartoff)
+	reducedf = divexact(f-constpartoff,gen(parent(f)))
+    #f(TA*A)'c
+    v = Hecke.evaluate(reducedf,TA,A,y).*a
+    return true, (v-r)
+end
+=# 
+function Hecke.evaluate(f,TA,A::SMat{T},c) where T <:Union{gfp_elem, nmod}
```
thofma, line 169: Indentation.

src/Sparse/Wiedemann.jl
```diff
@@ -0,0 +1,235 @@
...
+
+#=
+F = GF(fmpz(34589))
+F2 = ResidueRing(ZZ, fmpz(17294))
+a = F(2)
+set_attribute!(F, :a, a)
+sieve(F)
+A = change_base_ring(F2, get_attribute(F,:RelMat))
+TA = transpose(A)
+=#
```
thofma, line 235:
```suggestion
```

thofma (commented):
I left another round of comments, mostly about wrong indentation.

Once those are addressed, this can be merged.

src/Sparse/Wiedemann.jl
BD-00: It differs in the computation of the wiedemann sequence. For fmpz I use multi! (defined in the same file below) instead of mul!.

src/Sparse/Wiedemann.jl
BD-00: mul! and multi! again

thofma:
The tests fail with `UndefVarError: disc_log not defined`. Either export `disc_log` or use `Hecke.disc_log` in the test.

src/Sparse/Wiedemann.jl
thofma: OK, thanks for the explanation. For the future, to avoid code duplication, one would do
```
my_mul!(c::Vector{T}, A::SMat{T}, b::Vector{T}) where T <:Union{gfp_fmpz_elem, fmpz_mod} = multi!(c, A, b)

my_mul!(c::Vector{T}, A::SMat{T}, b::Vector{T}) where T <:Union{gfp_elem, nmod} = mul!(c, A, b)
```
and then have only one version of `wiedemann` where `my_mul!` is used.

But this can be done later.

BD-00:
> The tests fail with `UndefVarError: disc_log not defined`. Either export `disc_log` or use `Hecke.disc_log` in the test.

export in src/Sparse.jl or the file where the function is defined? same with IdxCalc function?

thofma:
I think it would be easier to use `Hecke.disc_log` and `Hecke.IdxCalc` in the tests.

thofma:
It seems that the file `src/Sparse/DiscLog.jl` is never included in Hecke, so the tests are failing.

The include should also be cleaned up. Things are included multiple times, which is not a thing in julia.

The rule we follow is that any file in `src/Sparse/*.jl` should be included in `src/Sparse.jl`.

test/Sparse/DiscLog.jl
```diff
@@ -0,0 +1,41 @@
...
+  for i = 5:20
+    p = rand_dec_prime(i)
+    F = GF(p)
+    a = Hecke.primitive_element(F)
+    b = rand(F)
+    g = Hecke.disc_log(a,b)
+    @test a^g == b 
+  end
+end
+=#
```
thofma, line 41:
```suggestion
=#
end
```

thofma:
The real error is now here: https://github.com/thofma/Hecke.jl/runs/7443994926?check_suite_focus=true#step:6:629

src/Sparse/Matrix.jl
thofma: Has this been fixed? It is also not clear what `TA` is (not mentioned in docstring).

src/Sparse/Matrix.jl
thofma: still does not match the signature

src/Sparse/Preprocessing.jl
thofma: should be removed

src/Sparse/Wiedemann.jl
thofma: @fieker like this?

src/Sparse/Preprocessing.jl
BD-00: @thofma This has been removed a few commits ago. My version  of this file ends at line 97.

src/Sparse/Matrix.jl
BD-00: @thofma 

@doc Markdown.doc"""
    function scale_col!(A::SMat{T}, TA::SMat{T}, j, c) -> SMat{T}

Returns $A$ after scaling j-th column in place using the transposed matrix TA of A.
"""
function scale_col!(A::SMat{T}, TA::SMat{T}, j, c) where T #A[_j]->c*A[_,j]


Is this better?

src/Sparse/Preprocessing.jl
thofma: Sorry, you are right. I got confused by the github interface.

src/Sparse/Matrix.jl
BD-00: I revised some of the docstrings. TA is now mentioned for every function it is used

thofma:
Tests are failing. Note that you can run the tests locally by doing `Pkg.add("RandomExtensions")` (needs to be done only once) and then `Hecke.test_module("Sparse")`.

BD-00:
> Tests are failing. Note that you can run the tests locally by doing `Pkg.add("RandomExtensions")` (needs to be done only once) and then `Hecke.test_module("Sparse")`.

They all pass locally. I forgot to add "Hecke."  in front of some functions. This should now work.

test/Sparse/IndexCalculus.jl
```diff
@@ -0,0 +1,58 @@
...
+  #lift(a3) in FB (not on first position)
+  g6,_ = Hecke.IdxCalc(aq, bq)
+  @test aq^g6 == bq
+
+
+  if long_test
+    for i = 10:20
+      pr = cryptoprime(i)
+      F = GF(pr)
+      a = primitive_elem(F, false)
```
thofma, line 49:
```suggestion
      a = Hecke.primitive_element(F)
```

PR #707, opened 2022-05-21
Fix range for cochain complex

AlexD97:
Actually, I'm not quite sure if `range` is now finally correct. I would assume that the range starts with given `start`-value but this is not the case since the indices of `maps` start with 1 (not with 0).


fieker:
range is the object index range, not the map index range if I recall correctly. Whether that is what one wants, I don't know. Maybe we also need a map range?

fieker (approved):

AlexD97:
IMO a map range would be very convenient.

PR #708, opened 2022-05-21
Adding lots of tests and removing unused code to increase coverage + bug

JHanselman:
fixes

### DIFF
--- a/src/EllCrv/Isogeny.jl
+++ b/src/EllCrv/Isogeny.jl
@@ -316,19 +316,22 @@ function multiplication_by_m_map(E::EllCrv, m::S) where S<:Union{Integer, fmpz}
     return isomorphism_to_isogeny(identity_map(E))
   end
   
-  a1, a2, a3 = a_invars(E)
-
+  p = characteristic(base_field(E))
+  
+  if p == 3
+    error("Not yet implemented")
+  end
+  
+  if !is_simplified_model(E)
+    F, pre_iso, post_iso = simplified_model(E)
+    return pre_iso * multiplication_by_m_map(F, m) * post_iso
+  end
+      
   Kx, x = PolynomialRing(base_field(E), "x")
   Kxy, y = PolynomialRing(Kx, "y")
   
   mult_mx = multiplication_by_m_numerator(E, m, x)//multiplication_by_m_denominator(E, m, x)
-  mult_mx_deriv = derivative(mult_mx)
-
-  x = Kxy(x)
-  mult_mxy = multiplication_by_m_numerator(E, m, x)//multiplication_by_m_denominator(E, m, x)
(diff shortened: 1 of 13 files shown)


src/EllCrv/Isomorphisms.jl
```diff
@@ -538,7 +568,9 @@ function automorphism_group_generators(E::EllCrv{T}) where {T}
       end
     end
     size = length(auts)
+    
     if size == 6 #Group isomorphic to Quaternions
+      println("hi")
```
thofma, line 573:
```suggestion
```

src/EllCrv/Isomorphisms.jl
```diff
@@ -538,7 +568,9 @@ function automorphism_group_generators(E::EllCrv{T}) where {T}
       end
     end
     size = length(auts)
+    
     if size == 6 #Group isomorphic to Quaternions
```
thofma, line 572: Is this comment correct?

src/EllCrv/Isogeny.jl
```diff
@@ -343,6 +346,19 @@ function multiplication_by_m_map(E::EllCrv, m::S) where S<:Union{Integer, fmpz}
   return mul_m
 end
 
+
+function defrobenify(f::RingElem, p)
```
thofma, line 350: Does not seem to be called anywhere?

src/EllCrv/MinimalModels.jl
```diff
@@ -894,29 +896,35 @@ end
...
+      EllCrvIso{fmpq}, EllCrvIso{fmpq}
 
 Returns a model of $E$, which is minimal at $p$. It is assumed that $p$
 is prime.
 """
 function minimal_model(E::EllCrv{fmpq}, p::Int)
   Ep = tates_algorithm_local(E, p)
-  return Ep
+  phi = isomorphism(E, Ep)
+  return Ep, phi, inv(phi)
```
thofma, line 914:
```suggestion
  return Ep, phi
```
Should we do the same as for the global minimal model?

JHanselman: I think it's more convenient for the user to return both the map and its inverse everywhere. That's also the way Magma does it.The global minimal model also returns both maps.

thofma: Sorry, I did not know that the global minimal model also returned both. Then I propose to also change the global minimal model (but I will do this in a follow up PR). I think returning both maps in Hecke is from a time when there were no proper map objects and calling `inv` was not an option.

I find the Magma behavior inconsistent with the rest of Magma (but also inconsistent in the elliptic curve world itself, since `MinimalModel(E, p) -> CrvEll, Map` in the number field case).

So I'd rather have `minimal_model` in its various forms behave like all other cases where we return an invertible map, e.g. `is_isomorphic`.

@fieker Do you have an opinion?

PR #709, opened 2022-05-23
Updated NfRelNS with some accessors and tests

antonydellavecchia:
Merge allows Non Simple extensions of relative number field to be created with vector of symbols,
also adds some basic tests for these types of extensions.

### DIFF
--- a/test/NfRel/NonSimple.jl
+++ b/test/NfRel/NonSimple.jl
@@ -2,7 +2,7 @@
   Qx, x = FlintQQ["x"]
   K, _ = number_field(x^2 - 2)
   Ky, y = K["y"]
-  L, (a, b) = @inferred number_field([x^2 - 3, x^3 - 5])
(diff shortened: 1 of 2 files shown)


test/NfRel/NonSimple.jl
```diff
@@ -0,0 +1,20 @@
+@testset "NfRelNS" begin
+  Qx, x = FlintQQ["x"]
+  K, _ = number_field(x^2 - 2)
+  Ky, y = K["y"]
+  L, (a, b) = @inferred number_field([x^2 - 3, x^3 - 5])
```
thofma, line 5: Do you want to use `y` here? At the moment this is not calling the new methods.

antonydellavecchia: yup, thanks

thofma:
Thanks, will merge once green.

PR #712, opened 2022-05-24
Add some SparseTrafoElem code comments

fingolfin:
Also handle type 7 in apply_left! and apply_right!


thofma:
Did this come up for the invariant theory? Just curious

PR #713, opened 2022-05-24
Remove trailing whitespace

fingolfin:
Commit created via this command:

    git ls-files | xargs perl -pi -e 's/[ \t]+$//'

*If* this should be merged, I would suggest to afterwards add the resulting commit to a `.git-blame-ignore-revs` file (see https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/)


examples/Suri.jl
```diff
@@ -526,7 +526,7 @@ function extend(A::RRSMat, P::Int)
 
   # have (f mod p)(a_i) - need f mod p  #TODO: list of p!
   fp = Dict{Int, Array{nmod_mat, 1}}()
-  k = 
+  k =
```
thofma, line 529: :)

fingolfin: 😂 I guess this assigns the return value of the following loop, and hence does not raise an error? Of course it's still not very useful...

PR #720, opened 2022-05-28
Fix map indexing for cochain complexes

AlexD97:
I'm not sure, but I think also `getindex` for cochain complexes is wrong.


fieker (changes requested):
I have the feelig that the "same" error is present in some other functions as well, e.g. `grp_ab_fill`

PR #725, opened 2022-06-03
Add _roots over QQ

thofma:


thofma:
@JHanselman: You can then use `_, rts, _ = _roots(f, P, prec = bla)` and `rts` will contain the real roots (as elements of `arb`). This works for number fields and Q.

PR #732, opened 2022-06-18
Height bounds, rational points, kernel_polynomial

JHanselman:


src/EllCrv/Heights.jl
```diff
@@ -34,8 +34,8 @@
 #
 ################################################################################
 
-export local_height, canonical_height, naive_height, height_pairing, regulator,
-       neron_tate_height
+export local_height, real_height, canonical_height, naive_height, height_pairing,
```
thofma, line 37:
```suggestion
export local_height, canonical_height, naive_height, height_pairing, 
```

PR #734, opened 2022-06-20
fixed `snf_with_transform`

ThomasBreuer:
In the case of negative diagonal entries,
if one wants to change the signs by adjusting the left transformation matrix
then this must be done from the left, that is, by multiplying rows.


ThomasBreuer:
It is interesting that the code coverage is decreasing:
Exactly one code line was changed (and became shorter), and a test has been added.

thofma:
We had this conversation before: https://github.com/thofma/Hecke.jl/pull/592#issuecomment-1036168657

PR #750, opened 2022-07-13
call roots_squarefree in conjugates

fieker:


thofma:
Yes, thanks. I forgot to adjust it after I adjusted the `_roots` function.

PR #756, opened 2022-07-15
Selmer

fieker:
WIP

but workable


thofma:
Do you want to add an example to the tests?

fieker:
Oh yes, and do the qq-case as well, and a non-fac-elem

src/NumField/Selmer.jl
```diff
@@ -142,4 +177,122 @@ function pselmer_group_fac_elem(p::Int, S::Vector{<:NfOrdIdl})
...
+      fl, r = is_power_with_root(k, p)
+      fl || return false, a
+      push!(K, r)
+      push!(V, v)
+    end
+  end
+  return true, FacElem(QQ, K, V, parent = parent(a))
+end
+
+#TODO: do we want that?
```
thofma, line 279: I would say no.

fieker:
I'll remove it. It came in since fac elem parents in number fields are
cached, but QQ is not...

PR #757, opened 2022-07-16
Change `mass(::ZLat)`

StevellM:
It is faster to compute the mass of the genus of a `ZLat` rather than the mass of the corresponding `QuadLat`.


thofma:
That was easy :)

PR #758, opened 2022-07-19
LLL-reduction for `ZLat`

StevellM:
This adds a command to perform LLL reduction on a basis matrix for `ZLat`, I am not strict on the naming. There are some fixes on namings, documentations and conventions in the new file `indefiniteLLL.jl`, to try to be as close as possible to the definite version `lll_gram_*`.  Once again nothing has to be fixed, we can discuss the new namings and also the relevance of the cases chosen in the new function `lll_reduction`: I wanted to use as much as possible the new functions (which doesn't always provide the exact same outputs, but still right in any cases)


thofma:
I guess this is fine. I am not too big of a fan of providing lll directly for lattices, since it is not really a property of the lattice. But I see why it could be useful. What application do you have in mind?

StevellM:
The idea comes from enumeration of lattices in definite genera. For instance: in order to avoid "explosion" of the entries in the gram matrices of the neighbours, every time we find a new neighbour, you reduce it. In this way, when we compute chains of neighbours, the coefficients in the gram matrices stay small (for computing the partial masses and to do isometry tests).

thofma:
OK (the isometry test should be doing this already).

PR #761, opened 2022-07-27
Fix _automorphisms_center and is_cm_field

thofma:


thofma:
@fieker can you have a superficial look?

fieker:
Looks OK - I haven;t checked the math though, but it looks good to me

thofma:
Not that much math going on. Unless the tentative automorphisms obtained from `_automorphisms_center` contain complex conjugation, we redo the computation with the full automorphism group.

It seems that `_automorphisms_center` is not used anywhere else.

PR #772, opened 2022-08-07
Iterator for `TorQuadMod`

StevellM:
Adding iterators for `TorQuadmod`, based on the iterator of the underlying abelian group.

### DIFF
--- a/src/QuadForm/Torsion.jl
+++ b/src/QuadForm/Torsion.jl
@@ -453,26 +453,19 @@ end
 
 Base.length(T::TorQuadMod) = Int(order(T))
 
-Base.IteratorSize(::TorQuadMod) = Base.HasLength()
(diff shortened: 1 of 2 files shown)


thofma:
Looks good. I think it would make sense to also define `elttype` (see https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-iteration).

It does not matter too much, but is there a good reason to not call `iterate` on the abelian group `A` directly? Like
```
Base.iterate(T::TorQuadMod) = begin; a, st = iterate(abelian_group(T)); return T(a), st; end
```
etc.

StevellM:
Yes indeed, I will add the `elttype`.
For the second part, to be honest I have thought about doing this, but I can modify it and make it in this way.

PR #774, opened 2022-08-09
ChainComplexes: Add map/object_range/extract; fix start and show

AlexD97:


thofma:
Is it expected that the Oscar tests will fail with this? Just asking so we know how to proceed.

jankoboehm:
It fixes the print function for complexes (so printout will change), introduces functions for map and object range, and to extract parts of the complex using map and object range, also a shift operation for the degree. These functions are not used yet in OSCAR, but will be used in a PR to come. ChainComplex is exported to help to improve the docu in Oscar avoiding to write Hecke.

thofma:
This PR here makes the OSCAR tests fail: https://github.com/thofma/Hecke.jl/runs/7788161199?check_suite_focus=true#step:7:799

Is this intentional?

jankoboehm:
No it should not.

AlexD97:
Yes, it should break the tests in UngradedModules.jl. They will then be fixed with another PR.

jankoboehm:
We will make the PR for OSCAR. Once everyhting is right there this one can be merged.

AlexD97:
https://github.com/oscar-system/Oscar.jl/pull/1508 should fix the failed tests.

PR #779, opened 2022-08-12
lll as alias for lll_reduction

simonbrandhorst:


thofma:
Why not change it directly to `lll`?

simonbrandhorst:
sounds good

PR #782, opened 2022-08-16
`ZLat` to Magma/Sage

StevellM:
Add conversions to parseble Magma and Sage codes for `ZLat`. 

Minor fix on the Hecke one: one entry argument was missing (while useless in this case... is it necessary though for `HermLat` and `QuadLat` ?)


thofma:
Nice, thanks! Could you add a test? Just calling the functions should be sufficient.

StevellM:
Sure, I'll do that now.

PR #792, opened 2022-08-19
Root lattice recognition

simonbrandhorst:


src/QuadForm/Quad/ZLattices.jl
```diff
@@ -1099,13 +1100,238 @@ function lll(L::ZLat; same_ambient::Bool = true)
...
+end
+
+
+################################################################################
+#
+#  Primitive extensions and glue maps
+#
+################################################################################
+
+
```
StevellM, line 1334: The 
```
@doc Markdown.doc"""
```
dropped here

PR #798, opened 2022-08-24
Cleanup for close vectors functionality

thofma:
Here are the changes implemented after discussions with @simonbrandhorst:

- Only user facing function is `close_vectors(::ZLat, ::Vector, ::Any)`. Return values are the vectors and the distance.
- Add a `filter` keyword.
- Add tests testing the result.
- Remove quadratic triple code.
- Keep `closest_vectors` functions for now (backwards compatibility).


src/QuadForm/CloseVectors.jl
```diff
@@ -0,0 +1,178 @@
...
+end
+
+function _close_vectors(L::ZLat, v::Vector{fmpq}, upperbound::fmpq;
+                                equal::Bool=false, sorting::Bool=false,
+                                check=true, filter = nothing) where T <: RingElem
+  epsilon = QQ(1//10)   # some number > 0, not sure how it influences performance
+  d = length(v)
+  G1 = gram_matrix(rational_span(L))
+  if check
+    @req is_definite(L) == true && G1[1, 1] > 0 "Zlattice must be positive definite"
```
simonbrandhorst, line 47:
```suggestion
    @req is_positive_definite(L)  "Zlattice must be positive definite"
```

src/QuadForm/CloseVectors.jl
```diff
@@ -0,0 +1,178 @@
...
+  V = rational_span(L)
+  for a in sv
+    _a, _l = a
+    al = _a[end]
+    if iszero(al)
+      continue
+    end
+    @assert al == 1 || al == -1
+
+    x = fmpz[_a[i] for i in 1:d]
```
simonbrandhorst, line 86: What is your reason for this conversion? We are not doing it for `short_vectors` so why here?

simonbrandhorst (approved):

src/QuadForm/CloseVectors.jl
thofma: It was there before and I guess you had a reason. I will adjust it.

src/QuadForm/CloseVectors.jl
thofma: Thanks, I forgot that we had that.

thofma:
I fix the filter in a separate PR.

PR #802, opened 2022-08-25
Fix root_sublattice

simonbrandhorst:


thofma:
Do you have a failing example we could add as a test?

Issue #647, opened 2022-04-14
issquarefree is not doing what it says


ulthiel:
The function issquarefree for f::PolyElem (not assuming that f has coefficients in a field, which may be dangerous) returns isconstant(gcd(f, derivative(f))). But this is not the condition for squarefree, rather for being separable, see [1]. Over non-perfect fields you have squarefree polynomials which are not separable, e.g. x^p - t in 𝔽_p(t)[x] (an irreducible, thus squarefree, polynomial, but it is not separable).

1. Bourbaki, N. (2003). *Algebra II. Chapters 4--7*. Springer-Verlag, Berlin. (Translated from the 1981 French edition by P. M. Cohn and J. Howie, Reprint of the 1990 English edition). Chapter 5, §7, No. 2.
2. Becker, T. & Weispfenning, V. (1993). *Grobner bases*. Springer-Verlag, New York. §7.3.

thofma:
Fixed in #806.

PR #806, opened 2022-08-26
Fix is_squarefree

thofma:


thofma:
This was quite a rabbit hole, but should close #647.

PR #815, opened 2022-09-20
Factor bug

fieker:


fieker:
as far as I can see: the errors are unrelated. A "fluke" in the quad-form database and a serious regression in latest julia: an infinite recursion in the type-system again...

PR #816, opened 2022-09-21
additional input check for close and short vectors functions

simonbrandhorst:


simonbrandhorst:
@thofma Do you know why  the tests` AlgAssAbsOrd/Conjugacy/Conjugacy.jl ` sometimes fail? Is that expected?

thofma:
I have not seen the error. Do you remember what it said?

thofma:
It seems there are some tests which used negative definite matrices.

PR #835, opened 2022-10-08
`VecSpaceRes` as `HeckeMap`

StevellM:
Needed a way to access the domain and codomain of the map from `restrict_scalars` for `AbsSpace`. Seemed that the type `VecSpaceRes` was not seen as a subtype of `HeckeMap`, so it had no header. I made this change, and modify the functions accordingly to match the new caching.

### DIFF
--- a/src/QuadForm/Misc.jl
+++ b/src/QuadForm/Misc.jl
@@ -4,32 +4,27 @@
 #
 ################################################################################
 
-mutable struct VecSpaceRes{S, T}
-  field::S
-  domain_dim::Int
-  codomain_dim::Int
-  absolute_basis::Vector{T}
-  absolute_degree::Int
-end
+mutable struct VecSpaceRes{S, T} <: Map{S, T, HeckeMap, VecSpaceRes}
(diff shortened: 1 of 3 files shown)


thofma:
Makes sense. The error in the tests is https://github.com/thofma/Hecke.jl/issues/836.

thofma:
Hm, I should have looked at the changes haha. I am not happy with them. I need this for the general K^n to Q^m restriction, which is independent of the quadratic spaces.

thofma:
I should have mentioned that the uses of `VecSpaceRes` that I have are in code not contained in Hecke itself. @StevellM What do you think of the following? We restore the old behaviour of `VecSpaceRes` and a thin wrapper to get a `SpaceRes` or something like that?

StevellM:
@thofma Sounds good to me, as far as we can store the domain and codomain somehow, it is fine for me. So if possible, we can undo this pr and once I will have a bit more time, I will add sort of a wrapper as a HeckeMap. Let's see for the naming.

PR #840, opened 2022-10-11
Fix `Hecke.to_*` methods

StevellM:
There were some issues with changes between `//` and `/`. Also some missing change of symbols for the coefficients in some places, and also we missed the cases where the base field for `QuadLat` might not be absolute. 

I had to change also the symbol for `CyclotomicRealSubfield` in the `cyclotomic_field_as_cm_extension` because it prints only one `/` for the quotients. So when getting the symbol, there was some issues I preferred to get rid of. For instance, trying 
```
E, b = cyclotomic_field_as_cm_extension(7);
Q = quadratic_space(E, 2);
L = lattice(Q);
Hecke.to_hecke(L)
```
this would have printed some things of the form `b + 1/b` which we don't want. It is not strict of course, any other alternative would be good.


StevellM:
Ah yes indeed. I will make the changes.

PR #843, opened 2022-10-11
Allow "column vectors" in the construction of group elements

joschmitt:
A small change to allow this:
```
julia> G = abelian_group([3, 0]);

julia> N = matrix(FlintZZ, 2, 1, [ 1, 1 ]);

julia> G(N)
Element of
GrpAb: Z/3 x Z
with components [1 1]

```
Seems to me to be the easiest way to make the construction in https://github.com/oscar-system/Oscar.jl/blob/5c7916ed7831ef993ba108418151f01ff0f5c9f2/src/Rings/mpoly-graded.jl#L182 work (without having to write a separate function for `fmpz_mat` there).


thofma:
Are you happy with this @fieker?

PR #845, opened 2022-10-12
Multiplicative order of rational matrices

StevellM:
This is something I already discussed one day on Slack, and this was one of the suggested solution. It was implemented on one of my private repositories so I brought it here. I am not strict, I guess it can still be improved but this does the job and in quite good time. 
I took the freedom to add another function for cyclotomic polynomials: the methood `cyclotomic` is not exported on its own and the naming is not really explicit.


thofma:
Looks good, but can you change all the `Int64` to just `Int`?

fieker:
I am confused: while I agree that cyclotomic is not the best name it is exported and has been exported for ages...

StevellM:
In fact yes, I am wrong on this part. Sorry, I don't know why I thought it was not, I have always used it with `Hecke.` ... 
Still, the naming and the necessity to input a `fmpz_poly` as second argument is quite inconvenient, I guess.

fieker:
On Wed, Oct 12, 2022 at 02:44:03AM -0700, Stevell Muller wrote:
> In fact yes, I am wrong on this part. Sorry, I don't know why I thought it was not, I have always used it with `Hecke.` ... 
> Still, the naming and the necessity to input a `fmpz_poly` as second argument is quite inconvenient, I guess.
> 
I fully agree - in fact, in line with swinnerton_dyer, I'd actually
prefer to also have
 - a version that just uses Hecke.Globals.Zx
 - a version acceptong a ring elem and returning the poly evaluated at
   the 2nd argument. In case the arguement is x it would return the
   poly...

StevellM:
What about keeping `cyclotomic_polynomial` and having something like `cyclotomic_polynomial_evaluated_at` ? Or similar kind of name which does what you mean: it evaluates the $n$-th cyclotomic polynomial at a ring element, and if this element if the generator `x` of a univariate polynomial ring, we just return the corresponding cyclotomic polynomial in `parent(x)` ? 

I could actually change the order of the argument, having a version for `cyclotomic_polynomial` accepting one argument `n` and returning the $n$-th cyclotomic polynomial as `fmpz_poly`, and another version where we could specify the ring (as first argument)

At the end they will be just extended versions of `cyclotomic` but a bit more explicit and flexible on the use.

fieker:
On Wed, Oct 12, 2022 at 03:02:27AM -0700, Stevell Muller wrote:
> What about keeping `cyclotomic_polynomial` and having something like `cyclotomic_polynomial_evaluated_at` ? Or similar kind of name which does what you mean: it evaluates the $n$-th cyclotomic polynomial at a ring element, and if this element if the generator `x` of a univariate polynomial ring, we just return the corresponding cyclotomic polynomial in `parent(x)` ? 
> 
> I could actually change the order of the argument, having a version for `cyclotomic_polynomial` accepting one argument `n` and returning the $n$-th cyclotomic polynomial as `fmpz_poly`, and another version where we could specify the ring (as first argument)
> 
> At the end they will be just extended versions of `cyclotomic` but a bit more explicit and flexible on the use.
> 
if the argument is an FmpzPolyRiong, this would make "sense", however if
the argument is "x" then do we want to check that this is a generator,
use only parent() or evaluate? If the answer is evaluate, than "x" is no
diffrent from "3" or "x+1"
Other uses are also to get fmpq_poly or Poly{nf_elem}...

I was certain we have cyclotomic(n) - but aparently we don't
There are similar issues with
 - swinnerton_dyer
 - cos_minpoly
 - a handful others..
(including charpoly, minpoly)

StevellM:
Well I was thinking of the same way it is already done: if `is_gen(x)` then evaluating at `x` or taking the corresponding cyclotomic polynomial in `parent(x)`, given that it is a a polynomial ring, will be the same, so we avoid the evaluation process. I don't see where it would not make sense..
In the actual form, we are able to create cyclotomic polynomial in any `PolyRing{T}` where `T <: RingElement`, so we can cover `fmpq_poly`, `PolyElem{nf_elem}`, etc...

StevellM:
@fieker Would you like we open an issue for a later PR (or set of PRs...) regarding the question of the wrappers for this "polynomial functions" ?
In the meantime, if you agree with their actual format, we could keep the functions regarding cyclotomic polynomials I have added for now and then we will discuss for a more general pattern to be applied to all the `cyclotomic`, `swinnerton_dyer`, etc... functions. Does it sound good to you ?

src/Misc/Poly.jl
```diff
@@ -1382,3 +1382,68 @@ function is_squarefree(f::PolyElem)
...
 
+################################################################################
+#
+#  Cyclotomic polynomial
+#
+################################################################################
+
+@doc Markdown.doc"""
+    cyclotomic_polynomial(n::Int, R::PolyRing{T} = ZZ["x"]) where T
+                                                                  -> PolyElem{T}
```
fieker, line 1393: I think ZZ["x"] should be Hecke.Globals.Zx

src/Misc/Poly.jl
```diff
@@ -1382,3 +1382,68 @@ function is_squarefree(f::PolyElem)
...
+
+julia> cyclotomic_polynomial(9)
+x^6 + x^3 + 1
+
+julia> typeof(ans)
+fmpz_poly
+```
+"""
+function cyclotomic_polynomial(n::Int, R::PolyRing{T} = Hecke.Globals.Zx) where T
+  @req n > 0 "n must be positive"
```
fieker, line 1418: very good!

src/Misc/Poly.jl
```diff
@@ -1382,3 +1382,68 @@ function is_squarefree(f::PolyElem)
...
+  x = gen(Hecke.Globals.Zx)
+  p = Hecke.cyclotomic(n, x)
+  return R(collect(coefficients(p)))::PolyElem{T}
+end
```
fieker, line 1422: the "usual" would be ```map_coefficients(base_ring(R), p, parent = R)```

fieker:
In general I can live with the interface (and for the sake of the argument the current implementation)

simonbrandhorst:
Please resolve the merge conflict.

simonbrandhorst (approved):

thofma:
Too many `Hecke.*`.

PR #854, opened 2022-10-19
Maximal overlattice

simonbrandhorst:


src/Misc/Matrix.jl
```diff
@@ -590,6 +590,10 @@ function lift(a::fmpz_mod_mat)
   return z
 end
 
+function lift(x::gfp_fmpz_mat)
+  return map_entries(lift , x)
+end
+
```
simonbrandhorst, line 596: @thofma could you please take a look if this is the way to go? Maybe there is a flint function for it?

thofma: I guess this is fine for now.

src/QuadForm/Quad/ZLattices.jl
```diff
@@ -918,12 +918,214 @@ end
...
+    v[1,n] = 1
+    return true, v
+  end
+  if D[1]+D[n-1] + D[n] == 0
+    v[1,1] = 1
+    v[1,n-1] = 1
+    v[1,n] = 1
+    return true, v
+  end
+  @assert false
```
thofma, line 1126:
```suggestion
  error("Something wrong!")
```
Julia might decide to remove asserts at some later point.

PR #855, opened 2022-10-20
Correctly set `nnz` in `can_solve_with_solution`

joschmitt:
I forgot to set the `nnz` of the solution, when I wrote `can_solve_with_solution` for sparse matrices. Sorry!


thofma:
Thanks!

PR #858, opened 2022-10-24
Primary/elementary discriminant group

StevellM:
I am not strict on the naming, though it seems to be the ones usually used.

I fixed the `rescale` function for `TorQuadMod` which was buggy (for the printing) in the case where we were rescaling a trivial torsion quadratic module.

### DIFF
--- a/src/QuadForm/Quad/ZLattices.jl
+++ b/src/QuadForm/Quad/ZLattices.jl
@@ -1813,3 +1813,81 @@ function overlattice(glue_map::TorQuadModMor)
   return lattice(ambient_space(S), B[end-rank(S)-rank(R)+1:end,:])
 end
 
+################################################################################
+#
+#  Primary/elementary lattices
+#
+################################################################################
+
+@doc Markdown.doc"""
+    is_primary(L::ZLat) -> Bool, fmpz
+
+Given a $\mathbb Z$-lattice `L`, return whether `L` is primary, that is whether its
+discriminant group is a `p`-group for some prime number `p. In case it is, `p` is
+also returned as second output.
+
+Note that for unimodular lattices, this function returns `(true, 1)`. If the
+lattice is not primary, the second return value is `-1` by default.
+"""
+function is_primary(L::ZLat)
+  @req is_integral(L) "L must be integral"
+  d = ZZ(det(L))
+  if d == 1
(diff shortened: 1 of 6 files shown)


thofma:
I think we are transitioning to `is_*` just returning a Bool (people seem to favour that). So `is_primary` would be a Bool and the other would be `is_primary_with_prime`. See also https://github.com/oscar-system/Oscar.jl/pull/1574.

(I know that many functions in Hecke do not follow this yet).

src/QuadForm/Quad/ZLattices.jl
```diff
@@ -1813,3 +1813,82 @@ function overlattice(glue_map::TorQuadModMor)
...
+end
+
+@doc Markdown.doc"""
+    is_elementary(L::ZLat, p::Union{Integer, fmpz}) -> Bool
+
+Given a $\mathbb Z$-lattice `L` and a prime number `p`, return whether `L`
+is `p`-elementary, that is whether its discriminant group is an elementary
+`p`-group.
+"""
+function is_elementary(L::ZLat, p::Union{Integer, fmpz})
```
thofma, line 1890: Is this standard definition? Since L itself is an abelian group, it could be elementary itself (which it never will be, but still). It is like defining the order of L to be the order of the discriminant group.

But I am not too familiar with the literature. @simonbrandhorst any thoughts?

StevellM: I am following the definition of Conway and Sloane for p-elementary lattices yes. For primary I am not sure whether it is standard, but it is coherent with the elementary definition... I remember seeing it used in some paper also.

thofma: Ok. It just looks odd to forward some properties of L to the discrimant group (and some not). But I am happy with it if this the convention.

StevellM: Good. Simon wanted to have a look at it in any case so he will be able to confirm (or not!)

simonbrandhorst: Yes, `p`-elementary for `p` a prime is standard.

simonbrandhorst (changes requested):
Please add the new functions to the documentation.

PR #864, opened 2022-11-02
Fix #859 take 2

### LINKED ISSUE #859: `pseudo_hnf` for non-full rank pseudo-matrix
```
julia> Qx, x = PolynomialRing(FlintQQ, "x");

julia> f = x - 1;

julia> K, a = NumberField(f, "a", cached = false);

julia> Kt, t = PolynomialRing(K, "t");

julia> g = t^2 + 1;

julia> E, b = NumberField(g, "b", cached = false);

julia> D = matrix(E, 3, 3, [1, 0, 0, 0, 1, 0, 0, 0, 1]);

julia> gens = Vector{Hecke.NfRelElem{nf_elem}}[map(E, [b + 2, 1, 0])];

julia> L = hermitian_lattice(E, gens, gram = D)
Hermitian lattice of rank 1 and degree 3
over
Relative maximal order with pseudo-basis (1) * 1//1 * <1, 1>, (b) * 1//1 * <1, 1>

julia> pseudo_hnf(pseudo_matrix(L));

julia> lattice(ambient_space(L), ans);

julia> ans == L
false

```
In theory, this should return equal lattices. The fact here is that `L` is not of full rank and its coefficient ideals are non-trivial, so it seems that the `pseudo_hnf` function is doing something unexpected. Here is a block code to copy-paste for reproduction:

```
Qx, x = PolynomialRing(FlintQQ, "x");
f = x - 1;
K, a = NumberField(f, "a", cached = false);
Kt, t = PolynomialRing(K, "t");
g = t^2 + 1;
E, b = NumberField(g, "b", cached = false);
D = matrix(E, 3, 3, [1, 0, 0, 0, 1, 0, 0, 0, 1]);
gens = Vector{Hecke.NfRelElem{nf_elem}}[map(E, [b + 2, 1, 0])];
L = hermitian_lattice(E, gens, gram = D);
pseudo_hnf(pseudo_matrix(L));
lattice(ambient_space(L), ans);
ans == L
```

thofma:
Fix also the containment check which accepts a matrix.

### DIFF
--- a/src/NumFieldOrd/NfOrd/LinearAlgebra.jl
+++ b/src/NumFieldOrd/NfOrd/LinearAlgebra.jl
@@ -906,16 +906,23 @@ function _contained_in_span_of_pseudohnf(v::Generic.Mat{T}, a::S, P::PMat{T, S},
   end
   w = deepcopy(v)
   for i = start:step:stop
+    # find pivot
+    if shape === :upperright
+      piv = findfirst(k -> !iszero(P.matrix[i, k]), 1:ncols(P))::Int
+    else
+      piv = findlast(k -> !iszero(P.matrix[i, k]), 1:ncols(P))::Int
+    end
+
     if !(w[1, i]//P.matrix[i, i] in P.coeffs[i]*inv(a))
       return false
     end
-    e = w[1, i]//P.matrix[i, i]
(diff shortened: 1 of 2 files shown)


thofma:
This is why code duplication is bad.

Issue #545, opened 2022-01-19
Module operations for lattices over number fields


simonbrandhorst:
Some natural module operations are missing for modules over the maximal order of a number field. Namely,

- Intersection in case both lattices do not have full rank.
- Saturation

A trivial example:
```
  K, a = rationals_as_number_field()
  OK = maximal_order(K)
  q = quadratic_space(K, K[1 0; 0 1])
  L = fractional_ideal(OK, K(1//2))*lattice(q)
  S = lattice(q, matrix(generators(L)[1:1]))
  intersect(L, S)
ERROR: AssertionError: nrows(m) == length(c)
Stacktrace:
 [1] PseudoMatrix
```
Until we implement this properly using pseudo matrices a workaround using restriction of scalars would be good.

simonbrandhorst:
Closed via #866

PR #866, opened 2022-11-05
Intersection, saturation and orthogonal submodule of `AbsLat`

StevellM:
This is an implementation of intersection, saturation and orthogonal submodule for `HermLat` and `QuadLat`, using restriction of scalars. These work also in the case of non full-rank modules/pseudo-matrices.

This pull request shall replace the previously opened #857 .

I modified `orthogonal_complement` into `orthogonal_submodule` to fit with the function defined for `ZLat`... I hope the implementation does what it should do, @simonbrandhorst ?

At least the broken tests pass now.

### DIFF
--- a/src/QuadForm/Lattices.jl
+++ b/src/QuadForm/Lattices.jl
@@ -616,7 +616,7 @@ function _intersect_via_restriction_of_scalars(L::AbsLat, M::AbsLat)
   @assert has_ambient_space(L) && has_ambient_space(M)
   @assert ambient_space(L) === ambient_space(M)
   @assert !(L isa ZLat)
-  Lres, f = restrict_scalars_with_map(L)
+  Lres, f = restrict_scalars_with_map(L, FlintQQ)
   Mres = restrict_scalars(M, f)
   Nres = intersect(Lres, Mres)
   Bres = basis_matrix(Nres)
@@ -966,14 +966,19 @@ is_isotropic(L::AbsLat, p) = is_isotropic(rational_span(L), p)
 ################################################################################
 
 @doc Markdown.doc"""
-    restrict_scalars(L::AbsLat) -> ZLat
+    restrict_scalars(L::AbsLat, K::FlintRationalField,
+                                alpha::FieldElem = one(base_field(L))) -> ZLat
 
-Given a lattice `L`, return the $\mathbb Z$-lattice obtained by restriction of
-scalars.
+Given a lattice `L` in a space $(V, \Phi)$, return the $\mathcal O_K$-lattice
+obtained by restricting the scalars of $(V, \alpha\Phi)$ to the number field `K`.
+The rescaling factor $\alpha$ is set to 1 by default.
+
+Note that for now one can only restrict scalars to $\mathbb Q$.
 """
-function restrict_scalars(L::AbsLat)
+function restrict_scalars(L::AbsLat, K::FlintRationalField,
+                                     alpha::FieldElem = one(base_field(L)))
   V = ambient_space(L)
-  Vabs, f = restrict_scalars(V, FlintQQ)
+  Vabs, f = restrict_scalars(V, K, alpha)
   Babs = absolute_basis(L)
   Mabs = zero_matrix(FlintQQ, length(Babs), rank(Vabs))
   for i in 1:length(Babs)
@@ -986,14 +991,21 @@ function restrict_scalars(L::AbsLat)
 end
 
 @doc Markdown.doc"""
-    restrict_scalars_with_map(L::AbsLat) -> Tuple{ZLat, SpaceRes}
+    restrict_scalars_with_map(L::AbsLat, K::FlintRationalField,
+                                         alpha::FieldElem = one(base_field(L)))
+                                                        -> Tuple{ZLat, SpaceRes}
 
-Given a lattice `L`, return the $\mathbb Z$-lattice obtained by 
-restriction of scalars, together with the map `f` for extending scalars back.
+Given a lattice `L` in a space $(V, \Phi)$, return the $\mathcal O_K$-lattice
+obtained by restricting the scalars of $(V, \alpha\Phi)$ to the number field `K`,
+together with the map `f` for extending scalars back.
+The rescaling factor $\alpha$ is set to 1 by default.
+
+Note that for now one can only restrict scalars to $\mathbb Q$.
 """
-function restrict_scalars_with_map(L::AbsLat)
+function restrict_scalars_with_map(L::AbsLat, K::FlintRationalField,
+                                              alpha::FieldElem = one(base_field(L)))
   V = ambient_space(L)
-  Vabs, f = restrict_scalars(V, FlintQQ)
+  Vabs, f = restrict_scalars(V, K, alpha)
   Babs = absolute_basis(L)
   Mabs = zero_matrix(FlintQQ, length(Babs), rank(Vabs))
   for i in 1:length(Babs)
@@ -1008,8 +1020,12 @@ end
 @doc Markdown.doc"""
     restrict_scalars(L::AbsLat, f::SpaceRes) -> ZLat
 
-Given a lattice `L` and a map `f`  of restriction of scalars,
-return the $\mathbb Z$-lattice obtained by the map `f`.
+Given a lattice `L` in a space $(V, \Phi)$ and a map `f` for restricting the
+scalars of $(V, \alpha\Phi)$ to a number field `K`, where $\alpha$ is in the
+base algebra of `L`, return the associated $\mathcal O_K$-lattice obtained from
+`L` with respect to `f`.
+
+Note that for now one can only restrict scalars to $\mathbb Q$.
 """
 function restrict_scalars(L::AbsLat, f::SpaceRes)
   @req ambient_space(L) === codomain(f) "Incompatible arguments: ambient space of L must be the same as the codomain of f"
@@ -1497,7 +1513,7 @@ Return the largest submodule of `L` orthogonal to `M`.
 function orthogonal_submodule(L::AbsLat, M::AbsLat)
   @req ambient_space(M) == ambient_space(L) "Lattices must be in the same ambient space"
   V = ambient_space(L)
-  EM = basis_matrix_of_rational_span(primitive_closure(L, M))
+  EM = basis_matrix_of_rational_span(M)
   Morth = orthogonal_complement(V, EM)
   N = lattice(V, Morth)
   N = intersect(L, N)
@@ -1598,11 +1614,13 @@ maximal_integral_lattice(::AbsSpace)
 Given two lattices `M` and `N` defined over a number field `E`, with
 $N \subseteq E\otimes M$, return the primitive closure $M \cap E\otimes N$
 of `N` in `M`.
+
+One can also use the alias `saturate(L, M)`.
 """
 function primitive_closure(M::AbsLat, N::AbsLat)
   @assert has_ambient_space(N) && has_ambient_space(M)
   @req ambient_space(N) === ambient_space(M) "Lattices must be in the same ambient space"
-  Mres, f = restrict_scalars_with_map(M)
+  Mres, f = restrict_scalars_with_map(M, FlintQQ)
   Nres = restrict_scalars(N, f)
   Lres = primitive_closure(Mres, Nres)
   B = basis_matrix(Lres)
@@ -1610,3 +1628,9 @@ function primitive_closure(M::AbsLat, N::AbsLat)
   return lattice(ambient_space(M), B2)
 end
 
+@doc Markdown.doc"""
+    saturate(L::AbsLat, M::AbsLat) -> AbsLat
+
+Alias for `primitive_closure`.
+"""
+saturate(L::AbsLat, M::AbsLat) = primitive_closure(L::AbsLat, M::AbsLat)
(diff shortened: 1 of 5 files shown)


thofma:
Thanks!

I would suggest we do not make `intersect_via_restriction_of_scalars` and `saturate_via_restriction_of_scalars` part of the official API.

src/QuadForm/Lattices.jl
```diff
@@ -1471,20 +1489,21 @@ end
...
 @doc Markdown.doc"""
-    _orthogonal_complement(M::AbsLat, L::AbsLat) -> AbsLat
+    orthogonal_submodule(L::AbsLat, M::AbsLat) -> AbsLat
 
-Return the orthogonal complement of the lattice `L` inside the lattice `M`.
+Return the largest submodule of `L` orthogonal to `M`.
 """
-function _orthogonal_complement(M::AbsLat, L::AbsLat)
-  @req ambient_space(M) == ambient_space(L) "lattices must be in the same ambient space"
+orthogonal_submodule(L::AbsLat, M::AbsLat) = orthogonal_submodule(L, M)
```
thofma, line 1497: Hm, what is this method for?

thofma: Is this just some leftover?

StevellM: Ah no it is to get the documentation correctly called in the file for the html documentation, so that I can leave the signatures `AbsLat` while for the method I use ` T where T <: Union{HermLat, QuadLat}`

src/QuadForm/Lattices.jl
```diff
@@ -1471,20 +1489,21 @@ end
...
-  M = basis_matrix_of_rational_span(M)
-  Morth = orthogonal_complement(V, M)
-  # Now intersect KM with L
-  pm = _intersection_modules(pseudo_matrix(Morth), pseudo_matrix(L))
-  return lattice(V, pm)
+  EM = basis_matrix_of_rational_span(M)
+  Morth = orthogonal_complement(V, EM)
+  N = lattice(V, Morth)
+  N = saturate(N)
+  return intersect(L, N)
```
simonbrandhorst, line 1506: I think this won't work in some cases. For instance if `L` does not have integer coordinates. 
Also fails if `M` is degenerate since then it does not have an orthogonal complement (or at least `orthogonal_complement(V,EM)` should fail).

src/QuadForm/Lattices.jl
```diff
@@ -1569,3 +1588,37 @@ and which is maximal in `V` satisfying this property.
...
+#
+################################################################################
+
+@doc Markdown.doc"""
+    saturate(L::AbsLat) -> AbsLat
+
+Given a lattice `L` over a field `E`, return the closure $L\otimes E \cap O_E^n$
+of `L` in its rational span (see [`rational_span`](@ref)). Here $O_E$ is the base
+ring of `L` (see [`base_ring`](@ref)) and `n` is the degree of `L`.
+"""
```
simonbrandhorst, line 1603: I think saturation is more of a computational tool for (pseudo)-matrices. 
For lattices we called it `primitive_closure` and it takes two lattices as arguments. 
So you will need `_primitive_closure_via_restriction_of_scalars` or something similar.

src/QuadForm/Lattices.jl
```diff
@@ -1569,3 +1588,37 @@ and which is maximal in `V` satisfying this property.
...
+saturate(L::AbsLat) = _saturate_via_restriction_of_scalars(L)
+
+function _saturate_via_restriction_of_scalars(L::AbsLat)
+  @assert has_ambient_space(L)
+  bool = L isa ZLat
```
simonbrandhorst, line 1608: style wise you would rather define a method
`_saturate_via_restriction_of_scalars(L::ZLat)` 
instead of this kind of branching inside the method. It is more transparent when it is part of the signature.

src/QuadForm/Lattices.jl
simonbrandhorst: The upshot is: do not implement `saturation` but rather `primitive_closure`. 
But we can allow the alias `saturate(L::AbsLat, M::AbsLat) = primitive_closure(L, M)`

simonbrandhorst (changes requested):

src/QuadForm/Lattices.jl
StevellM: Don't we assume all our lattices to be non-degenerate, currently ? So what should be done instead, should I add for the moment an input check to ensure that the lattice `L` is integral ?

src/QuadForm/Lattices.jl
StevellM: I see but the problem is that I am afraid of "ambiguity" because formally, on Hecke, `ZLar` is a subtype of `AbsLat`, so I don't want to create ambiguity leading to problems having the method for `ZLat` and `AbsLat`. And if there are no ambiguity problems, then I don't understand how it works for Julia/Hecke to choose the right method to call...

src/QuadForm/Lattices.jl
simonbrandhorst: No, we do not assume this (though some things will break). That is the point of calling the function `orthogonal_submodule` and not `orthogonal_complement`.  
If you implement `primitive_closure`, you can use this to compute the `orthogonal_submodule` from the corresponding orthogonal subspace.

src/QuadForm/Lattices.jl
simonbrandhorst: This won't lead to an ambiguity since `ZLat <: AbsLat` is more specialized. Julia always looks for the most special method.

src/QuadForm/Lattices.jl
StevellM: I made changes regarding what you said: is it better now or still incomplete/not quite true ?

src/QuadForm/Lattices.jl
```diff
@@ -1471,20 +1489,19 @@ end
...
-  @req ambient_space(M) == ambient_space(L) "lattices must be in the same ambient space"
+function orthogonal_submodule(L::AbsLat, M::AbsLat)
+  @req ambient_space(M) == ambient_space(L) "Lattices must be in the same ambient space"
   V = ambient_space(L)
-  M = basis_matrix_of_rational_span(M)
-  Morth = orthogonal_complement(V, M)
-  # Now intersect KM with L
-  pm = _intersection_modules(pseudo_matrix(Morth), pseudo_matrix(L))
-  return lattice(V, pm)
+  EM = basis_matrix_of_rational_span(primitive_closure(L, M))
```
simonbrandhorst, line 1500:
```suggestion
  EM = basis_matrix_of_rational_span(M)
```

src/QuadForm/Lattices.jl
```diff
@@ -953,6 +965,12 @@ is_isotropic(L::AbsLat, p) = is_isotropic(rational_span(L), p)
 #
 ################################################################################
 
+@doc Markdown.doc"""
+    restrict_scalars(L::AbsLat) -> ZLat
+
+Given a lattice `L`, return the $\mathbb Z$-lattice obtained by restriction of
+scalars.
```
simonbrandhorst, line 972: Note that restriction of scalars for a quadratic_space has a different signature
```
restrict_scalars(V::AbsSpace, K::FlintRationalField,
                                       alpha = one(base_ring(V)))
```
Perhaps
```
restrict_scalars(V::AbsLat, K::FlintRationalField,
                                       alpha = one(base_ring(V)))
```
would be the better signature for this function?

simonbrandhorst: Could you please include a docstring for 
```
restrict_scalars(V::AbsSpace, K::FlintRationalField,
                                       alpha = one(base_ring(V)))
```

src/QuadForm/Lattices.jl
simonbrandhorst: Looks pretty good. Thank you :-)

src/QuadForm/Lattices.jl
StevellM: I guess it depends on what we want to do with restriction of scalars for `AbsLat`... Are we willing to leave to choice not to restrict to `QQ` but only a subfield of the `base_field` ? Is it relevant in practice ? If yes then I could make the change also for the one returning the map.

simonbrandhorst: It is not relevant for me but it could be relevant for others. Thus I like the syntax. It allows to add a restriction of scalars for subfields whenever we need to.
Please also add `saturate(X::AbsLat,Y::AbsLat) = primitive_closure(X, Y)`

StevellM: Yes but now that I look more at it, I realized that: `K::FlintRationalField` is part of the signatures but we don't have the choice... we can only have `QQ` for this. So in a sense, we put `K` as argument for where we restrict the scalars, but at the end of the day, we restrict everything, and everytime, to `QQ`

simonbrandhorst: Yes, the point is that in the future we can add a method for restriction to other fields without changing the syntax
and by simply writing a new method.  
We could allow `restriction(V::AbsSpace, alpha=...)` for restriction to `QQ` though. Any opinion @thofma ?

StevellM: I guess in general one would need to do a total change on `VecSpaceRes` and `SpaceRes` to allow restriction of scalars to smaller but non rational field, and then starts to change all the functions using restriction of scalars accoridngly. Could it be worth to dedicate a proper PR for this ?

simonbrandhorst: I would not implement it for now because we don't need it. But maybe someone will need it in the future. You never know.

thofma: I am not too big of a fan of `restriction(V::AbsSpace,...)`, as this could mean other things. Also I think either there is a function argument making it clear that one restricts to the rationals or it is indicated in the name.

For lattices, would you restrict scalars to QQ or to ZZ? So what about

```
restrict_scalars(V::AbsLat, K::FlintIntegerRing,
                                       alpha = one(base_ring(V)))
```

It is a bit annoying, since one will have to explain the `alpha` again.

StevellM: I have added appropriate documentation and made things consistent with the restriction of scalars for `AbsSpace`. I have left open the idea of extending this to any subfield of the base algebra there, so that users know what is available, and if someone needs it one day, it can be implemented without changing too much the format (maybe adapting the documentation and all the attached functions)

StevellM:
Just found out that this should also solve #545 .

src/QuadForm/Lattices.jl
```diff
@@ -953,9 +965,20 @@ is_isotropic(L::AbsLat, p) = is_isotropic(rational_span(L), p)
...
 ################################################################################
 
-function restrict_scalars(L::AbsLat)
+@doc Markdown.doc"""
+    restrict_scalars(L::AbsLat, K::FlintRationalField,
+                                alpha::FieldElem = one(base_field(L))) -> ZLat
+
+Given a lattice `L` in a space $(V, \Phi)$, return the $\mathcal O_K$-lattice
+obtained by restricting the scalars of $(V, \alpha\Phi)$ to the number field `K`.
+The rescaling factor $\alpha$ is set to 1 by default.
```
simonbrandhorst, line 974:
```suggestion
Given a lattice `L` in a space $(V, \Phi)$ over ``E``, return the $\mathcal O_K$-lattice
obtained by restricting the scalars of $(V, \alpha\Phi)$ to the number field `K`.
The form is defined by $\mathrm{Tr} \circ \Phi$ where $\mathrm{Tr}\colon E \to K$ is the trace.

The rescaling factor $\alpha$ is set to 1 by default.
```

src/QuadForm/Spaces.jl
```diff
@@ -588,9 +588,22 @@ end
...
+# to non rational subfields
+@doc Markdown.doc"""
+    restrict_scalars(V::AbsSpace, K::FlintRationalField,
+                                  alpha::FieldElem = one(base_ring(V)))
+                                                          -> QuadSpace, SpaceRes
+
+Given a space $(V, \Phi)$ and a subfield `K` of the base algebra `E` of `V`, return the
+quadratic space `W` obtained by restricting the scalars of $(V, \alpha\Phi)$ to `K`,
+together with the map `f` for extending the scalars back.
+The rescaling factor $\aplha$ is set to 1 by default.
```
simonbrandhorst, line 601:
```suggestion
The rescaling factor $\alpha$ is set to 1 by default.
```

src/QuadForm/Spaces.jl
```diff
@@ -588,9 +588,22 @@ end
...
+# TODO: Change VecSpaceRes/SpaceRes to allow restriction of scalars
+# to non rational subfields
+@doc Markdown.doc"""
+    restrict_scalars(V::AbsSpace, K::FlintRationalField,
+                                  alpha::FieldElem = one(base_ring(V)))
+                                                          -> QuadSpace, SpaceRes
+
+Given a space $(V, \Phi)$ and a subfield `K` of the base algebra `E` of `V`, return the
+quadratic space `W` obtained by restricting the scalars of $(V, \alpha\Phi)$ to `K`,
+together with the map `f` for extending the scalars back.
```
simonbrandhorst, line 600:
```suggestion
together with the map `f` for extending the scalars back.
The form on the restriction is given by ``Tr \circ \Phi`` where ``Tr: E \to K`` is the trace form.  
```

simonbrandhorst (approved):

PR #868, opened 2022-11-08
Temporarily remove some definitions

thofma:


thofma:
To get rid of the warnings in Oscar. We will add them back in once we do the next breaking release.

PR #869, opened 2022-11-08
Extract iteration logic out of meataxe function

mgkurtz:
In the meataxe algorithm, previously the lazy iteration over the irreducible factors of the minimal polynomial was intertwined with the actual algorithmic logic. This PR separates these concerns.

The iteration now contains a dispatch with a fallback for arbitrary rings, which are no finite fields. Similarly I added stupid provisional methods `random_coefficient(R::FinField) = rand(R)` and `random_coefficient(R::Ring) = R(rand(Int))` for use in the `meataxe`.

Overall those changes allow to call the `meataxe` for infinite fields as well. Though, of course with no guarantee on ever finishing. But the simple examples from the current tests now also work for some infinite fields.

There is still some work to do and questions to answer for this PR:
- [x] Shall the extracted methods really live in `ModAlgAss/MeatAxe.jl`? – Moved to `Poly/Misc.jl`.
- [ ] Where shall the meataxe be documented? – Not at all, but some other methods should.
- [x] What to do with the `random_coefficient` method? – Renamed. Can be improved later.
- [x] Which tests shall there be for the meataxe? – There are some at least.
- [x] How to fix the existing tests? – Done, see commit fixing `__split`.


thofma (commented):
> The iteration now contains a dispatch with a fallback for arbitrary rings, which are no finite fields. Similarly I added stupid provisional methods `random_coefficient(R::FinField) = rand(R)` and `random_coefficient(R::Ring) = R(rand(Int))` for use in the `meataxe`.

Maybe call them `random_element_for_meataxe` to make it clearer that they should be used only here.
 
> There is still some work to do and questions to answer for this PR:
> 
> * [ ]  Shall the extracted methods really live in `ModAlgAss/MeatAxe.jl`? – I hope, others can tell me.

You can put them into `Misc/Poly.jl`.

> * [ ]  Where shall the meataxe be documented? – I do not know.

The meataxe itself should not be user-facing. But the `irreducible_modules` methods etc should be documented together with the modules.

> * [ ]  What to do with the `random_coefficient` method? – I hope for inspiration.

In what sense? Maybe rename it as suggested above.

> * [ ]  Which tests shall there be for the meataxe? – I can try finding some.

The following should be reducible for rationals or extensions thereof:

```
G = [matrix(QQ, 4, 4, [ 0, 1, 0, 0, -1, -1, 0, 0, 0, 0, 0, 1, 0, 0, -1, -1 ))]
M = Hecke.Amodule(G)
```
> * [ ]  How to fix the existing tests? – I can probably do that.

Which ones?

src/ModAlgAss/MeatAxe.jl
```diff
@@ -234,8 +234,9 @@ end
 
 #  Restriction of the action to the submodule generated by C and the quotient
 function __split(C::T, G::Vector{T}) where T <: MatElem
-  action_sub = Vector{T}(undef, length(G))
-  action_quo = Vector{T}(undef, length(G))
+  View = only(Base.return_types(view, (T, UnitRange{Int}, UnitRange{Int})))
```
thofma, line 237: Hm, this is a pattern we try to avoid, since `return_types` is not documented and part of any official interface. So it could vanish in any upcoming version or could start returning just `Any`.

One either has to introduce a helper function computing the type or refactor the loop below.

thofma (changes requested):

thofma:
Apart from the one issue above, I am happy to merge this.

thofma (approved):
Will merge once CI shows green.

PR #874, opened 2022-11-11
More rcf

fieker:
fixes the "next" problems...


fieker:
On Sat, Nov 12, 2022 at 04:27:29AM -0800, Tommy Hofmann wrote:
> Merged #874 into master.
> 
many thanks - but too early....

just fond an error in the conductor

PR #881, opened 2022-11-15
Improve computation of complex conjugation

thofma:


fieker:
You can probably do even better by decomposing into cyclic components....

fieker:
I think the difference is that aut(classified) does not create all of them, while aut(number field) does

PR #896, opened 2022-11-25
Orthogonal sums with injections and projections

StevellM:
I guess this should be here temporarily, until we decide what we exactly do with `orthogonal_sum`. Things is that I would need the part for `TorQuadMod` on Oscar. Since it can be made on Hecke and could be of general use, I decided to implement this here.

Ideally with this pattern, we could have 4 functions each time: one without maps, one with injections, one with projections and one with both.

### DIFF
--- a/src/QuadForm/Spaces.jl
+++ b/src/QuadForm/Spaces.jl
@@ -703,9 +703,9 @@ function orthogonal_sum(V::HermSpace, W::HermSpace)
 end
 
 function _orthogonal_sum_with_injections_and_projections(x::Vector{<:QuadSpace})
-  @assert length(x) >= 2
+  @req length(x) >= 2 "Input must contain at least two quadratic spaces"
   K = base_ring(x[1])
-  @assert all(i -> base_ring(x[i]) === K, 2:length(x))
+  @req all(i -> base_ring(x[i]) === K, 2:length(x)) "All spaces must be defined over the same field"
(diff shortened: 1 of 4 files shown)


thofma:
Seems reasonable to me, thanks

thofma (approved):

src/QuadForm/Spaces.jl
```diff
@@ -702,6 +702,33 @@ function orthogonal_sum(V::HermSpace, W::HermSpace)
   return VplusW, f1, f2
 end
 
+function _orthogonal_sum_with_injections_and_projections(x::Vector{<:QuadSpace})
+  @assert length(x) >= 2
+  K = base_ring(x[1])
+  @assert all(i -> base_ring(x[i]) === K, 2:length(x))
```
simonbrandhorst, line 708:
```suggestion
  @req all(i -> base_ring(x[i]) === K, 2:length(x)) "lattices not defined over the same field."
```

StevellM: Do you want I make all the `@assert` into `@req` ? Since they were supposedly internal functions I just wanted to make sure we put the write input. But if you want I can make them in all the functions as `@req`.

simonbrandhorst: Apparently `@assert` can get optimized away by the compiler. For input checking `@req` is probably better.

PR #897, opened 2022-11-29
Add pre and postinverse for abelian groups

thofma:
CC: @HereAround

### DIFF
--- a/src/GrpAb/Map.jl
+++ b/src/GrpAb/Map.jl
@@ -345,7 +345,7 @@ Returns whether $h$ is surjective.
     return (order(codomain(A)) == order(H))::Bool
   else
     KK = cokernel(A)[1]
-    return (is_finite(KK) && order(KK) == 1)::Bool
+    return is_finite(KK) && isone(order(KK))
   end
 end
 


src/GrpAb/Map.jl
```diff
@@ -496,3 +497,28 @@ function hom(G::GrpAbFinGen, H::GrpAbFinGen; task::Symbol = :map)
...
+function preinverse(f::GrpAbFinGenMap)
+  @req is_surjective(f) "Map must be surjective"
+  return hom(codomain(f), domain(f), _prepostinverse(f))
+end
+
+# Given surjective f, find g such that gf = id
+function postinverse(f::GrpAbFinGenMap)
+  @req is_injective(f) "Map must be injective"
+  return hom(codomain(f), domain(f), _prepostinverse(f))
+end
```
fieker, line 524: both return the same? A more efficient version of pseudo_inverse? (well a linear version of pseud_inverse)?

thofma: It is just by accident that we can do the "same" (because `haspreimage` returns two things). It is a more mathematical version of `pseudo_inverse`.

fieker:
On Tue, Nov 29, 2022 at 12:24:00AM -0800, Tommy Hofmann wrote:
> 
> 
> 
> > +  # if it does not exist, we find any element of the domain,
> +  # which is also fine
> +  return [haspreimage(f, g)[2] for g in gens(codomain(f))]
> +end
> +
> +# Given surjective f, find g such that fg = id
> +function preinverse(f::GrpAbFinGenMap)
> +  @req is_surjective(f) "Map must be surjective"
> +  return hom(codomain(f), domain(f), _prepostinverse(f))
> +end
> +
> +# Given surjective f, find g such that gf = id
> +function postinverse(f::GrpAbFinGenMap)
> +  @req is_injective(f) "Map must be injective"
> +  return hom(codomain(f), domain(f), _prepostinverse(f))
> +end
> 
> It is just by accident that we can do the "same" (because `haspreimage` returns two things). It is a more mathematical version of `pseudo_inverse`.
> 
> -- 
> Reply to this email directly or view it on GitHub:
> https://github.com/thofma/Hecke.jl/pull/897#discussion_r1034430678
> You are receiving this because you commented.

A section? - I am happy with it as is

thofma:
Yes, one of them is a section, the other a retraction.

src/GrpAb/Map.jl
```diff
@@ -339,12 +339,13 @@ end
...
 """
-function is_surjective(A::GrpAbFinGenMap)
+@attr Bool function is_surjective(A::GrpAbFinGenMap)
   if isfinite(codomain(A))
     H, mH = image(A)
     return (order(codomain(A)) == order(H))::Bool
   else
-    return (order(cokernel(A)[1]) == 1)::Bool
+    KK = cokernel(A)[1]
+    return (is_finite(KK) && order(KK) == 1)::Bool
```
HereAround, line 348:
```suggestion
    return is_finite(KK) && isone(order(KK))
```

HereAround: I am also not sure if the conversion into `Bool` is strictly necessary. You probably know that better than I do.

HereAround:
@thofma Thank you for quickly adding. Looks good to me. I left a minor comment above.

HereAround (approved):

src/GrpAb/Map.jl
thofma: You are right. It is not necessary anymore

PR #899, opened 2022-11-30
23 constructions of the leech lattice

simonbrandhorst:


thofma:
At the moment doctests do not count towards coverage I think. I will have to adjust the setup slightly to include this.

simonbrandhorst:
Yea, that would safe me some work.

PR #911, opened 2022-12-06
Add `sparse_matrix(R,n,m)` as alternative to zero_matrix(SMat,R,n,m)

fingolfin:
I discussed with @HechtiDerLachs today how to create sparse matrices, and it took us a bit to determine that we should use `zero_matrix(SMat,R,n,m)` to specify an initial column count (i.e., we used `n=0` and set `m` to the final column count).

We both expected `sparse_matrix` to allow us to specify a row and/or column count. Based on this, my suggestion was to just add that. Instead of writing an issue, here is a proof of concept. If this is deemed acceptable, I can finish it up by linking the new constructor into the manual and adding a test.

### DIFF
--- a/docs/src/sparse/intro.md
+++ b/docs/src/sparse/intro.md
@@ -90,6 +90,7 @@ As a consequence, unlike their dense counterparts, sparse matrices have a mutabl
 ### Construction
 ```@docs
 sparse_matrix(::Ring)
+sparse_matrix(::Ring, ::Int, ::Int)
 ```
 
 Sparse matrices can also be created from dense matrices as well as from julia arrays:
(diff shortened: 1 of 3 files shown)


fieker:
I don't think this is any good. What is the use case? Creating a zero matrix and setting entires? This will be slow as the sparse_matrices cannot add entries (efficiently). The data structure is designed around creating sparse_rows and appending them to the matrix. Any attempt to use the sparse matrix as an inplace replacement for a dense one is doomed to be inefficient...

thofma:
This seems fine. And I agree that `sparse_matrix(R, 0, n)` is a good idea. Coincidentally @esabo asked something similar yesterday.

fingolfin:
Oops, I had forgotten about this one...

I've now put the new method into the docs (though perhaps I should put it elsewhere, suggestions welcome), and added trivial tests.

fingolfin:
The failure in Julia 1.8 on windows looks like a fluke?

thofma:
Yes

PR #925, opened 2022-12-14
More on `_basis_of_commutator_algebra`

StevellM:
- Add the possibility to have different order for the matrices on the left and the right.
- Add some tests from representation theory.

### DIFF
--- a/src/AlgAssAbsOrd/Conjugacy/Conjugacy.jl
+++ b/src/AlgAssAbsOrd/Conjugacy/Conjugacy.jl
@@ -441,37 +441,39 @@ end
 #
 ################################################################################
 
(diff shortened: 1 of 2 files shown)


src/AlgAssAbsOrd/Conjugacy/Conjugacy.jl
```diff
@@ -542,18 +557,24 @@ end
 function _basis_of_integral_commutator_algebra(As::Vector{fmpq_mat},
                                                Bs::Vector{fmpq_mat})
```
thofma, line 558: This function is not tested I think. Since you changed it, could you add a test to make sure it still works?

thofma (approved):

thofma:
Thanks!

PR #937, opened 2023-01-11
`automorphism_group_generators` for rank 2 indefinite `ZLat`

StevellM:
This was a missing corner case that can be handled using algorithms for binary quadratic forms.

Note: there was a unexpected behaviour of `automorphism_group_generators(::QuadBin{fmpz})`, so I decided to export the function locally in the file `src/QuadForm/Quad/GenusRep.jl`.


thofma:
Looks good to me, thanks! @simonbrandhorst any comment?

simonbrandhorst:
Thank you.

PR #938, opened 2023-01-13
Change `generic_completion` to `completion`

thofma:


thofma:
@fieker as we discussed it

PR #941, opened 2023-01-16
Local h2

fieker:


thofma:
Also needs the `completion <-> generic_completion` change.

PR #943, opened 2023-01-17
Isometry between `TorQuadMod`'s in the non split degenerate case

StevellM:
We cover the last case which was not covered by the previous algorithms, namely isometry test between two `TorQuadMod` which are not semi-regular and which do not split their respective radical quadratic. It uses partial isometries.

### DIFF
--- a/src/QuadForm/Torsion.jl
+++ b/src/QuadForm/Torsion.jl
@@ -901,7 +901,7 @@ function _isometry_non_split_degenerate(T::TorQuadMod, U::TorQuadMod)
     card = prod([order(Ts[k]) for k in 1:(i+1)], init = ZZ(1))
     for u in u_cand[i+1]
       if all(k -> u*f[k] == t*Ts[k], 1:i)
-        fnew = [j for j in f]
+        fnew = copy(f)
         push!(fnew, u)
         if order(sub(U, fnew)[1]) == card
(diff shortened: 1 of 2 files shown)


src/QuadForm/Torsion.jl
```diff
@@ -878,6 +883,35 @@ function _isometry_degenerate(T::TorQuadMod, U::TorQuadMod)
...
+    i = length(f)
+    if i == n
+      return (true, compose(inv(TstoT), hom(Ts, U, f)))
+    end
+
+    t = Ts[i+1]
+    card = prod([order(Ts[k]) for k in 1:(i+1)], init = ZZ(1))
+    for u in u_cand[i+1]
+      if all(k -> u*f[k] == t*Ts[k], 1:i)
+        fnew = [j for j in f]
```
simonbrandhorst, line 904: looks like a `deepcopy` might be more appropriate?

simonbrandhorst (approved):

src/QuadForm/Torsion.jl
thofma: `copy` should be also enough.

StevellM: I had a problem of parents at some point, I thought it was coming from this.. apparently not, I will make the change

PR #945, opened 2023-01-17
fix precision issue in normal form

simonbrandhorst:
Some time ago I thought I was clever and set the precision lower than I did in sage. Turns out I am not that clever.


thofma:
Tests are failing

PR #958, opened 2023-02-03
Trivia

fieker:


thofma:
Do you have the example for `prime_dec_gen`?

fieker:
It's 5 and 29 in the corresponding pmax order of the splitting field of
quintic. But the splitting field is only nice to compute using Galois in
Oscar and this also unearthed problems...
The field is deg 120 with huge coeffs

fieker:
In released oscar do
k,a = number_field(f)
G,S = galois_group(k, pStart=1000)
fixed_field(S, sub(G, [one(G)])[1])

And despair

thofma:
Can you post the quintic?

fieker:
-4*x^5 + 4*x^4 + 24*x^3 + 5*x^2 - 10*x + 5

thofma:
I am not running into the problem that the minimum is not set.

I can reproduce the other error in the `anti_uniformizer` call.

fieker:
I guess I ran into the minimum prob in the non prime_dec_gen path, that why
I also added the is maximal and know stuff to force it through the gen

PR #960, opened 2023-02-06
Add divisibility in `ZLat`

StevellM:
I do not know whether it is the most efficient way, but it should do the job.

### DIFF
--- a/src/QuadForm/Quad/ZLattices.jl
+++ b/src/QuadForm/Quad/ZLattices.jl
@@ -1401,8 +1401,7 @@ function Base.in(v::fmpq_mat, L::ZLat)
 end
 
 @doc Markdown.doc"""
-    is_primitive(L::ZLat, v::Union{Vector, fmpq_mat})
-                                                   -> Bool
+    is_primitive(L::ZLat, v::Union{Vector, fmpq_mat}) -> Bool
 
 Return whether the vector `v` is primitive in `L`.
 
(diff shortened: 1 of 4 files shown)


thofma:
What is the divisibility of something? Can you share a reference?

StevellM:
Hum I do not have a reference. If `v` is a primitive vector in a `ZLat` `(L,b)`, then the divisibility of `v` in `(L, b)` is the positive generator of the $\mathbb Z$-ideal $b(v, L)$.

thofma:
Could you add this sentence to the documentation?

src/QuadForm/Quad/ZLattices.jl
```diff
@@ -1392,13 +1393,63 @@ end
...
+
+For a primitive vector `v` in a $\mathbb Z$-lattice `L`, we
+call the divisibility of `v` in `L` the positive generator
+of the $\mathbb Z$-ideal $b(v, L)$, where `b` is the bilinear
+form associated to `L`.
+"""
+divisibility(::ZLat, ::Union{Vector, fmpq_mat})
+
+function divisibility(L::ZLat, v::Vector{<: RationalUnion})
+  @req is_primitive(L, v) "v must be a primitive vector in L"
```
simonbrandhorst, line 1440: No need to assume primitive.
```suggestion
Return the divisibility of `v` with respect to `L`.

For a vector `v` in in the ambient space $(V,\Phi)$ of `L`, we
call the divisibility of `v` with respect to `L` the nonnegative generator
of the fractional $\mathbb Z$-ideal $\Phi(v, L)$.
"""
divisibility(::ZLat, ::Union{Vector, fmpq_mat})

function divisibility(L::ZLat, v::Vector{<: RationalUnion})
```

PR #970, opened 2023-02-17
guard some more valuations

fieker:


fieker:
should help https://github.com/oscar-system/Oscar.jl/actions/runs/4193097276/jobs/7269536888#step:6:925

thofma:
The identical check is three lines above, or am I missing something? I am not sure how this will help.

PR #971, opened 2023-02-17
Add the computation of the image of the logarithm

thofma:


src/LocalField/LocalField.jl
```diff
@@ -499,3 +499,46 @@ end
...
+      continue
+    end
+    push!(res, logg)
+  end
+  if length(C) == length(res)
+    return 1, [zero(K)]
+  else
+    return n, res
+  end
+end
```
fieker, line 544: is this better/ differnt than using map(log, one_unit_group_gens)? Just curious..

thofma: I thought about that. What you propose gives me $\mathbf{Z}_p$-generators of $\log(U^(1))$, but I actually need the decomposition into cosets of $\mathfrak{p}^n$.

fieker:
On Fri, Feb 17, 2023 at 06:36:48AM -0800, Tommy Hofmann wrote:
> 
> 
> 
> > +  pi = uniformizer(K)
> +  pipowers = [pi^i for i in 1:(n - 1)]
> +  res = elem_type(K)[]
> +  for c in C
> +    logg = log(1 + sum(c[i] * pipowers[i] for i in 1:(n - 1)))
> +    if any(x -> iszero(x - logg) || e * valuation(x - logg) >= n, res)
> +      continue
> +    end
> +    push!(res, logg)
> +  end
> +  if length(C) == length(res)
> +    return 1, [zero(K)]
> +  else
> +    return n, res
> +  end
> +end
> 
> I thought about that. What you propose gives me $\mathbf{Z}_p$-generators of $\log(U^(1))$, but I actually need the decomposition into cosets of $\mathfrak{p}^n$.

Isn't that "just" the Z_p module mod pi^n - as I said just curious...
I'm happy merge/ approve/ whatever

thofma:
Yes, you are right. What I basically want is all the elements of `log(U^(1)/U^(n)) = log(U^(1))/p^n`. I already made it til the `log(U^(1))` part, but was not sure how to proceed given our tools.

PR #974, opened 2023-02-22
Fail CI on deprecation warning

HereAround:
The same as https://github.com/oscar-system/Oscar.jl/pull/1957, i.e. the tests should now fail once a deprecation warning is encountered.

cc @thofma @fieker @fingolfin


thofma:
Thanks!

PR #976, opened 2023-02-22
Rename RealField(Elem) -> RealNumberField(Elem)

fingolfin:
... in preperation for the "big renaming": in the next release, Nemo should have its own `RealField` and `RealElemField`.

These names could perhaps still be improved, but at least the connection to number fields is now clearly visible, and @fieker seems OK with it.


fingolfin:
@fieker @thofma if this is OK for you, please merge once/if the tests pass. It'll be part of my upcoming renaming PR anyway

fingolfin:
(but I wanted to get it out before that, in the hopes this might speed up things)

fingolfin:
OK perhaps I should also add aliases for backwards compatibility? I am trying to find out if anyone uses them (@micjoswig maybe?). Just now @thofma told me on slack:
> This is superseded
> I think I asked Claus a while ago whether I can remove it or not

So perhaps this PR is redundant and we should instead *remove* this code? Please advice.

micjoswig:
> OK perhaps I should also add aliases for backwards compatibility? I am trying to find out if anyone uses them (@micjoswig maybe?). Just now @thofma told me on slack:
> 
> > This is superseded
> > I think I asked Claus a while ago whether I can remove it or not
> 
> So perhaps this PR is redundant and we should instead _remove_ this code? Please advice.

I am not directly affected.  Please check: @antonydellavecchia  @benlorenz @alexej-jordan

fingolfin:
There is some serialization code for `RealField` and `RealElem` in Oscar, but that's trivial to adjust (esp. if nothing actively uses it right now?)

thofma:
> I don't use these types. We do use `embedded_number_field` (`EmbeddedNumField`) instead which might be what this was replaced with?

Yes, this is the correct one. I remember we talked about it, but I did not find it in the Oscar codebase.

PR #981, opened 2023-02-23
Simplify some macro implementations

fingolfin:


thofma:
Seems to be some bad usages in Oscar too.

fingolfin:
Yeah, I just submitted https://github.com/oscar-system/Oscar.jl/pull/1975

PR #982, opened 2023-02-23
Some renamings

StevellM:
Renamings done:
- `AbsSpace` -> `AbstractSpace` (`Abs` is reserved for `Absolute`)
- `AbsSpaceMor` -> `AbstractSpaceMor` (same as above)
- `AbsLat` -> `AbstractLat` (same as above)
- `TorQuadMod` -> `TorQuadModule` (modules are called `Module`)
- `TorQuadModElem` -> `TorQuadModuleElem` (same as above)
- `TorQuadModMor` -> `TorQuadModuleMor` (same as above)
- `LatticeDB` -> `LatDB` (lattices defined as `Lat` except group lattices (which are not the same kind of lattices))
- `NfLattice` -> `NfLat` (same as above)

Some other renamings for genera, in order to have something consistent. I chose `ZGenus` as a standard, and so:
- `genera` -> `Zgenera` ( the enumeration function)
- `GenusHerm` -> `HermGenus` (since it is a genus, we put `Genus` at the end)
- `LocalGenusHerm` -> `HermLocalGenus` (same as before)
- `GenusQuad` -> `QuadGenus` (same as before)
- `LocalGenusQuad` -> `QuadLocalGenus` (same as before)
- `genera_hermitian` -> `hermitian_genera`
- `local_genera_hermitian` -> `hermitian_local_genera`
- `genera_quadratic` -> `quadratic_genera`
- `local_genera_quadratic` -> `quadratic_local_genera`

### DIFF
--- a/docs/src/quad_forms/Zgenera.md
+++ b/docs/src/quad_forms/Zgenera.md
@@ -37,7 +37,7 @@ genus(A::MatElem)
 ### Enumeration of genus symbols
 
 ```@docs
-genera(sig_pair::Tuple{Int,Int}, determinant::Union{Int,fmpz})
+Zgenera(sig_pair::Tuple{Int,Int}, determinant::Union{Int,fmpz})
 ```
 ### From other genus symbols
 ```@docs
--- a/docs/src/quad_forms/basics.md
+++ b/docs/src/quad_forms/basics.md
@@ -43,14 +43,14 @@ The *discriminant* $\text{disc}(V, \Phi)$ of $(V, \Phi)$ is defined to be
 $(-1)^{(m(m-1)/2)}\text{det}(V, \Phi)$, where $m$ is the rank of $(V, \Phi)$.
 
 ```@docs
-rank(::AbsSpace)
-dim(::AbsSpace)
-gram_matrix(::AbsSpace)
-involution(::AbsSpace)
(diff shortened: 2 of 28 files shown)


fingolfin:
@StevellM error:
```
Torsion: Error During Test at /home/runner/work/Hecke.jl/Hecke.jl/test/QuadForm/Torsion.jl:1
  Got exception outside of a @test
  UndefVarError: genera not defined
  Stacktrace:
    [1] macro expansion
      @ ~/work/Hecke.jl/Hecke.jl/test/QuadForm/Torsion.jl:203 [inlined]
```

fingolfin:
Minor nitpick: as far as I can tell, `Zgenera` is a function, so it shouldn't contain upper case letters. Maybe `zz_genera`?

That said, I'd be fine with this being merged as-is (@thofma?) before I regenerate my PRs, then we avoid dealing with conflicts here.

thofma:
Are the `Quad -> Quadratic` etc abbreviations written down somewhere?

StevellM:
Mmmh no, since there have been no discussions about them, I left this part as it was. If it seems clear that we should claim the abbreviation `Quad` for `Quadratic`, then it should be written down on the HackMD I guess

PR #977, opened 2023-02-22
The big renaming

fingolfin:
Companion to https://github.com/Nemocas/AbstractAlgebra.jl/pull/1267, https://github.com/Nemocas/Nemo.jl/pull/1376, https://github.com/oscar-system/Oscar.jl/pull/1972

PR #983, opened 2023-02-23
NumberField -> number_field

fieker:
doc strings for types
removal of some very dodgy nmod_struct


thofma:
Was this included in #977?

fieker:
it wasn't...

src/HeckeTypes.jl
```diff
@@ -371,6 +403,11 @@ mutable struct SMatSpace{T} <: Ring
   end
 end
 
+"""
+  SMat{T}
```
fingolfin, line 407:
```suggestion
    SMat{T}
```

fieker: when the test have passed I accept that change and merge

Issue #919, opened 2022-12-13
Document `@req`


fingolfin:
In https://github.com/oscar-system/Oscar.jl/pull/1811 I saw code using `@req` and wondered what it does. Had to grep around a bit. Would have been nice if `?@req` had cleared it up :-). Perhaps we can document it and then use it in more places?

thofma:
Without any bias I fully support this proposal.

thofma:
Fixed by #984

PR #984, opened 2023-02-23
Documentation for the macros in `Assertions.jl`

StevellM:
This is a suggestion for documentations of the macros in the file `Assertions.jl`.
It is not perfect, and namings can be discussed, nothing is strict.

I have added also something on the online manual: I have not been able to render it on my local machine so I cannot say whether or not it "looks good". Same as before, the namings are not perfect neither definitive. I am open to change these.

I have also moved the export of `@vtime` to the correct file.


StevellM:
This should also closed #919

docs/src/features/macros.md
```diff
@@ -0,0 +1,61 @@
+# Macros
+```@meta
+CurrentModule = Hecke
+DocTestSetup = quote
+    using Hecke
+  end
+```
+
+We refer here to the extra macros available on Hecke.
```
fingolfin, line 9:
```suggestion
We describe here various macros provided by Hecke.
```

docs/src/features/macros.md
```diff
@@ -0,0 +1,61 @@
...
+get_assert_level(s::Symbol)
+```
+
+### Check
+
+```@docs
+@hassert
+```
+
+## Misc
```
fingolfin, line 57:
```suggestion
## Miscellaneous
```

src/Assertions.jl
```diff
@@ -89,6 +86,64 @@ function _global_indent()
...
+@doc Markdown.doc"""
+    @vprint(S::Symbol, k::Int, msg::String)
+    @vprint S k msg
+
+    @vprint(S::Symbol, msg::String)
+    @vprint S msg
+
+This macro can be used to control printings inside the code.
+
+The macro `vprint` has two or three entry arguments: a symbol `S` refered as a
```
fingolfin, line 97:
```suggestion
The macro `vprint` takes two or three arguments: a symbol `S` specifying a
```

src/Assertions.jl
```diff
@@ -89,6 +86,64 @@ function _global_indent()
...
+*verbose scope*, an optional integer `k` and a string `msg`. If `k` is not
```
fingolfin, line 98: This is of course not your fault, but: "verbose scope" and "verbose level" IMHO really should be "verbosity scope" and "verbosity level". But of course that's something @thofma and @fieker should decide.

src/Assertions.jl
```diff
@@ -89,6 +86,64 @@ function _global_indent()
...
+specified, it is set by default to $1$.
+
+To each verbose scope `S` is associated a *verbose level* `l` which is cached.
```
fingolfin, line 101:
```suggestion
To each verbose scope `S` is associated a *verbose level* `l`.
```
What do you mean by "is cached"? Maybe that it is persistent?

thofma:
If you could rename them as @fingolfin suggested, e.g. `verbosity_scope` etc, this would be great.

StevellM:
Sure, it was ready to go, I was waiting for extra approval as Max suggested.

src/Assertions.jl
```diff
@@ -125,11 +235,61 @@ macro v_do(s, l::Int, action)
   end
 end
 
+@doc Markdown.doc"""
+    set_verbose_level(s::Symbol, l::Int) -> Int
```
fingolfin, line 239: Also rename this and other such functions? I.e.:
```suggestion
    set_verbosity_level(s::Symbol, l::Int) -> Int
```

src/Assertions.jl
```diff
@@ -125,11 +235,61 @@ macro v_do(s, l::Int, action)
...
+`s` to `l`.
+
+One can access the current verbosity level of `s` by calling the function
+[`get_verbose_level`](@ref).
+
+If `s` is not yet known as a verbosity scope, the function raises an `ErrorException`
+showing the error message "Not a valid symbol". One can add `s` to the list of
+verbosity scopes by calling the function [`add_verbose_scope`](@ref).
+
+# Example
```
fingolfin, line 251: The convention (at least in Oscar, and Julia itself) is to always call this section "Examples", so:
```suggestion
# Examples
```

src/Assertions.jl
```diff
@@ -196,11 +352,96 @@ function set_assert_level(s::Symbol, l::Int)
...
 end
 
+@doc Markdown.doc"""
+    @hassert(S::Symbol, k::Int, assert::Expr)
+    @hassert S k assert
+
+    @hassert(S::Symbol, assert::Expr)
+    @hassert S assert
+
+This macro can be used to control assertions check inside the code.
```
fingolfin, line 395:
```suggestion
This macro can be used to control assertion checks inside the code.
```

src/Assertions.jl
StevellM: Well of course, I can also maybe change the `assert` to `assertion` in the similar way, if I transform the `verbose` into `verbosity` ? I will add some aliases for backwards compatibility (if used on Oscar)

thofma:
Good to go?

StevellM:
For me it is.

PR #985, opened 2023-02-23
Changes in direct sums construction

StevellM:
We try here to suggest a consistent patterns for direct sums/direct products/biproducts for some objects which abstractly have biproducts. The general idea is:

Those objects have finite direct sums and finite direct products which agree, and which are called biproducts. We leave as an interface to the user the following choices:
- construct the biproduct as a `direct_sum` and only get as extra output the injection maps;
- construct the biproduct as a `direct_product` and only get as extra output the projection maps;
- construct the biproduct as a `biproduct` and return the injection maps as well as the projection maps.

Moreover, for the objects with "orthogonality relations" i.e. `AbstractLat`, `AbstractSpace` and `TorQuadModule`, we get rid of `orthogonal_sum` since this was misleadingly a `direct_sum` in disguise.

For consistency, we rename anything in this context named `orthogonal_sum` as `direct_sum`.

We also provide for `TorQuadModule`  a normal sum `+`, similarly to what is possible for `ZLat`, whenever it makes sense (this could be seen as an "inner sum")


thofma:
Needs rebase

thofma:
P.S.: Also documentation build is failing.

PR #989, opened 2023-02-26
Towards one finite field

thofma:


thofma:
@benlorenz  Since you have experience with interesting bugs, you may have an idea how to track this down. In this PR here I am switching to use more of `fq_default` from flint. For some reason, on ubuntu, the runtime increases dramatically, e.g., https://github.com/thofma/Hecke.jl/actions/runs/4479361302/jobs/7880155841#step:6:209.
It seems there is some problem with memory. I had hoped that it would be fixed by https://github.com/Nemocas/Nemo.jl/commit/59bdb9ad8e10c36508ceaf4a39939bdca51e9f07, but no luck so far. The biggest problem is that I have a hard time reproducing it/debugging it. Do you have any idea how to track this down?

PR #990, opened 2023-02-27
fix carmichael_lambda

wjyoumans:
`carmichael_lambda` gave an error on input 1 and the wrong result for input 4, 24, 48, etc, this should fix all methods.


thofma:
Thanks @wjyoumans. @fieker does that look good?

fieker:
one can optimize the code by using init, but essentially it looks good and I'd merge it as it.

PR #993, opened 2023-02-28
Avoid .base_ring field of ZZMatrix, ZZPoly, QQPoly

fingolfin:


fingolfin:
I did not realize that `is_zero_entry` is super-new in Nemo and not yet in a release. Will adapt.

PR #994, opened 2023-02-28
Remove is_zero_entry for ZZMatrix, it is already in Nemo

fingolfin:


fingolfin:
The documentation CI test fails, but this failure is already present on master, AFAICT it is completely unrelated to this PR.

PR #1000, opened 2023-03-14
Fix variable name in `morphism` of `ModAlgAss`

mgkurtz:
The `morphism` function for `ModAlgAss` just did not work due to `M` erroneously being called `matrix`:

```julia
m1 = matrix(QQ, [1;;])
M1 = Amodule([m1])
Hecke.morphism(M1, M1, m1)
```

results in

    ERROR: MethodError: no method matching *(::QQMatrix, ::typeof(matrix))


thofma:
Thanks!

PR #1004, opened 2023-03-16
Miscellaneous

StevellM:
Some little things I have done on my local codes that could be done. I have removed the deprecated codes which are not used anymore in Oscar, I have added some documentation and an export.

I have added `coinvariant_lattice` to go along with `invariant_lattice` and the possibility to check whether a `TorQuadModuleElem` has a preimage along a `TorQuadModuleMor` (a `has_preimage` function)


thofma:
Looks good, but tests are failing.

PR #1007, opened 2023-03-18
jldoctest: add some missing end markers

benlorenz:
(lets hope that the tests succeed)

x-ref: https://github.com/oscar-system/Oscar.jl/pull/2083


fingolfin (approved):
Looks good to me, but @thofma will have to approve

PR #1012, opened 2023-03-26
Fix additon of sparse matrices

thofma:


thofma:
Closes https://github.com/oscar-system/Oscar.jl/issues/2135

PR #1014, opened 2023-03-26
Allow `diagonal_matrix(QQ, [1, -1, -1])`

fingolfin:
Resolves https://github.com/oscar-system/Oscar.jl/issues/2061


thofma:
`NCRing` is not exported

PR #1025, opened 2023-03-31
Fix generic `_hermitian_form_with_invariants`

StevellM:
This was a piece of code which I guess was not triggered by the tests we currently had.
At first it was suppose to be a fix for `representative`, but the example I had is taking forever, so I just restricted to computing the middle step which was bugged (to avoid to have a test running for 20min because of the computation of a maximal integral lattice)

For help I have added another `to_hecke` function for `HermGenus`.

### DIFF
--- a/src/QuadForm/Herm/Genus.jl
+++ b/src/QuadForm/Herm/Genus.jl
@@ -1,7 +1,7 @@
 export genus, representative, rank, det, uniformizer, det_representative,
        gram_matrix, hermitian_genera, hermitian_local_genera, rank,
        is_inert, scales, ranks, dets, is_split, is_ramified, is_dyadic,
-       norms, primes, signatures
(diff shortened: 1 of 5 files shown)


src/QuadForm/Herm/Genus.jl
```diff
@@ -1556,14 +1556,17 @@ function representative(G::HermGenus)
   if !is_integral(G)
     s = denominator(_scale(G))
     L = representative(rescale(G, s))
-    return rescale(L, 1//s)
+    L = rescale(L, 1//s)
+    return L
```
thofma, line 1560: Why?

StevellM: In my first changes, I was caching the representative on the genus symbol so I changed like this. They I left this idea aside for later, especially because my example would not fit. I forgot to turn it back as it was.

PR #1028, opened 2023-04-04
Move type definitions in `QuadForm/` into the respective `Types.jl` files

StevellM:
This was something I wanted to do on a bigger clean-up of `QuadForm/` later in the next months, but I had trouble with a current problem known with Revise: sometimes, when modifying a file with a struct defined in it, Revise goes crazy. It does not happen for all files, neither for all structures (quite arbitrary), so I moved everything into ~~a unique file~~ the appropriate `Types.jl` files, which contain the type and structure definitions.

There is more to do, like adding docstrings and stuff, but for now, I only needed to move everything so that I can work on what I am doing without having to either change my version of Revise, or to restart Julia every time I fix a typo in a string...


src/QuadForm/Types.jl
```diff
@@ -1,3 +1,35 @@
...
+export AbstractSpace
+export AbstractSpaceMor
+export AbstractSpaceRes
+export HermGenus
+export HermLat
+export HermLocalGenus
+export HermSpace
+export JorDec
+export LatCloseEnumCtx
+export LineEnumCtx
```
simonbrandhorst, line 11:
```suggestion
```
We do not want to make these types part of the interface. They are internal.
Were they exported before?

src/QuadForm/Types.jl
```diff
@@ -1,3 +1,35 @@
...
+export LocMultGrpModSquMap
+export LocalGenusSymbol
+export LocalQuadSpaceCls
+export QuadGenus
+export QuadLat
+export QuadLocalGenus
+export QuadSpace
+export QuadSpaceCls
+export SCPComb
+export SpinorGeneraCtx
```
simonbrandhorst, line 21:
```suggestion
```

src/QuadForm/Types.jl
```diff
@@ -1,3 +1,35 @@
...
+export TorQuadModule
+export TorQuadModuleElem
+export TorQuadModuleMor
+export VecSpaceRes
+export VectorList
+export ZGenus
+export ZLat
+export ZLatAutoCtx
```
simonbrandhorst, line 29:
```suggestion
```

src/QuadForm/Types.jl
```diff
@@ -1,3 +1,35 @@
...
+export LineEnumCtx
+export LocMultGrpModSquMap
+export LocalGenusSymbol
+export LocalQuadSpaceCls
+export QuadGenus
+export QuadLat
+export QuadLocalGenus
+export QuadSpace
+export QuadSpaceCls
+export SCPComb
```
simonbrandhorst, line 20:
```suggestion
```

thofma:
I am not a big fan of changing this because of a bug in Revise. I don't understand how `add Revise@3.4.0` is more complicated than doing this (unnecessary) code churn here.

src/QuadForm/Types.jl
StevellM: They were not, but if I do not export them, will the algorithms in the other files detect the type definitions ?

thofma: Everything resides in the `module Hecke`, and there every type/function is visible to everyone. One just has to be careful about the order of inclusion.

simonbrandhorst (approved):

PR #1038, opened 2023-04-14
Fix input checks for creation of `ZLat`

StevellM:
Calling the function `ZLattice` was made by assuming that the entry matrix was actually a basis matrix, without checking it. Now it is always checked by default, but we leave the optional arguments `isbasis` and `check` for internal use.

There were few other places where the entry matrix was not always a basis so I have modified the functions accordingly (namely `lattice_in_same_ambient_space`, `is_primitive` and multiplication of a lattice by a scalar).

Then to avoid checking arguments in functions for which we know what is happening, I have disabled the input checks.

### DIFF
--- a/src/QuadForm/Quad/ZLattices.jl
+++ b/src/QuadForm/Quad/ZLattices.jl
@@ -67,9 +67,9 @@ function Zlattice(;gram, check=true)
 end
 
 @doc raw"""
-    lattice(V::QuadSpace{QQField, QQMatrix}, B::QQMatrix; isbasis=true, check=true) -> ZLat
+    lattice(V::QuadSpace{QQField, QQMatrix}, B::QQMatrix) -> ZLat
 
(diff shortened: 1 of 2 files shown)


thofma (approved):
Looks good to me. @simonbrandhorst also happy?

src/QuadForm/Quad/ZLattices.jl
```diff
@@ -67,9 +67,9 @@ function Zlattice(;gram, check=true)
 end
 
 @doc raw"""
-    lattice(V::QuadSpace{QQField, QQMatrix}, B::QQMatrix; isbasis=true, check=true) -> ZLat
+    lattice(V::QuadSpace{QQField, QQMatrix}, B::QQMatrix) -> ZLat
```
simonbrandhorst, line 70: I would keep the keyword arguments in the docu.

Issue #1018, opened 2023-03-29
Bug in `genus_representative` for non-full rank lattice


thofma:
```
B = matrix(FlintQQ, 5, 8 ,[0, 0, 0, 16, 0, 0, 0, 0, 0, 0, 0, 2, 2, 0, 0, 0, 0, 0, 0, 4, 0, 1, 0, 0, 0, 0, 0, 31//2, 3//2, 1//2, 1, 0, 0, 0, 0, 5//2, 3//2, 0, 0, 1//2]);
G = matrix(FlintQQ, 8, 8 ,[16, -8, 0, 0, 0, 0, 0, 0, -8, 16, 0, 0, 0, 0, 0, 0, 0, 0, 16, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 6]);
L = Zlattice(B, gram = G);
genus_representatives(L)
```
throws according @simonbrandhorst

thofma:
Fixed in #1047

PR #1047, opened 2023-04-18
Fix issue #1018

StevellM:
Fixes issue #1018 

There are many lines changes, mostly because I commented out around 800 lines of codes to reduce by almost a half the number of tests in `QuadForm/Quad/GenusRep.jl`.


thofma:
I don't undestand. What is with the tests? If we don't need them for coverage, we should just remove them.

StevellM:
We actually need some of them for correctness of the output, but we are testing too many examples, which makes the runtest rather slow on this file. I did not deleted the commented lines directly to see whether it was a reasonable removal (not having to remove, and the re-write later). If it looks fine, I can make a new commit with a proper deletion.

thofma:
So with these disabled, we still have the same coverage. So can we remove them? Or does coverage not imply correctness?

simonbrandhorst:
I think we added lots of test to spot any mathematical bugs. But now that we know it works, 
fewer tests seem good enough to spot regressions. And we still have the sanity check that the mass formula holds.
Therefore I am okay with removing the tests.

PR #1049, opened 2023-04-19
Fix some typos in number field docs

lgoettgens:


lgoettgens:
In both runs of the ci, some few, seemingly random, jobs fail. I cannot see any connection to the changes in this pr.

PR #1050, opened 2023-04-20
Global h2

fieker:
just to be visible - should work will probably be changed further


src/GrpAb/Dual.jl
```diff
@@ -146,6 +146,13 @@ function Base.:(==)(a::QmodnZElem, b::QmodnZElem)
   end
 end
 
+function Base.hash(a::QmodnZElem, h::UInt = UInt(121231))
```
thofma, line 149:
```suggestion
function Base.hash(a::QmodnZElem, h::UInt)
```

PR #1059, opened 2023-04-28
Global h2

fieker:


src/Misc/RatRecon.jl
```diff
@@ -64,11 +64,23 @@ end
 Tries to solve $ay=x mod b$ for $x,y < sqrt(M/2)$. If possible, returns
   (`true`, $x$, $y$) or (`false`, garbage) if not possible.
 """
-function rational_reconstruction(a::ZZRingElem, b::ZZRingElem)
-  res = QQFieldElem()
-  a = mod(a, b)
-  fl = ccall((:fmpq_reconstruct_fmpz, libflint), Int, (Ref{QQFieldElem}, Ref{ZZRingElem}, Ref{ZZRingElem}), res, a, b)
-  return fl!=0, numerator(res), denominator(res)
+function rational_reconstruction(a::ZZRingElem, b::ZZRingElem; ErrorTolerant::Bool = false)
```
thofma, line 67: Can you add an explanation of to the docstring?

fieker:
On Fri, Apr 28, 2023 at 03:14:27AM -0700, Tommy Hofmann wrote:
> 
> 
> 
> > @@ -64,11 +64,23 @@ end
>  Tries to solve $ay=x mod b$ for $x,y < sqrt(M/2)$. If possible, returns
>    (`true`, $x$, $y$) or (`false`, garbage) if not possible.
>  """
> -function rational_reconstruction(a::ZZRingElem, b::ZZRingElem)
> -  res = QQFieldElem()
> -  a = mod(a, b)
> -  fl = ccall((:fmpq_reconstruct_fmpz, libflint), Int, (Ref{QQFieldElem}, Ref{ZZRingElem}, Ref{ZZRingElem}), res, a, b)
> -  return fl!=0, numerator(res), denominator(res)
> +function rational_reconstruction(a::ZZRingElem, b::ZZRingElem; ErrorTolerant::Bool = false)
> 
> Can you add an explanation of to the docstring?
> 
Would you be happy with

ries to solve $ay=x mod b$ for $x,y < sqrt(b/2)$. If possible, returns
  (`true`, $x$, $y$) or (`false`, garbage) if not possible.

By default `y` and `b` have to be coprime for a valid solution. It is
well known that then the solution is unique.

If `ErrorTolerant` is set to `true`, then a solution is also accepted if
`x`, `y` and `b` have a common divisor `g` and if
  `a(y/g) = (x/g) mod (b/g)` is true and if the combined size is small enough,
The typical application are modular algorithms where
 - there are finitely many bad primes (ie. the `mod p` datum does
   not match the global solution modulo `p`)
 - that cannot be detected
In this case `g` will be the product of the bad primes.

thofma:
Yes, sounds good.

PR #1060, opened 2023-05-04
Fix a performance bug in abelian_extensions

thofma:
See the comment in the source. Here is a script to exhibit the original
problem:

using Hecke
Qx, x = QQ["x"]

str="""
[148,x^3-x^2-3*x+1,1,[]],
"""
ll = [eval(Meta.parse(line[1:length(line)-1])) for line in split(str)]

for a = reverse(ll)
  println("!")
  bnd=5*fmpz(10)^9
  k, _ = number_field(a[2])
  L = abelian_extensions(k, [2], bnd, signatures=[(4,1),(6,0)])
end


thofma:
cc: @fieker

fieker:
On Thu, May 04, 2023 at 06:34:32AM -0700, Tommy Hofmann wrote:
> cc: @fieker 
> 
I don't mind, but for curiosity: why not just use deepcopy?

thofma:
`deepcopy` feels not quite right. Now that I think about it, even what I am doing here feels wrong. The sane thing would be to create a new map object with the new generators at the place where the generators are created. Mutating the generators of an existing map is bad interface.

PR #1062, opened 2023-05-06
Use `VarName` for GF constructors

lgoettgens:
Resolves oscar-system/Oscar.jl#862.

This has to wait for corresponding changes in Nemo (nemocas/Nemo.jl#1444 and nemocas/Nemo.jl#1454).


lgoettgens:
After seeing Nemocas/Nemo.jl#1453, I do not know if this PR is needed. If the changed functions will disappear in the near future, this change is not sensible. Nemocas/Nemo.jl#1454 contains these changes for the copied-to location.

thofma:
Does not hurt

PR #1069, opened 2023-05-14
Add `quadratic_lattice(QQ, ...)` constructor for elements of type `ZZLat`

StevellM:


StevellM:
@simonbrandhorst does it sound good to you ? (sorry I forgot to ping you earlier)

src/QuadForm/Quad/Lattices.jl
```diff
@@ -87,12 +91,15 @@ matrix over `K` of size the length of the elements of `gens`.
...
+If $K = \mathbb{Q}$, then the output lattice is of type `ZZLat`, seen as a lattice
+over the ring $\mathbb{Z}$.
 """
-function quadratic_lattice(K::Field, gens::Vector ; gram = nothing)
+function quadratic_lattice(K::Field, gens::Vector; gram = nothing, check::Bool = true)
   if length(gens) == 0
     @assert gram !== nothing
     pm = pseudo_matrix(matrix(K, 0, nrows(gram), []))
-    L = quadratic_lattice(K, pm, gram = gram, check = false)
+    L = quadratic_lattice(K, pm, gram = gram, check = check)
```
simonbrandhorst, line 102: What is checked here? 
`quadratic_space` also has a check keyword, so the `check=check` should be passed through to the other constructors as well.

src/QuadForm/Quad/ZLattices.jl
```diff
@@ -61,11 +61,42 @@ function integer_lattice(B::ZZMatrix; gram = identity_matrix(FlintQQ, ncols(B)),
...
+    return quadratic_lattice(QQ, B, gram = gram, check = check)
+  end
+  @assert length(gens[1]) > 0
+  if gram == nothing
+    gram = identity_matrix(QQ, length(gens[1]))
+  end
+  if check
+    @assert gram isa MatElem
+    @req gram == transpose(gram) "Gram matrix must be symmetric"
+    @req all(v -> length(v) == ncols(gram), gens) "Incompatible arguments: elements in gens must all have the number of entries, equal to the numbers of gram (if specified)"
```
simonbrandhorst, line 89:
```suggestion
    @req all(v -> length(v) == ncols(gram), gens) "Incompatible arguments: elements in gens must all have the same number of entries. This number must be equal to the size of the square matrix gram (if specified)"
```

src/QuadForm/Quad/Lattices.jl
StevellM: Yes indeed now I realised that here it is not necessary, since it is checked that `pm` has full rank. But here it has rank 0 I guess we always force the check for creation of spaces, namely checking that the gram matrix is symmetric and with the correct type of entries.

PR #1081, opened 2023-05-23
Add is_modular(L, p) for ZZLat

thofma:


thofma:
@StevellM please give this a quick glance

test/QuadForm/Quad/ZLattices.jl
```diff
@@ -712,6 +712,7 @@ let
   @test Hecke.init(C) isa Hecke.ZLatAutoCtx
 end
 
+<<<<<<< HEAD
```
StevellM, line 715:
```suggestion
```

StevellM:
It looks good. Maybe could you duplicate the documentation ? The global documentation will not cover this function since the input is not an ideal but just an integer.

thofma:
There is already a docstring for `is_modular(::AbstractLat, p)`. I adjusted the wording.

PR #1085, opened 2023-05-26
Add `@vprintln`

lgoettgens:
Most uses of `@vprint` add a newline by providing `'\n'` at the end of the message string. This PR adds a dedicated macro for this.


thofma:
Yes, this is almost always what one wants anyway.

thofma:
Thanks!

PR #1099, opened 2023-05-31
Abs deg

fieker:


src/RCF/class_fields.jl
```diff
@@ -109,6 +109,15 @@ function degree(::Type{ZZRingElem}, A::ClassField)
   end
 end
 
+function degree(A::ClassField, QQ::QQField)
+  return Int(degree(ZZRingElem, A, QQ))
+end
+
+function degree(::Type{ZZRingElem}, A::ClassField, ::QQField)
+  return degree(ZZRingElem, A) * degree(base_field(A))
```
thofma, line 117:
```suggestion
  return degree(ZZRingElem, A) * absolute_degree(base_field(A))
```
If we ever have relative class fields

thofma (approved):

PR #1100, opened 2023-06-01
hashes for ZZ genera

simonbrandhorst:


simonbrandhorst:
It is not unique. But the implication `g1==g2` implies `symbol(g1)==symbol(g2)` holds for 2-adic symbols. 
That is what the documentation of `hash` says I should do.
Is it o.k.?

src/QuadForm/Quad/ZGenus.jl
```diff
@@ -896,6 +896,25 @@ function Base.:(==)(G1::ZZGenus, G2::ZZGenus)
   return true
 end
 
+Base.hash(G::ZZGenus, u::UInt) = xor(hash(G), u)
+
+
+function Base.hash(G::ZZGenus)
```
lgoettgens, line 902: One is only supposed to define `hash(::MyType, ::Uint)`. The 1-argument version is then created by julia Base.

thofma:
While thinking about this, I remembered a design problem with the (comparison) of our genera implementations:
```
julia> L = integer_lattice(;gram = identity_matrix(ZZ, 2))
Integer lattice of rank 2 and degree 2
with gram matrix
[1   0]
[0   1]

julia> G = ZGenus((2, 0), [genus(L, 2), genus(L, 3)]);

julia> G == genus(L)
false
```
This is wrong. We either strip away "trivial" symbols in the `ZGenus` constructor, or we have to adjust the `==` (and `hash`) method.

simonbrandhorst:
good point

Issue #1101, opened 2023-06-02
`saturate` fails for singular matrices


mgkurtz:
Saturating `ZZ[0 0;0 0]` should return (a copy of) the input, I think. It seems like the code for `saturate` manually computes the transform from `hnf_with_transform`. Perhaps, we could just use `hnf_with_transform` directly? I will try this in a PR.

fieker:
On Thu, Jun 01, 2023 at 11:58:21PM -0700, Markus Kurtz wrote:
> Saturating `ZZ[0 0;0 0]` should return (a copy of) the input, I think. It seems like the code for `saturate` manually computes the transform from `hnf_with_transform`. Perhaps, we could just use `hnf_with_transform` directly? I will try this in a PR.

It does not T^-1 would always be a square matrix but the result is
usually not square. (it's boring in this case)

saturate as it currently is, requires full rank input which shoudl
possibly be mentioned/ enforced (this is equivalent to hnf in the code
being invertable)

julia> saturate(ZZ[2 2 2; 0 1 0])
[1   1   1]
[0   1   0]

julia> hnf_with_transform(ZZ[2 2 2; 0 1 0])
([2 0 2; 0 1 0], [1 -2; 0 1])

thofma:
Fixed in #1102

PR #1102, opened 2023-06-02
Allow singular matrices in `saturate`

mgkurtz:
Possible fix for #1101.

### DIFF
--- a/src/Misc/Matrix.jl
+++ b/src/Misc/Matrix.jl
@@ -118,8 +118,9 @@ Computes the saturation of $A$, that is, a basis for $\mathbf{Q}\otimes M \cap
 \mathbf{Z}^n$, where $M$ is the row span of $A$ and $n$ the number of columns
 of $A$.
 
-Equivalently, return $TA$ for an invertible rational matrix $T$ such that $TA$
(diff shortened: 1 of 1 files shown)


fieker:
probably correct - but I'd expect it to be much slower (for suitable evil examples)
 - hnf_with_transform is much slower than hnf without
 - inv(T) is inverting a much larger matrix than inverting H - furthermore, H is square, upper triangular, so inverting is just writing it down. inv(T) is hard...

mgkurtz:
I had some fun, adapting the original concept to the singular case. Indeed, the `hnf_with_transform` takes longer than the whole `saturate`. Curiously, for random input, inverting the unimodular transform U was faster than inverting the hnf matrix H. Probably, because (U⁻¹)ᵀ is the pretty looking result matrix, while the hnf and its inverse may have a few large entries.

mgkurtz:
Hmph, I should listen better. You told me, not to compute the inverses, but to just solve the linear system directly. Well, this reduces the code to less than it was before, works out of the box for singular matrices, and speeds it up.

mgkurtz (commented):
Just look at commit ea0b343 – unless you are interested in all the comments in the previous commits.

fieker (approved):

Issue #1112, opened 2023-06-07
Unexpected behaviour for pseudo-hnf


StevellM:
Here is one example from our test file `QuadForm/Lattices.jl`. I am currently working on the hashes for `QuadForm` and it turned out that while modifying the equality test for `AbstractLat`, the following test fails:
```
K, a = CyclotomicRealSubfield(8, "a");
Kt, t = K["t"];
E, b = number_field(t^2 - a * t + 1, "b");
V = hermitian_space(E, gram_matrix(root_lattice(:E, 8)));
L = lattice(V);
L2 = dual(L);
@test L == L2
```
In fact, this should be true. The modification for the equality test was changing the previous `_modules_equality` between the pseudo-hnf of the pseudo-bases of both lattices, to a regular equality `==`. Though here, if one computes the lower-left pseudo-hnf for L2 (choice made for the internal `_pseudo_hnf(::AbstractLat)`, we have the following
```
julia> pseudo_matrix(L2)
Pseudo-matrix over Relative maximal order of Relative number field of degree 2 over maximal real subfield of cyclotomic field of order 8
with pseudo-basis 
(1, 1//1 * <1, 1>)
(b, 1//1 * <1, 1>)
Fractional ideal with row [4 7 10 8 6 4 2 5]
Fractional ideal with row [7 14 20 16 12 8 4 10]
Fractional ideal with row [10 20 30 24 18 12 6 15]
Fractional ideal with row [8 16 24 20 15 10 5 12]
Fractional ideal with row [6 12 18 15 12 8 4 9]
Fractional ideal with row [4 8 12 10 8 6 3 6]
Fractional ideal with row [2 4 6 5 4 3 2 3]
Fractional ideal with row [5 10 15 12 9 6 3 8]

julia> pseudo_hnf(ans, :lowerleft)
Pseudo-matrix over Relative maximal order of Relative number field of degree 2 over maximal real subfield of cyclotomic field of order 8
with pseudo-basis 
(1, 1//1 * <1, 1>)
(b, 1//1 * <1, 1>)
Fractional ideal with row [1 0 0 0 0 0 0 0]
Fractional ideal with row [0 1 0 0 0 0 0 0]
Fractional ideal with row [0 0 1 0 0 0 0 0]
Fractional ideal with row [0 0 0 1 0 0 0 0]
Fractional ideal with row [0 0 0 0 1 0 0 0]
Fractional ideal with row [0 0 0 0 0 1 0 0]
[...]

PR #1113, opened 2023-06-07
Hashes for `Quad`

StevellM:
Complements #1100 

- Change of the comparison test for `AbstractLat` which brings a broken test as mentioned in #1112 
- Add `==` for `TorQuadModule` which are defined with by lattices which all are in the same ambient (quadratic) space.


src/QuadForm/Herm/Genus.jl
```diff
@@ -1322,6 +1345,8 @@ function direct_sum(G1::HermGenus, G2::HermGenus)
   sig1 = G1.signatures
   sig2 = G2.signatures
   g3 = merge(+, sig1, sig2)
+  bd = union(support(2*maximal_order(base_field(E))), support(discriminant(maximal_order(E))))
+  filter!(g -> (prime(g) in bd) || (scales(g) != Int[0]), LGS)
```
thofma, line 1349: Is the `scales(g) != Int[0]` for the block not being unimodular?

thofma:
I think this all looks good, but we need some comments to remember what we did in a few weeks (a few days?).

Somewhere we need to write down which primes we throw away etc.

src/QuadForm/Herm/Genus.jl
```diff
@@ -1018,6 +1020,21 @@ function ==(G1::HermLocalGenus, G2::HermLocalGenus)
...
+  u = Base.hash(base_field(g), u)
+  u = Base.hash(prime(g), u)
+  h = xor(hash(det(g)), reduce(xor, (hash(s[1:2]) for s in g.data)))
+  if is_ramified(g)
+    if !is_dyadic(g)
+      h = xor(h, reduce(xor, (hash(s[3]) for s in g.data if iseven(s[1]))))
+    else
+      h = xor(h, reduce(xor, (hash(g.data[i][1] == 2*g.norm_val[i]) for i = 1:length(g.data))))
+      h = xor(h, hash(g.ni))
+    end
```
simonbrandhorst, line 1033: Please explain.

src/QuadForm/Quad/Genus.jl
```diff
@@ -758,10 +763,22 @@ end
...
 
+function Base.hash(g::QuadLocalGenus, u::UInt)
+  u = Base.hash(base_field(g), u)
+  u = Base.hash(prime(g), u)
+  h = xor(hash(scales(g)), hash(ranks(g)))
+  if is_dyadic(g)
+    h = xor(h, hash(weights(g)))
+    h = xor(h, hash(witt_invariant(g)))
+    h = xor(h, hash(norms(g)))
+  end
```
simonbrandhorst, line 778: Are the norms really invariants?

simonbrandhorst (commented):
Please enforce the sparsity of the local genus symbols in the constructor.
Either by a test or more conveniently by filtering the symbol in the constructor method.

src/QuadForm/Herm/Genus.jl
StevellM: Exactly, we keep the local symbol if the prime of bad, or if it is not unimodular. A local symbol is unimodular if and only if it consists of only one block of scale zero.

StevellM:
I will add a bunch of comments regarding what we discard and the choices made for the hashes. I will also re-enable the broken test since the issue has be solved.

simonbrandhorst (approved):

PR #1114, opened 2023-06-07
Correctly reduce rows in `pseudo_hnf_kb`

### LINKED ISSUE #1112: Unexpected behaviour for pseudo-hnf

joschmitt:
Fixes #1112 

The only meaningful change is the added row reduction in line 1389. I also added some comments (apparently I wasn't a big fan 6 years ago) and restructured the code a little bit.
Sorry for the wrong implementation.

I did not touch `is_pseudo_hnf`. It does not test whether the entries above the diagonal are reduced, but there is no documentation whether it should and it is an internal function any way.

### DIFF
--- a/src/NumFieldOrd/NfOrd/LinearAlgebra.jl
+++ b/src/NumFieldOrd/NfOrd/LinearAlgebra.jl
@@ -3,6 +3,7 @@ import Base.vcat, Base.hcat
 
 add_verbosity_scope(:PseudoHnf)
 add_assertion_scope(:PseudoHnf)
+add_verbosity_scope(:PseudoHnfKB)
 
 function _det_bound(M::Generic.Mat{NfOrdElem})
   n = nrows(M)
@@ -522,7 +523,7 @@ matrix part of $P$ will be upper triangular with some technical normalisation
 for the off-diagonal elements. This operation preserves the module.
 The used transformation is returned as a second return value.
 
-A optional second argument can be specified as a symbols, indicating the desired
+A optional second argument can be specified as a symbol, indicating the desired
 shape of the echelon form. Possible are
 `:upperright` (the default) and `:lowerleft`
 """
@@ -1190,6 +1191,7 @@ function kb_search_first_pivot(H::PMat, start_element::Int = 1)
    return 0, 0
 end
 
+# Reduces the row r with pivot in column c (so r = pivot[c]) with other known pivots
 function kb_reduce_row!(H::PMat{T, S}, U::Generic.Mat{T}, pivot::Vector{Int}, c::Int, with_transform::Bool) where {T <: NumFieldElem, S}
    r = pivot[c]
    A = H.matrix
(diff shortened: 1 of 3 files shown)


thofma:
Thanks!

PR #1115, opened 2023-06-09
Increase precision to isolate roots in `n_positive_roots`

mgkurtz:
A `schur_index` computation failed for me, due to `n_positive_roots` failing if two roots are rather close by. This proposal simply doubles the precision in that case. See the added test for an example.

### DIFF
--- a/src/Misc/Poly.jl
+++ b/src/Misc/Poly.jl
@@ -856,7 +856,7 @@ function _n_positive_roots_sqf(f::PolyElem{nf_elem}, P::NumFieldEmb; start_prec:
     try
       rts = roots(g)
     catch e
-      startswith(e.msg, "unable to isolate all roots") || rethrow()
+      e isa ErrorException && startswith(e.msg, "unable to isolate all roots") || rethrow()
       prec *= 2
       continue
     end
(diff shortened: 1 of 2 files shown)


src/Misc/Poly.jl
```diff
@@ -853,7 +853,14 @@ function _n_positive_roots_sqf(f::PolyElem{nf_elem}, P::NumFieldEmb; start_prec:
       coeffs[i + 1] = evaluate(coeff(f, i), P, prec)
     end
     g = Cx(coeffs)
-    rts = real.(Hecke.roots(g))
+    try
+      rts = Hecke.roots(g)
```
thofma, line 857:
```suggestion
      rts = roots(g)
```

src/Misc/Poly.jl
```diff
@@ -853,7 +853,14 @@ function _n_positive_roots_sqf(f::PolyElem{nf_elem}, P::NumFieldEmb; start_prec:
...
+    catch e
+      startswith(e.msg, "unable to isolate all roots") || rethrow()
```
thofma, line 859: I think you also need to check the type of `e`, right?

thofma:
Looks good! I left two comments.

PR #1118, opened 2023-06-11
Revert "set the name more properly to help Janko (#1106)"

thofma:
This reverts commit 782291bec6dda68de84fb7b3d5de269e39505dff.


thofma:
This was breaking for Oscar, so let's wait with this change till we do the next breaking release.

fieker:
It needs a doc fix in Oscar after...

thofma:
I need a quick release, so I revert this for now (but will reapply it).

PR #1119, opened 2023-06-13
Use `ZZRingElem` instead of `Int` to prevent overflow in `maximal_order`

mgkurtz:
I know, I should not use utterly random input. Either way, I did so and had a `schur_index` of 2 were it should be 1. This was due to `Int`s silently overflowing and giving a wrong `maximal_order`. So, here is the easy fix for that.


thofma:
Nice catch. Which function did produce the overflow?

mgkurtz:
Well, as overflows do not throw errors, I did not notice. But once happily even the `Int` conversion failed, which led me on the right track. Looking at the code, the operations with `p` are squares (can overflow), `_maximal_ideals`  (only used as `quo(O, p*O, p)`), converting it to a `ZZRingElem` 😄, and `pradical`. For the latter, we could do a trick like
https://github.com/thofma/Hecke.jl/blob/aa3c436a3c38c4791b03c613e827ca76251f4bd2/src/NumFieldOrd/NfOrd/Ideal/Ideal.jl#L2104-L2109
if feasible.

src/AlgAssAbsOrd/Order.jl
```diff
@@ -849,11 +849,11 @@ function MaximalOrder(O::AlgAssAbsOrd{S, T}) where { S <: AlgGrp, T <: AlgGrpEle
       if mod(d, p^2) != 0
```
fieker, line 849: can overflow

src/AlgAssAbsOrd/Ideal.jl
```diff
@@ -1166,7 +1166,7 @@ end
 Given an order $O$ and a prime $p$, it returns the radical of the ideal
 generated by $p$.
 """
-function pradical(O::AlgAssAbsOrd, p::Int)
+function pradical(O::AlgAssAbsOrd, p::Union{Int, ZZRingElem})
```
lgoettgens, line 1169:
```suggestion
function pradical(O::AlgAssAbsOrd, p::IntegerUnion)
```
Why not use `IntegerUnion`?

src/AlgAssAbsOrd/Order.jl
```diff
@@ -694,7 +694,7 @@ function pmaximal_overorder_meataxe(O::AlgAssAbsOrd{S, T}, p::Union{ZZRingElem,
   return O
 end
 
-function pmaximal_overorder_tr(O::AlgAssAbsOrd, p::Int)
+function pmaximal_overorder_tr(O::AlgAssAbsOrd, p::Union{ZZRingElem, Int})
```
lgoettgens, line 697:
```suggestion
function pmaximal_overorder_tr(O::AlgAssAbsOrd, p::IntegerUnion)
```
Why not use `IntegerUnion`?

src/AlgAssAbsOrd/Order.jl
mgkurtz: This `p` is a factor of `degree(O) = dim(algebra(O))`. Can this really become larger than $2^{31}$? The `p` in the `p^2` above that is a factor of `discriminant(O::AlgAssAbsOrd) :: ZZRingElem`.

thofma:
I am still puzzled. None of the fixes here fixes an overflow, but just make some functions not cast to `Int`. Where did the overflow occur?

mgkurtz:
The overflow occurs if `pmaximal_overorder(O, p::Int)` is called with large `p`. Such as `pmaximal_overorder(O, Int(p))`, where `p` is a factor of `discriminant(O) :: ZZRingElem`. By using `pmaximal_overorder(O, p)` instead, we compute in `ZZ`, and the overflow disappears.

This PR does not aim to prevent any overflows in `pmaximal_overorder(O, ::Int)`.

thofma:
Ah, I misunderstood the meaning of overflow. Thanks for the explanation. But how did you get the wrong result instead of a conversion error? `Int(p)` will throw if `p` is too large and not return a wrong result.

src/AlgAssAbsOrd/Order.jl
fingolfin: If it really can't overflow, it'd still be useful to add a comment stating (and justifying) that

mgkurtz:
Sorry, here a case distinction:
* `p` is small: `pmaximal_order(O, Int(p))` works as intended
* `p` fits into an `Int` but some calculations with `p`, say `p^2`, do not: at some point inside `pmaximal_order(O, Int(p))`, say when computing `p^2`, an overflow happens yielding wrong results
* `p` does not even fit into an `Int`: `pmaximal_order(O, Int(p))` throws an error on converting `Int(p)`.

thofma:
OK thanks. Can you add the example that went wrong as a test? Or was it too large?

src/AlgAssAbsOrd/Order.jl
mgkurtz: Added an assertion

mgkurtz:
Yes, it was somewhat large. Now, I just added a small example, both for the $p>2^{63/2}$ and the $p>2^{63}$ case. As suspected, it worked for the prime $3037000493<2^{63/2}$, and not anymore for the next prime $3037000507>2^{63/2}$.

thofma:
Thanks

PR #1123, opened 2023-06-15
Cleanup `indefiniteLLL`

StevellM:
There was a conflict of conventions between PARI left actions and our right actions, so at that time the code was written with a bunch of unnecessary `transpose`. This PR is about fixing this problem, as well as turning the files related to `indefiniteLLL` to a format similar to other files in the repo.

### DIFF
--- a/src/QuadForm/indefiniteLLL.jl
+++ b/src/QuadForm/indefiniteLLL.jl
@@ -348,7 +348,7 @@ Gram matrix of `L` with respect to `U`.
 julia> G = ZZ[1 0 0; 0 4 3; 0 3 2];
 
 julia> lll_gram_indef_ternary_hyperbolic(G)
-([0 0 -1; 0 1 0; -1 0 0], [0 1 -2; 1 0 0; 0 1 -1])
+([0 0 -1; 0 1 0; -1 0 0], [-1 -1 0; 0 0 -1; -2 -1 0])
 ```
 """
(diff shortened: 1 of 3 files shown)


src/QuadForm/indefiniteLLL.jl
```diff
@@ -70,73 +68,79 @@ function _complete_to_basis(v::MatElem{ZZRingElem}, redflag::Bool = false)
...
+      H[1,:] = sol
 
-      return transpose(H)*G*H, transpose(H), transpose(sol)
+      return H*G*transpose(H), H, sol
     end
   end
 
   #Case 2: G has a block +- [1 0 ; 0 -1] on the diagonal
   for i = 2:n
     if G[i-1,i] == 0 && G[i-1,i-1]*G[i,i] == -1
```
simonbrandhorst, line 133:
```suggestion
    if G[i-1,i] == 0 && abs(G[i-1,i-1])==1 &&abs(G[i,i])==1 && sign(G[i-1,i-1])*sign(G[i,i]) == -1
```

simonbrandhorst (approved):
Thanks for cleaning this up. It looks good.
I did not check all the transposes, but if the tests pass, this is fine by me.

src/QuadForm/indefiniteLLL.jl
StevellM: Indeed, good catch, thanks!

PR #1124, opened 2023-06-15
Change the order of arguments in `MapFromFunc`

lgoettgens:
Resolves https://github.com/oscar-system/Oscar.jl/issues/1104.

Unfortunately, one cannot add deprecations, as the constructor has no type annotations at all (cf. https://github.com/thofma/Hecke.jl/blob/master/src/Map/MapType.jl#L225). Thus, this PR is breaking.

The changes needed in Oscar are proposed in https://github.com/oscar-system/Oscar.jl/pull/2475.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Hecke"
 uuid = "3e1990a7-5d81-5526-99ce-9ba3ff248f21"
-version = "0.18.16"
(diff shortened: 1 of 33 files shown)


lgoettgens:
Failures seem to be unrelated to the changes. I have seen the same errors in other PRs as well.

PR #1127, opened 2023-06-17
Ignore complex embeddings for Schur index

mgkurtz:
As described by Nebe and Steel on the third page of https://doi.org/10.1016/j.jalgebra.2009.04.026, the complex embeddings do not contribute to the Schur index, as $ℂ⊗A≅ℂ^{s×s}$ is always a full matrix algebra over ℂ. Only the infinite completions with real embeddings contribute, as there $ℝ⊗A$ can be isomorphic to $ℍ^{s/2×s/2}$ giving a Schur index of 2, or isomorphic to $ℝ^{s×s}$ giving trivial Schur index 1.

See the tests for an example of the full matrix algebra over $K=ℚ[i]$ but with the basis matrices multiplied with the some unit $α$. This should not change the Schur index, but with the current implementation this multiplication changes the Schur index from 1 to either 1, 2, or an error stemming from getting the “positive” roots of a polynomial like $y^4-1$.

### DIFF
--- a/src/AlgAss/Ramification.jl
+++ b/src/AlgAss/Ramification.jl
@@ -136,15 +136,14 @@ function schur_index(A::AbsAlgAss{nf_elem}, P::InfPlc)
   @req iscentral(A) "Algebra must be central"
   @req issimple(A) "Algebra must be simple"
 
-  if dim(A) % 4 != 0
-    return 1
-  end
+  dim(A) % 4 == 0 && is_real(P) || return 1
 
-  x = trace_signature(A, P)
+  x, = trace_signature(A, P)
   n = root(dim(A), 2)
-  if x[1] == divexact(n * (n + 1),2)
+  if x == n * (n+1) / 2
     return 1
   else
+    @assert x == n * (n-1) / 2
     return 2
   end
 end
--- a/test/AlgAss/Ramification.jl
+++ b/test/AlgAss/Ramification.jl
@@ -118,4 +118,21 @@
 
   A = matrix_algebra(QQ, 2)
   @test schur_index(A) == 1
+
+  @testset "ignore complex embeddings" begin
+    # The schur index of QQ[i]^{2×2} is 1
+    # Test this for various ways of writing QQ[i]^{2×2}
+
+    X = zero_matrix(QQ, 4, 4)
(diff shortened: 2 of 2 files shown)


src/AlgAss/Ramification.jl
```diff
@@ -136,15 +136,14 @@ function schur_index(A::AbsAlgAss{nf_elem}, P::InfPlc)
...
-  if dim(A) % 4 != 0
-    return 1
-  end
+  dim(A) % 4 == 0 && is_real(P) || return 1
 
-  x = trace_signature(A, P)
+  x, = trace_signature(A, P)
   n = root(dim(A), 2)
-  if x[1] == divexact(n * (n + 1),2)
+  if x == n * (n+1) / 2
```
thofma, line 143: We introduce floating point numbers?

src/AlgAss/Ramification.jl
```diff
@@ -136,15 +136,14 @@ function schur_index(A::AbsAlgAss{nf_elem}, P::InfPlc)
   @req iscentral(A) "Algebra must be central"
   @req issimple(A) "Algebra must be simple"
 
-  if dim(A) % 4 != 0
-    return 1
-  end
+  dim(A) % 4 == 0 && is_real(P) || return 1
```
thofma, line 139: Hm, this check does not look right. We have `dim(A) % 4 != 0 || !is_real(P)` implies Schur index equal one. The new check would fail for a quaternion algebra?

mgkurtz: It is just a matter of perspective to say either $¬a ∨ ¬b ⇒ \text{return }1$ or $a ∧ b \ ∨ \ \text{return }1$. I prefer the latter, as it means that afterwards $a∧b$ holds, which, in my eyes, helps understanding the code that follows.

mgkurtz: By the way, `schur_index(Hecke.quaternion_algebra2(QQ, -1, -1)) == 2` works fine with my code. This should also be covered by the existing tests for the `schur_index` which run successfully.

thofma: Ah sorry, you are of course right. I parsed this wrong with `&&` at the end (that is, if the first part holds, then do this).

PR #1128, opened 2023-06-19
Fix `is_isotropic`

StevellM:
Typo in the code of `is_isotropic` for quadratic spaces. I have fixed and put an example from Cassels for which the previous implementation could return a wrong output (I have checked it by hand, following Cassels propositions)


simonbrandhorst (approved):

src/QuadForm/Quad/Spaces.jl
```diff
@@ -818,6 +818,7 @@ is_isotropic(V::QuadSpace{QQField,QQMatrix}, p::Int) = is_isotropic(V, ZZ(p))
 is_isotropic(V::QuadSpace{QQField,QQMatrix}, p::PosInf) = _isisotropic(diagonal(V), p)
 
 function is_isotropic(V::QuadSpace, p)
+  @req is_prime(p) "p must be prime"
```
simonbrandhorst, line 821: Does this work in all cases?  integers, prime ideals, places, infinite places?

PR #1129, opened 2023-06-20
Fix renaming of local genera for integer lattices

StevellM:
Some weeks ago, I have continued the big renaming within `QuadForm/` and I renamed `ZpGenus` to `LocalZZGenus`. But the latter was actually inconsistent with `HermLocalGenus` and `QuadLocalGenus` used for the other lattices ( and inconsistent with the new naming conventions). I have therefore changed it to the appropriate form `ZZLocalGenus`. 

I have checked and there are no occurrences of this type in the master of Oscar so it should not break it.


thofma:
Makes sense to me. Any objection @simonbrandhorst?

PR #1131, opened 2023-06-22
Use package extensions and remove Requires.jl

thofma:
Fixes https://github.com/oscar-system/Oscar.jl/issues/2400


fingolfin:
@thofma what's the status of this? Anything we can do to help?

thofma:
It is scheduled for the next breaking release, which hopefully happens tomorrow after the work with Lars.

PR #1132, opened 2023-06-22
Speed up `multiplication_table(::AlgMat)`

mgkurtz:
This mostly improves the case of matrix algebras with small dimension `d=dim(A)` compared to the dimension `D=degree(A)^2*dim_of_coefficient_ring(A)` of the full matrix algebra.

This improves `coefficients(B[i]*B[j])` by exploiting, that we know `B[i]*B[j]` to be representable as linear combination of the $d$ entries of `B`. The old code in `_check_matrix_in_algebra` does a $D×D$ basis transform of our element `B[i]*B[j]`, checking, that it is then really spanned by the first $d$ unit vectors. The new `_matrix_in_algebra`  just transforms the $d$ pivot columns without a check.

This gave slight speedup in the full matrix algebra case (if one disables the late #1130 special handling) and tenfold and more, when $D÷d$ gets larger.

Probably, the `_check_matrix_in_algebra` code could also benefit from some of the changes I incorporated in the new variant.

### DIFF
--- a/src/AlgAss/AlgMat.jl
+++ b/src/AlgAss/AlgMat.jl
@@ -32,7 +32,7 @@ matrix_algebra_type(K::Field) = matrix_algebra_type(typeof(K))
 
 matrix_algebra_type(::Type{T}) where {T <: Field} = AlgMat{elem_type(T), dense_matrix_type(elem_type(T))}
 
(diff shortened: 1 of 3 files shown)


mgkurtz:
Fixed it.

PR #1133, opened 2023-06-22
Disable several `AlgAss` checks by default

mgkurtz:
More speedup when creating new `AlgAss` with computed multiplication tables, that mathematically have to be correct and where checking often needs a lot more time than everything else.

As I just fixed a bug in [47a74b6](https://github.com/thofma/Hecke.jl/pull/1132/commits/47a74b635b3a5b4dfc0d00df1ffe48b9d236bb77), where I did my maths wrong, optionally reenabling those and some previously disabled tests seems sensible to me.

### DIFF
--- a/src/AlgAss/AlgAss.jl
+++ b/src/AlgAss/AlgAss.jl
@@ -169,7 +169,7 @@ function AlgAss(R::Ring, mult_table::Array{T, 3}; check::Bool = true) where {T}
   return A
 end
 
-function AlgAss(R::Ring, d::Int, arr::Vector{T}) where {T}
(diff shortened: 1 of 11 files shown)


thofma:
Could we condense to one assertion flag? We probably don't need such a fine control for the assertions.

mgkurtz:
> Could we condense to one assertion flag? We probably don't need such a fine control for the assertions.

Okay, I put them on different levels.

Additionally, now all calls to `AlgAss` with multiplication tables got `check = get_assertion_level(:AlgAss)>0`. As an alternative, I made this the default for `check` in #1134.

mgkurtz:
Closed in favor of #1134.

PR #1134, opened 2023-06-23
Use `associative_algebra` as frontend to `AlgAss` and disable checks when calling `AlgAss` directly

mgkurtz:
Concept for an alternative to #1133.


thofma:
I like it. We should merge this one here.

mgkurtz:
Rethinking it, perhaps `AlgAss(f::PolyElem)` should have another name like `quotient_algebra(f)`?

thofma:
Let's keep it for now.

PR #1135, opened 2023-06-23
Refactoring parts of `MSet.jl`

lgoettgens:
Some small changes to `misc/MSet.jl`:

- prettifying some docstrings, fixing typos
- `IteratorSize`, `IteratorEltype` and `eltype` now only implement the one dispatch taking a type. The other dispatches (for objects) are defined in julia base.
- `push!`, `pop!`, `delete!` and `multiplicity` now use the same ad-hoc coercion for the input
- export `multiplicity`
- add `filter`

### DIFF
--- a/src/Misc/MSet.jl
+++ b/src/Misc/MSet.jl
@@ -1,30 +1,30 @@
-export MSet, multiplicities, subsets
+export MSet, multiplicities, multiplicity, subsets
 
 @doc raw"""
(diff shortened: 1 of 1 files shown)


fieker (approved):

thofma:
Because of import/export nonsense, this is breaking Oscar. But we will do a breaking release soon anyway.

PR #1138, opened 2023-06-26
Miscellaneous matrix methods

mgkurtz:
The call to `hnf(A)` gives a row echelon form, but no information about rank or pivots. These fast methods provide that information. (Without verifying the echelon form.)

Also fix the `transpose!` method, I have lately introduced.


thofma:
Good idea to not export them, but then the doctests need the module prefix.

thofma:
Thanks!

I am not too big of the `cond || bla` syntax. It is a bit hard to parse at a first glance (for errors it is fine, since it is usually clear what is checked). I will merge this as is, but maybe we can avoid it in the future.

PR #1141, opened 2023-06-27
Replace `== nothing` by `=== nothing`

lgoettgens:


thofma:
What is the reason for this change? The binary version is much more readable and with `===` replaced by `==` faster than `isnothing` (across all julia version >= 1.6).

fieker:
On Tue, Jun 27, 2023 at 08:47:58AM -0700, Lars Göttgens wrote:
> 
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/thofma/Hecke.jl/pull/1141
> 
> -- Commit Summary --
> 
>   * Replace `== nothing` by `isnothing`
> 
Why?
Is this better readable? To me: no
Is this faster? Don't know

lgoettgens:
Comparing against `nothing` with `===`/`!==` instead of `==`/`!=` is a recommended change by the Julia linter in VSCode.

Sorry for the `isnothing` stuff in the first version of this. Using `=== nothing` and `isnothing` seems to be a personal style of code, and it wasn't my intention to debate that here.

thofma:
Thanks for the adjustment. (https://docs.julialang.org/en/v1.6/manual/performance-tips/#Checking-for-equality-with-a-singleton is the reference for the `=== nothing` vs `isnothing` difference on some older julia versions, which we still support).

PR #1146, opened 2023-06-30
convenience functions for the oscar book

simonbrandhorst:


src/QuadForm/Spaces.jl
```diff
@@ -273,6 +273,8 @@ inner_product(V::AbstractSpace, v::MatElem, w::MatElem)
 
 _inner_product(L::AbstractLat, v, w) = inner_product(ambient_space(L), v, w)
 
+inner_product(L::AbstractLat, v, w) = inner_product(ambient_space(L), v, w)
```
thofma, line 276: This is the opposite of what is added for `ZZLat`, or am I missing something?

simonbrandhorst: We have to figure out if $v$ and $w$ are given with respect to the ambient space or are vectors with respect to the basis of $L$.  I would argue that 
`inner_product(L::ZLat, u::ZZMat, v::ZZMat)`
`inner_product(L::ZLat, u::Vector{ZZRingElem}, v::Vector{ZZRingElem})`
indicate that $u$ and $v$ are given w.r.t the basis of $L$. 

`inner_product(L::ZLat, u::QQMat, v::QQMat)`
`inner_product(L::ZLat, u::Vector{QQRingElem}, v::Vector{QQRingElem})`
indicate that $u$ and $v$ are given w.r.t the ambient space of $L$.

thofma: Hm, I am against it. It don't think that `inner_product(L, [ZZ(1)], [ZZ(1)])` and `inner_product(L, [QQ(1)], [QQ(1)])` should do something different. I would argue that this is not a nice interface for the user and very confusing. At the same time it goes against the (implicit) convention, that promotion should not change the result. It is then also ambiguous what `inner_product(L, [1], [1])` should be.

simonbrandhorst: Then I would say that 
`inner_product(V::AbsSpace, u, v)` means take the gram matrix of the space $V$ 
and 
`inner_product(L::ZZLat u, v)` means take the gram matrix of the lattice.... but this does not make sense over number fields ... damn it.

thofma: I think remember now why I did not add initially. We could allow `inner_product(L, v, w)` only for Z-lattice and integral v, w. This would be better than nothing.

fingolfin: I concur with @thofma, this is a no-go.

How about making the form an explicit object, and use it explicitly? And/or introducing a notion of "space with form" ?

src/Misc/Matrix.jl
```diff
@@ -882,6 +882,17 @@ end
...
+  if any(base_ring(i)!==R for i in x)
+    x = [change_base_ring(R,i) for i in x]
+  end
+  if length(x)>0
+    return diagonal_matrix(x)
+  else
+    return zero_matrix(R,0,0)
+  end
+end
+
```
lgoettgens, line 895: This should be added in AbstractAlgebra or Nemo (for the time until the remains of https://github.com/Nemocas/Nemo.jl/pull/1501 have been sorted out there)

src/QuadForm/Spaces.jl
simonbrandhorst: We do have a space with form.  I decided for the minimally invasive solution that
`inner_product(L::AbstractLat, u v) = inner_product(ambient_space(L), u, v)`

PR #1148, opened 2023-07-03
Fix overflow bug in C2xC2 field computation

thofma:


thofma:
Bug appears with bound ~10^12, so a bit hard to add a test.

PR #1094, opened 2023-05-29
Change argument order in `factor` and `roots`

lgoettgens:
Resolves https://github.com/thofma/Hecke.jl/issues/641. Similar changes have been added as PRs to AA (https://github.com/Nemocas/AbstractAlgebra.jl/pull/1374), Oscar (https://github.com/oscar-system/Oscar.jl/pull/2433), and Nemo (https://github.com/Nemocas/Nemo.jl/pull/1486).

PR #1153, opened 2023-07-06
Adjust to new AA and Nemo

thofma:


lgoettgens:
This contains https://github.com/thofma/Hecke.jl/pull/1143, https://github.com/thofma/Hecke.jl/pull/1088, and https://github.com/thofma/Hecke.jl/pull/1094.

PR #1165, opened 2023-07-12
Few more fixes in `QuadForm`

StevellM:
Some of the `reduce` were buggy when one has to reduce an empty collection. Change some comparisons into an `init`.


thofma:
Looks good, thanks! Do you want to add some of these offending genera as tests?

StevellM:
Yes you are right, I will add some tests!

StevellM:
While writing the tests, I saw also some functions not working for rank 0 lattices, and few lowercase printing missing.

PR #1189, opened 2023-08-15
Minor fixes for type stability

thofma:


src/QuadForm/Quad/GenusRep.jl
```diff
@@ -2396,7 +2402,9 @@ function automorphism_group_generators(g::QuadBin{ZZRingElem})
     end
   end
 
-  @assert all(T -> _action(gorig, T) == gorig, gens)
+  for T in gens
+    @assert action(gorig, T) == gorig
```
StevellM, line 2406:
```suggestion
    @assert _action(gorig, T) == gorig
```

StevellM:
Is there a specific reason to change all the `@assert all(-bla-)` into loops of assertions ?

thofma:
Type stability (https://github.com/JuliaLang/julia/issues/15276). There might be a different way to do which is less intrusive. I will explore it a bit more.

StevellM (approved):

PR #1190, opened 2023-08-23
Add Simon's constructor for elliptic curves.

HechtiDerLachs:
CC: @simonbrandhorst

### DIFF
--- a/test/EllCrv/EllCrv.jl
+++ b/test/EllCrv/EllCrv.jl
@@ -295,11 +295,11 @@ end
 @testset "additional constructors" begin
   R, (x, y) = polynomial_ring(QQ, [:x, :y])
   f = y^2 - x^3 - 3*x^2 + 7*x - 4
-  Hecke.elliptic_curve(f, x, y)
+  elliptic_curve(f, x, y)
 
   pt, t = polynomial_ring(QQ, :t)
   kt = fraction_field(pt)
   R, (x, y) = polynomial_ring(kt, [:x, :y])
   f = y^2 - x^3 - 3*t*x^2 + 7*x - 4*t^2 - 3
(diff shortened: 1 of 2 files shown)


test/EllCrv/EllCrv.jl
```diff
@@ -291,3 +291,15 @@
     @test_throws ErrorException P1//5
   end
 end
+
+@testset "additional constructors" begin
+  R, (x, y) = polynomial_ring(QQ, [:x, :y])
+  f = y^2 - x^3 - 3*x^2 + 7*x - 4
+  Hecke.elliptic_curve(f, x, y)
```
thofma, line 298:
```suggestion
elliptic_curve(f, x, y)
```

test/EllCrv/EllCrv.jl
```diff
@@ -291,3 +291,15 @@
...
+
+  pt, t = polynomial_ring(QQ, :t)
+  kt = fraction_field(pt)
+  R, (x, y) = polynomial_ring(kt, [:x, :y])
+  f = y^2 - x^3 - 3*t*x^2 + 7*x - 4*t^2 - 3
+  Hecke.elliptic_curve(f, x, y)
```
thofma, line 304:
```suggestion
  elliptic_curve(f, x, y)
```

thofma:
A docstring would be nice

HechtiDerLachs:
> A docstring would be nice

Really? I was afraid, this would not be appreciated because of too many documented methods of the same function. But I can try to add it. 

Btw: It seems you already had similar functionality. Is that true?

thofma:
At the moment there are two methods documented, so adding another won't hurt. And I don't think we have anything that accepts bivariate polynomials.

PR #1199, opened 2023-09-04
Fix alignment in printing of maps for restriction of scalars

StevellM:
This might break a doctest in Oscar (for `hermitian_structure(::ZZLatWithIsom)`)


thofma:
I am not sure we need this? Do we ever align things in this way? This will also look funny if the there is not enough space
```
Map of change of scalars
  from quadratic
  space of dimension 4
  to   hermitian space
  of dimension 1
```

StevellM:
Hum for the printing of the Scheme part in Oscar I did this spacing stuff. It is not necessary but I feel like it is a bit cleaner. We could close it and maybe change it another day; in the meantime I will remove the particular printing in Oscar so we avoid any break.

StevellM:
I understand. We can close the PR then.

PR #1205, opened 2023-09-11
Remove wildly general `check_parent`

### LINKED ISSUE #1145: Wildly general `check_parent`
There is some wildly general `check_parent` in https://github.com/thofma/Hecke.jl/blob/7b06fea4fab911e1849e6011147cc77216545fe1/src/NumField/NfRel/NfRel.jl#L446-L448

Where is it used? Can we add some type annotations to it to restrict its domain? I am afraid that, in the current generality, it shadows some real MethodErrors.

lgoettgens:
This removes the code mentioned in https://github.com/thofma/Hecke.jl/issues/1145 to see where it is currently used aka what fails without it.

Resolves #1145.

### DIFF
--- a/src/GenOrd/Ideal.jl
+++ b/src/GenOrd/Ideal.jl
@@ -251,8 +251,7 @@ end
 
 
 function Base.:(+)(a::GenOrdIdl, b::GenOrdIdl)
-  check_parent(a, b)
-  O = order(a)
+  @req order(a) === order(b) "Ideals must have same order"
 
   if iszero(a)
     return b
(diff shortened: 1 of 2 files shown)


thofma:
I will have a look at the tests

lgoettgens:
I am working on fixes for the failures

lgoettgens:
This looks surprisingly good. Please add some kind of `BREAKING` label.

lgoettgens:
The two failures are more or less expected. But otherwise, it looks good.

Issue #1139, opened 2023-06-26
Quaternion algebra is commutative?


lgoettgens:
```julia
julia> A = Hecke.AlgQuat(QQ, QQ(-1), QQ(-1))
Quaternion algebra over
Rational field
defined by i^2 = -1, j^2 = -1

julia> A isa Ring
true

julia> _,i,j,k = basis(A);

julia> i*j == j*i
false
```
I think the inheritance from `Ring` is a mistake here, as that should only be the case for commutative rings, which is not the case here.

thofma:
Once upon a time, matrices could only be constructed over `<: Ring`. I think this was changed some time ago, so we could fix it now.

PR #1206, opened 2023-09-11
Let `AbsAlgAss` no longer be a `Ring`

lgoettgens:
Fixes https://github.com/thofma/Hecke.jl/issues/1139.

Let's see if anything breaks. If not, I see no reason not to change this.
However, I would classify this a breaking change.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -28,7 +28,7 @@ GAPExt = "GAP"
 PolymakeExt = "Polymake"
 
 [compat]
-AbstractAlgebra = "^0.31.0"
+AbstractAlgebra = "^0.32.0"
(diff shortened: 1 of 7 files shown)


thofma:
Yes, correct

lgoettgens:
This needs a few more changes in AbstractAlgebra (https://github.com/Nemocas/AbstractAlgebra.jl/pull/1423). The relevant test files succeed locally. I am currently running the full testsuite locally.

lgoettgens:
I think this needs to be restarted once https://github.com/thofma/Hecke.jl/pull/1208 is merged.

PR #1208, opened 2023-09-13
Bump version and AbstractAlgebra and Nemo compats

lgoettgens:
cc @thofma


lgoettgens:
Let's see what broke in Hecke

lgoettgens:
Most errors looks like https://github.com/Nemocas/AbstractAlgebra.jl/pull/1424.
I can make a doctest fix run locally for that. Unless you are already doing that :)

thofma:
Should be fixed now

PR #1209, opened 2023-09-14
Towards non-nice relative number fields

thofma:


thofma:
@fieker This only introduces the `any_order` for relative extensions. It does not fix the maximal order yet.

PR #1212, opened 2023-09-16
Fix trace equivalence for infinite isometries

### LINKED ISSUE #1210: `hermitian_structure_with_transfer_data` fails for isometry with infinite order
Maybe this is not meant to be supported; but there *is* some code in there which explicitly checks for infinity, but then at some point it runs into code that assumes the order to be finite...

Example to reproduce:
```
julia> L = integer_lattice(identity_matrix(ZZ,2); gram=ZZ[1 2; 2 1]);

julia> is_definite(L)
false

julia> gens = automorphism_group_generators(L);

julia> f = gens[1]
[4   -1]
[1    0]

julia> multiplicative_order(f)
infinity

julia> hermitian_structure_with_transfer_data(L, f)
ERROR: MethodError: no method matching euler_phi(::PosInf)
```

There error is in `src/QuadForm/Lattices.jl:1393`.

StevellM:
Should fix #1210 .

### DIFF
--- a/src/QuadForm/Lattices.jl
+++ b/src/QuadForm/Lattices.jl
@@ -1135,6 +1135,7 @@ function trace_lattice_with_isometry_and_transfer_data(H::AbstractLat{T}; alpha:
     return H, f, AbstractSpaceRes(V, V, identity_matrix(E, n), identity_matrix(E, n))
   end
 
+  @req H isa HermLat "H must be hermitian or defined over the integers"
   @req maximal_order(E) == equation_order(E) "Equation order and maximal order must coincide"
 
   # This function perform the trace construction on the level of the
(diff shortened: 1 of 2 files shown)


src/QuadForm/Lattices.jl
```diff
@@ -1135,7 +1135,7 @@ function trace_lattice_with_isometry_and_transfer_data(H::AbstractLat{T}; alpha:
     return H, f, AbstractSpaceRes(V, V, identity_matrix(E, n), identity_matrix(E, n))
   end
 
-  @req (degree(E) == 2) && (is_totally_complex(E)) && (is_totally_real(base_field(E))) "The base field of H must be CM"
```
simonbrandhorst, line 1138: I think this still holds, even in the infinite order case.

StevellM: In the example of the issue it does not hold, `E` is real. It does not seem to be actually required to have `E/K` CM.

simonbrandhorst: You are right of course.

simonbrandhorst (approved):

PR #1213, opened 2023-09-17
pretty printing for HeckeMap

simonbrandhorst:


simonbrandhorst:
Will break doctests in Oscar.

thofma:
I find it a bit confusing, to do a `println(f)` in the doctest, just so that some other printing is tested.

simonbrandhorst:
So you suggest calling `show(X)` instead? Will do next time.

PR #1214, opened 2023-09-18
Add some tweaks for type stability etc.

fingolfin:
Several changes deal with JET warnings about "captured variables"
resulting in disabled optimizations resp. type unstable code.
E.g. `mFF` was assigned different values in two places which then
are referenced inside closures.

Likewise, `NN` was used in lambdas, which we now avoid.


fingolfin:
Indeed, it breaks experimental code:
```
  MethodError: no method matching elem_type(::Type{Oscar.GaloisCohomology_Mod.RelativeBrauerGroup})
```
Fix submitted as https://github.com/oscar-system/Oscar.jl/pull/2814

PR #1225, opened 2023-09-26
Add sparse block diagonal matrices

lgoettgens:
Needed for https://github.com/oscar-system/Oscar.jl/pull/2844. I started with the method in AbstractAlgebra and tried to adapt it as best as possible.


lgoettgens:
The CI errors are going to be fixed by https://github.com/Nemocas/AbstractAlgebra.jl/pull/1448 and are independent of the changes here.

PR #1227, opened 2023-09-27
Fix `kronecker_product` for sparse matrices with zero row

lgoettgens:
The following example fails:
```julia
A = sparse_matrix(FlintZZ, [2 0; 0 0])
kronecker_product(A, A)
```
I would expect it to instead return the sparse matrix `[4 0 0 0; 0 0 0 0; 0 0 0 0; 0 0 0 0]`.


fingolfin (approved):
LGTM

PR #1239, opened 2023-10-09
Fix `Order(K, [ ... ])`

### LINKED ISSUE #1223: Bugs in Order(...)
Here are some bugs in the current `Order(...)` function, a discussion of the algorithm and some half-baked ideas. I could try to implement a fix, but wanted to first ask for opinions about what the algorithm should actually be...

### Issues

There are several (more or less separate) bugs in the `Order(...)` [function](https://www.thofma.com/Hecke.jl/dev/orders/orders/#Order-Tuple{AnticNumberField,%20Vector{nf_elem}}) that in a given number field K should compute the order generated by a given list of elements.

```
using Oscar
R,x=QQ["x"]
```

**Issue 1:** Running ```K,w=NumberField(x,"w"); Order(K,[1])``` should produce the order Z in the number field Q, but gives:
```
ERROR: data does not define an order: dimension to small
Stacktrace:
 [1] _order(K::AnticNumberField, elt::Vector{nf_elem}; cached::Bool, check::Bool, extends::Nothing)
   @ Hecke ~/.julia/packages/Hecke/D2kay/src/NumFieldOrd/NfOrd/NfOrd.jl:1097
 [2] _order
   @ ~/.julia/packages/Hecke/D2kay/src/NumFieldOrd/NfOrd/NfOrd.jl:995 [inlined]
 [3] Order(::AnticNumberField, a::Vector{nf_elem}; check::Bool, isbasis::Bool, cached::Bool)
   @ Hecke ~/.julia/packages/Hecke/D2kay/src/NumFieldOrd/NfOrd/NfOrd.jl:766
 [4] Order
   @ ~/.julia/packages/Hecke/D2kay/src/NumFieldOrd/NfOrd/NfOrd.jl:750 [inlined]
 [5] Order(K::AnticNumberField, a::Vector{Int64}; check::Bool, isbasis::Bool, cached::Bool)
   @ Hecke ~/.julia/packages/Hecke/D2kay/src/NumFieldOrd/NfOrd/NfOrd.jl:778
 [6] Order(K::AnticNumberField, a::Vector{Int64})
   @ Hecke ~/.julia/packages/Hecke/D2kay/src/NumFieldOrd/NfOrd/NfOrd.jl:770
 [7] top-level scope
   @ REPL[7]:1
```
The problem is that although `bas` is initialized to the basis `1` of Q, the matrix `B` still has zero rows:
https://github.com/thofma/Hecke.jl/blob/d23dfff2cd230bc9b34325fda4ba13179bf4c8f4/src/NumFieldOrd/NfOrd/NfOrd.jl#L1027-L1028

**Issue 2:** Running ```K,w=NumberField(x,"w"); Order(K,[])``` should also produce the order Z in the number field Q, but gives:
```
ERROR: StackOverflowError:
Stacktrace:
 [1] Order(K::AnticNumberField, a::Vector{Any}; check::Bool, isbasis::Bool, cached::Bool)
   @ Hecke ~/.julia/packages/Hecke/D2kay/src/NumFieldOrd/NfOrd/NfOrd.jl:773
 [2] Order(K::AnticNumberField, a::Vector{Any}; check::Bool, isbasis::Bool, cached::Bool) (repeats 16360 times)
   @ Hecke ~/.julia/packages/Hecke/D2kay/src/NumFieldOrd/NfOrd/NfOrd.jl:778
 [3] Order(K::AnticNumberField, a::Vector{Any})
   @ Hecke ~/.julia/packages/Hecke/D2kay/src/NumFieldOrd/NfOrd/NfOrd.jl:770
 [4] top-level scope
   @ REPL[8]:1
```
I think this is a type confusion. Somehow, `map(K, [])` returns an object of type `Vector{Any}` instead of `Vector{nf_elem}`. (Why?) That's why this function
https://github.com/thofma/Hecke.jl/blob/d23dfff2cd230bc9b34325fda4ba13179bf4c8f4/src/NumFieldOrd/NfOrd/NfOrd.jl#L770-L779
keeps calling itself instead of calling
https://github.com/thofma/Hecke.jl/blob/d23dfff2cd230bc9b34325fda4ba13179bf4c8f4/src/NumFieldOrd/NfOrd/NfOrd.jl#L750-L751

**Issue 3:** Running `K,w=NumberField(x^4-10*x^2+1,"w"); a=1//2*w^3-9//2*w; b=1//2*w^3-11//2*w; Order(K,[a,b])` should compute the field $K=\mathbb Q(\sqrt2,\sqrt3)$, the elements $a=\sqrt2$, $b=\sqrt3$, and return the order $\mathbb Z[a,b]$ of $K$. Instead, it fails:
```
ERROR: data does not define an order: dimension to small
Stacktrace:
 [1] _order(K::AnticNumberField, elt::Vector{nf_elem}; cached::Bool, check::Bool, extends::Nothing)
   @ Hecke ~/.julia/packages/Hecke/D2kay/src/NumFieldOrd/NfOrd/NfOrd.jl:1097
 [2] _order
   @ ~/.julia/packages/Hecke/D2kay/src/NumFieldOrd/NfOrd/NfOrd.jl:995 [inlined]
 [3] Order(::AnticNumberField, a::Vector{nf_elem}; check::Bool, isbasis::Bool, cached::Bool)
   @ Hecke ~/.julia/packages/Hecke/D2kay/src/NumFieldOrd/NfOrd/NfOrd.jl:766
 [4] Order(::AnticNumberField, a::Vector{nf_elem})
   @ Hecke ~/.julia/packages/Hecke/D2kay/src/NumFieldOrd/NfOrd/NfOrd.jl:750
 [5] top-level scope
   @ REPL[34]:1
```
(Both `Order(K,[a,b,a*b])` and `Order(K,[a,b],check=false)` actually succeed and produce the correct output.)
In this function, when check=true, the function `_order(...)` actually never reaches "phase 2". It never computes the product a*b, only produces the Z-module generated by 1,a,b, which is indeed not an order. I'm confused about the purpose of the whole phase management. (See below.)

**Issue 4:** This coincidentally doesn't produce any serious problems, but I think there's a mistyped variable name. Compare these lines:
https://github.com/thofma/Hecke.jl/blob/d23dfff2cd230bc9b34325fda4ba13179bf4c8f4/src/NumFieldOrd/NfOrd/NfOrd.jl#L1009
https://github.com/thofma/Hecke.jl/blob/d23dfff2cd230bc9b34325fda4ba13179bf4c8f4/src/NumFieldOrd/NfOrd/NfOrd.jl#L1016
This branch, which would use the modular Hermite normal form optimization, is never taken:
https://github.com/thofma/Hecke.jl/blob/d23dfff2cd230bc9b34325fda4ba13179bf4c8f4/src/NumFieldOrd/NfOrd/NfOrd.jl#L1069
Regardless of this typo, I don't understand why this optimization should be used only when we're extending an order. (See below.)

There are also some unnecessary checks in the code, as was already pointed out by @fieker in inline comments:
https://github.com/thofma/Hecke.jl/blob/d23dfff2cd230bc9b34325fda4ba13179bf4c8f4/src/NumFieldOrd/NfOrd/NfOrd.jl#L1052
https://github.com/thofma/Hecke.jl/blob/d23dfff2cd230bc9b34325fda4ba13179bf4c8f4/src/NumFieldOrd/NfOrd/NfOrd.jl#L1091-L1092

### Fixed algorithm

I don't know if the current algorithm is documented anywhere, so here's an outline in pseudocode for what I think the function was trying to do to compute the order generated by the elements in a given vector `gens`, where the number field K has degree n:
```python
A and B will be Z-submodules of K.
A := 1*Z (or whatever order we're extending)
for g in gens:
    B := A
    for i in 1,2,...:
        if i == n and not check:
            break
        if g^i in B:
            break
        if i == n:
            raise error(g is not an algebraic integer)
        B := B + g^i * A
        # At this point, B = A + g*A + ... + g^i*A
    A := B
    # At this point, A is the subring of K generated by the generators up to g.
return A
```

The Z-submodules of K are represented by Hermite-reduced bases. One can use a modular HNF algorithm (the optimization mentioned above) as soon as B has full rank. (Using "the determinant of B" as the modulus.)

If you append k rows to an $n\times n$ Hermite reduced matrix and apply Hermite reduction to the result, I think the algorithm might do about $n^3 + kn^2$ operations in Z. (The summand $n^3$ arises because all diagonal entries might decrease, and then all the off-diagonal entries have to be reduced.)

Hence, it is faster by a factor of ~n to append n rows and Hermite reduce once than to n times append one row and reduce. The attempt to implement such an optimization is the root of all evil in the current implementation. :)

One easy way to solve this issue (if one is willing to sacrifice a constant factor in the running time) should be to represent a Z-module by a list of generators which is not necessarily Hermite reduced. Only apply Hermite reduction after line 12 if we have accumulated $\geq2n$ generators, and once at the end of the algorithm. It's okay to just skip the checks in lines 8-11 if the current list of generators of B hasn't been reduced, yet.

### More efficient methods?

I feel like there must be more efficient algorithms for this problem, at least when the number r of generators is large. Consider the total number N of rows in all calls to the Hermine normalization function `hnf`.

If you run the above algorithm (or the currently implemented one) with r >= 2 generators, I suppose N could be on the order of $rn^2$. (After the first generator, A can have rank n, so we then have to append ~n rows, for i=1,...,n, for each generator.)

Here's an algorithm with $N=\mathcal O(r+n^2\log(rn))$:
```python
A := Z-module generated by 1 and the given generators
for i in 1,2,...:
    B := A*A
    if B == A:
        break
    if 2^(i-1) >= r*n:
        raise error(not all algebraic integers)
    A := B
    # At this point, A is the Z-module generated by all products g_1 ... g_k of k <= 2^i of the given generators.
return A
```
(Computing `A*A` involves a matrix with $\leq n^2$ rows.)

joschmitt:
Closes #1223 .
I basically reimplemented the functionality now and all the errors in #1223 should be solved. In the examples collected from the tests by @thofma , this is always faster or at least comparable to the runtime of the old function.
I am not sure I take the best modulus for the modular HNF; maybe some guru could have a look.


fagu:
I don't have counterexamples, but this feels wrong.

1. You only apply Hermite reduction when you have at least n vectors: https://github.com/thofma/Hecke.jl/blob/401a7116869fc881c4ea339899aa271cc3da7a6a/src/NumFieldOrd/NfOrd/NfOrd.jl#L1052C1-L1052C1
I didn't check, but from the name of the function, I would guess that `is_zero_mod_hnf!(a,b)` assumes that b is Hermite reduced. This is called by `in_span_of_B` even before you have n vectors.

3. The `start` logic doesn't make sense to me. It for example means that if at the beginning of step i (in which add multiples of e^i), you already had n vectors, you will not do anything in step i+1 because `start=n+1` and `length(bas)=n`.

fagu:
Sorry, never mind about the start thing. I missed a later line.

fagu:
Ah, there's a difference between what's in B and what's in `bas`. That probably fixes my first concern. (Sorry about the noise...)

fieker:
I have not tried to check the math, but it looks fine to me. Do you happen to recall the examples that sparked the `extends` stuff? My guess would have been a large degree normal field where `extends` was applied with the equation order and either some subfield generator or an automorphism...
That would result in some interesting test cases to benchmark (I don't think this is worse than the old stuff)

fagu:
The algorithm looks correct to me now. Thanks for the fixes and clarifications!

Lines 1016-1018 and 1084-1086 seem unnecessary, but maybe I'm wrong.

Small comments / questions on the modular HNF:
- Any invertible matrix in Hermite normal form is triangular, so computing the determinant is easy. Does the `det` function recognize this?
- Whenever you save the determinant of the numerator (`m=...`), you could also save the denominator, and then later use modulus (current denominator)*(earlier determinant)/(earlier denominator), I think.
The point is that (determinant)/(denominator) * Z^n is contained in the Z-module spanned by the rows of B.
- Maybe also update the determinant of the numerator and the denominator after line 1072 if (new determinant)/(new denominator) < (old determinant)/(old denominator)?

src/NumFieldOrd/NfOrd/NfOrd.jl
```diff
@@ -999,114 +992,105 @@ The equation order of the number field.
...
   end
-
-  if !isdefined(B, :num)
-    error("data does not define an order: dimension to small")
-  end
-
-  # Make an explicit check
-  @hassert :NfOrd 1 defines_order(K, B)[1]
-  return Order(K, B, cached = cached, check = check)
+  return Order(K, B, cached = cached, check = check)::order_type(K)
```
fagu, line 1093: Doesn't `check = false` or `check = (check && extends !== nothing)` suffice, depending on whether or not we're supposed to check that `extends` is actually an order?

You've already checked that the generators are integral.

The `Order` constructor with `check = true` reduces all products of pairs of basis elements modulo the (Hermite reduced) basis. For example if there is only one generator, this seems like an unnecessary amount of work.

joschmitt:
> @joschmitt Can you add a test case for the line with missing coverage?

I added the test.

joschmitt:
> I have not tried to check the math, but it looks fine to me. Do you happen to recall the examples that sparked the `extends` stuff? My guess would have been a large degree normal field where `extends` was applied with the equation order and either some subfield generator or an automorphism... That would result in some interesting test cases to benchmark (I don't think this is worse than the old stuff)

A real life example would be good. In the tiny examples I had from the tests also the modular approach never really made sense.

joschmitt:
> Lines 1016-1018 and 1084-1086 seem unnecessary, but maybe I'm wrong.

Lines 1016-1018 are exactly to catch the case that somebody submits an empty array like in #1223 Issue 2.
Lines 1084-1086 are 'unnessary', but don't hurt in my opinion. They certainly helped with debugging.

> Small comments / questions on the modular HNF:
> 
>     * Any invertible matrix in Hermite normal form is triangular, so computing the determinant is easy. Does the `det` function recognize this?

I do not know.
>     * Whenever you save the determinant of the numerator (`m=...`), you could also save the denominator, and then later use modulus (current denominator)*(earlier determinant)/(earlier denominator), I think.
>       The point is that (determinant)/(denominator) * Z^n is contained in the Z-module spanned by the rows of B.

This modular business is always confusing to me when 'fractional types' are involved. My understanding is the following: I have the `FakeFmpqMat` `B` given by an integer matrix `num(B)` and a (integer) denominator `den(B)`. Now I know another `FakeFmpqMat` `M` such that the $\mathbb Z$-module spanned by `M` is contained in the one spanned by `B` and I want to get the HNF of `B` by computing modulo `M`. The function `hnf_modular_eldiv` takes a multiple of the largest elementary divisor of the matrix as modulus. It is not documented for `FakeFmpqMat`, but from looking at the implementation, I assume that I need to choose this modulus with respect to `num(B)`. So I have the inclusion `den(B)*num(M)/den(M) \subseteq num(B)` and it seems to me that the best possible guess for the largest elementary divisor of `num(B)` is `den(B)*(largest elementary divisor of num(M))` from this. And my best guess for the largest elementary divisor of `num(M)` is unfortunately `det(num(M))`.
>     * Maybe also update the determinant of the numerator and the denominator after line 1072 if (new determinant)/(new denominator) < (old determinant)/(old denominator)?

It is not clear to me whether the computation of the determinants (in practise) is worth the effort of finding a (slightly) better modulus. Here a real world example might help.

src/NumFieldOrd/NfOrd/NfOrd.jl
joschmitt: This depends on the general philosophy, I think. My understanding so far is that we by default rather check correctness one time too often, but let the user turn off the checks, if they know what they are doing.

joschmitt:
I added a specialized determinant for triangular matrices and tuned the modular stuff a bit as discussed with @thofma .

For me, this is good to go (assuming CI agrees). If anybody needs it to be faster, I would need a real-life example.

Issue #1238, opened 2023-10-09
Missing documentation for `MSet`


StevellM:
We have multisets in Hecke, under the type `MSet`, but it is unknown to the users. I would suggest to:
- add snake case constructors, like `multiset`
- revise the printing
- make it appear in the online manual with the main functions

PR #1241, opened 2023-10-10
Complement on multi-sets and sub-set iterators

StevellM:
Related to #1238 

This is built on what was already implemented by @fieker : 
- Fix some non-working functions
- Add more documentation and doctests
- Create an html documentation
- Test the different functions of the file
- Add some constructors and functionalities

Not all possible functionalities are available, there are still many existing functions for `Set` and `Dict` which could be implemented for `MSet`; this would have to be done on-demand I guess.

I have changed the printing, to be closer to the printing of `Set` and `Dict`. The code could have been done nicer if julia offered the possibility to use their printing methods for `Dict` where one can choose the separator they want... Instead I wrote something which works similarly (if we see that we will get beyond the size of the screen, we cut and put some dots).

Nothing is perfect or fixed, those are my suggestions of improvements together with a template for the html documentation so that Hecke/Oscar users are aware of the existence of the functionalities for multi-sets.

### DIFF
--- a/src/Misc/MSet.jl
+++ b/src/Misc/MSet.jl
@@ -211,7 +211,7 @@ function Base.show(io::IO, ::MIME"text/plain", s::MSet)
       end
     else
       lmax = maximum(length(sprint(show, a)) for a in un[1:szh])
-      for i in 1:sz
+      for i in 1:szh
         println(io)
         k = un[i]
         lk = length(sprint(show, k))
--- a/src/Misc/MSet.jl
+++ b/src/Misc/MSet.jl
@@ -1,14 +1,24 @@
-export MSet, multiplicities, multiplicity, subsets
+export MSet
+export multiplicities
+export multiplicity
+export multiset
+export subsets
 
 @doc raw"""
+    MSet{T} <: AbstractSet{T}
+
 Type for a multi-set, i.e. a set where elements are not unique, they
 (can) have a multiplicity. `MSet`s can be created from any finite iterator.
 
-# Example
+# Examples
 ```jldoctest
 julia> MSet([1,1,2,3,4,4,5])
-MSet(5, 4 : 2, 2, 3, 1 : 2)
-
+MSet{Int64} with 7 elements:
+  5
+  4 : 2
+  2
+  3
+  1 : 2
 ```
 `4 : 2` means the element `4` has multiplicity `2`, i.e. was included twice.
 """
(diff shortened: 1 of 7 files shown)


thofma:
I have a small question: Could you add the "sum" while you are at it? (see https://en.wikipedia.org/wiki/Multiset#Basic_properties_and_operations).

I will talk to some people about the printing (whether we want to keep `:` or whether we really want to drop the multiplicity if it is one).

StevellM:
Now I noticed that I had to change the `sum(::MSet)` which was falling back into `sun(::AbstractSet)`. Should I add then a deprecation since now we have a proper method `sum(::MSet, ::MSet)` ? Or maybe define `sum(::MSet)` as it was working before ? The latter would imply two different behaviours depending on the number arguments though.

src/Misc/MSet.jl
```diff
@@ -44,70 +191,372 @@ function Base.show(io::IO, ::MIME"text/plain", s::MSet)
...
+
+### Iteration
+
 Base.isempty(s::MSet) = isempty(s.dict)
 Base.length(s::MSet) = sum(values(s.dict))
 Base.IteratorSize(::Type{MSet}) = Base.HasLength()
 Base.IteratorEltype(::Type{MSet}) = Base.HasEltype()
 Base.eltype(::Type{MSet{T}}) where {T} = T
-Base.in(x, s::MSet) = haskey(s.dict, x)
+Base.in(x, s::MSet) = any(y -> x == y, keys(s.dict))
```
fieker, line 263: Isn't this much slower than the old version? O(n) Vs O(1)?

src/Misc/MSet.jl
```diff
@@ -119,48 +568,92 @@ function Base.filter(pred, s::MSet)
   return t
 end
 
+function Base.filter!(pred, s::MSet)
+  un = unique(s)
+  for x in un
```
fieker, line 573: This copies everything

src/Misc/MSet.jl
```diff
@@ -44,70 +191,372 @@ function Base.show(io::IO, ::MIME"text/plain", s::MSet)
...
+
+function Base.intersect(s1::MSet, s2::MSet)
+  un1 = unique(s1)
+  un2 = unique(s2)
+  val = filter(x -> any(y -> x == y, un2), un1)
+  T = promote_type(eltype(s1), typeof.(val)...)
+  s = similar(s1, T)
+  d = s.dict
+  for x in val
+    y = un2[findfirst(y -> x == y, un2)]
```
fieker, line 528: Use isequal(x)?

src/Misc/MSet.jl
```diff
@@ -1,41 +1,188 @@
...
+function multiset(l::Vector{T}, m::Vector{Int}) where {T}
+  @req length(m) == length(l) "Lists must have the same length"
+  @req minimum(m) > 0 "Multiplicity list must consist of positive integers"
+  return MSet{T}(l, m)
+end
+
+@doc raw"""
+    multiset(T::Type) -> MSet{T}
+
+Create an unitialized multi-set `M` with elements of type `T`.
```
fieker, line 125: It's empty, not uninitialised

src/Misc/MSet.jl
```diff
@@ -44,70 +191,372 @@ function Base.show(io::IO, ::MIME"text/plain", s::MSet)
...
+  un1 = unique(s1)
+  un2 = unique(s2)
+  val = filter(x -> any(y -> x == y, un2), un1)
+  @req promote_type(T, typeof.(val)...) == T "Cannot coerce elements"
+  d = s1.dict
+  for x in un1
+    if !(x in val)
+      delete!(s1, x)
+    else
+      y = un2[findfirst(y -> x == y, un2)]
```
fieker, line 549: Isequal

src/Misc/MSet.jl
```diff
@@ -119,48 +568,92 @@ function Base.filter(pred, s::MSet)
...
+
+julia> multiplicity(m, 6)
+0
+```
 """
-function multiplicity(s::MSet{T}, x::T) where {T}
-  y = x isa T ? x : T(x)
-  if haskey(s.dict, y)
-    return s.dict[y]
+function multiplicity(s::MSet, x)
```
fieker, line 640: Change from constant time to linear? Why?

src/NumField/NfAbs/Elem.jl
```diff
@@ -462,7 +462,7 @@ function _ds(fa)
   @assert all(x->x == 1, values(fa.fac))
   T = Int[degree(x) for x = keys(fa.fac)]
   M = MSet(T)
-  return Set(sum(s) for s = subsets(M) if length(s) > 0)
+  return Set(sum(collect(s)) for s = subsets(M) if length(s) > 0)
```
fieker, line 465: Why collect?

fieker (changes requested):

fieker:
I find the name terrible. Sum(mset) I'd expect a sum over the elements with
multiplicity. Sum of 2 msets I've never seen, or expected

src/Misc/MSet.jl
StevellM: I haven't tried but the previous version is not practical. See for instance what could happen in a very basic example: 
```
julia> d = Dict{QQFieldElem, Int}(2 => 2)
Dict{QQFieldElem, Int64} with 1 entry:
  2 => 2

julia> haskey(d, 2)
false
```
I have not thought about about a better work-around..

src/Misc/MSet.jl
StevellM: Same as explain for the `in` function: if I have a `MSet{QQFieldElem}`, then for testing the multiplicity of 2, one needs to check for `QQ(2)`. If we want to force this and then do not leave room for this kind of flexibility, I am ok with reverting this.

StevellM:
The naming `sum` for `MSet` is taken from what I have seen on Wikipedia. I had a look to other CAS: python apparently uses "combine" and maple uses the "+" operator. The thing with `sum(::MSet{T})` is that it would only work whenever `T <: RingElem` or at least the types for which `sum` is defined. We could change `sum` as I defined it with something of the form `multiset_sum`, `disjoint_union` or `combine` and keep maybe the `+` operator for simplicity ? In that case the `sum(::MSet)` will use directly `sum(::AbstractSet)` as it was doing before (provided that the user know what they are doing)

fieker:
On Tue, Oct 10, 2023 at 11:29:17PM -0700, Stevell Muller wrote:
> @StevellM commented on this pull request.
> 
> 
> 
> >  Base.isempty(s::MSet) = isempty(s.dict)
>  Base.length(s::MSet) = sum(values(s.dict))
>  Base.IteratorSize(::Type{MSet}) = Base.HasLength()
>  Base.IteratorEltype(::Type{MSet}) = Base.HasEltype()
>  Base.eltype(::Type{MSet{T}}) where {T} = T
> -Base.in(x, s::MSet) = haskey(s.dict, x)
> +Base.in(x, s::MSet) = any(y -> x == y, keys(s.dict))
> 
> I haven't tried but the previous version is not practical. See for instance what could happen in a very basic example: 
> ```
> julia> d = Dict{QQFieldElem, Int}(2 => 2)
> Dict{QQFieldElem, Int64} with 1 entry:
>   2 => 2
> 
> julia> haskey(d, 2)
> false
> ```
> I have not thought about about a better work-around..
> 
That was a deliberate decision to not support that. I'm (un)happy to
discuss this in detail:
 In Julia base, effort is made so that 
   hash(1) == hash(1//1) == hash(1.0) == ...
Despite them being different types. This is possible, for BigInt and
BigFloat only via some memory expensive operations, the float is
decomposed into a rational and then mutilated.
This screws up the performance in many important algorithms.
Thus we (I) decided to not support that madness
  hash functions should never allocate
The price is that for ZZRingElem(2)^100 the hash is different to
Int128(2)^100...
For in: it would be better, in my opinion to coerce into the type if we
want to support that. But the entire point of Dicts is to have an O(1)
lookup and not an O(length(D))
in(a::T, M::MSet{T}) = haskey(M.dict, a)
in(a::Any, M::MSet{T}) = haskey(M.dict, T(a))
it will still be "bad" if you put e.g. number field elements in - esp. 
in different fields....

fieker:
On Tue, Oct 10, 2023 at 11:43:44PM -0700, Stevell Muller wrote:
> The naming `sum` for `MSet` is taken from what I have seen on Wikipedia. I had a look to other CAS: python apparently uses "combine" and maple uses the "+" operator. The thing with `sum(::MSet{T})` is that it would only work whenever `T <: RingElem` or at least the types for which `sum` is defined. We could change `sum` as I defined it with something of the form `multiset_sum`, `disjoint_union` or `combine` and keep maybe the `+` operator for simplicity ? In that case the `sum(::MSet)` will use directly `sum(::AbstractSet)` as it was doing before (provided that the user know what they are doing)
> 
In the end I don't mind. I've never tried to add MSets and don't see
this happening in the near future....
sum and prod for Sets I haven't tried recently. I think it was not
allowed for a long time....

StevellM:
Thanks for the explanation! I will revert what I have changed to keep it in linear time then. I have modified also some other parts to avoid to create unnecessary lists.

For the sum, I will just keep the `+` for now and if one day one comes up with an acceptable name, it will be easy to add.

StevellM:
@fieker Do the last changes I have been seem good for you ?

fieker (approved):

thofma:
Merci Stevell

PR #1252, opened 2023-10-18
Extend `_order(K, gens)` for non-commutative algebras

joschmitt:


src/NumFieldOrd/NfOrd/NfOrd.jl
```diff
@@ -1046,18 +1048,50 @@ function _order(K::S, elt::Vector{T}; cached::Bool = true, check::Bool = true, e
...
     # start := k + 1. In the next iteration, we then have
     # new_bas := [e^2, e^2*a_2, ..., e^2*a_k] (assuming that there was no
     # reduction of the basis in between).
-    for i in 1:df
+
+    powers_of_e = 1 # count the iterations, so the power of e by which we multiply
+                    # (only relevant in the commutative case)
+    while true
+      # In the commutative case, this behaves like a "for _ in 1:df"-loop
+      if is_commutative(K) && powers_of_e == df + 1
```
thofma, line 1064: Maybe move the check to the start of the for loop? I know it should be cached, but still.

joschmitt: I'm not sure I understand. Do you mean putting the result of `is_commutative(K)` in a local variable?

thofma:
Merci

PR #1254, opened 2023-10-20
Attempt to improve some codes in `QuadForm`

StevellM:
Mainly trying to avoid creation of lists when it is not necessary and use more the the `!` functions whenever possible. 

I have tried to improve some type stability by giving new names to variables which were changing types during the sub-processes and inferring more type for lists. Hopefully this with help for compilation time (?).

I have added some semi-colons and spaces in some places where I noticed then missing, for code readability: this was not my first goal so the code is not perfectly reformatted.

### DIFF
--- a/src/QuadForm/Herm/Genus.jl
+++ b/src/QuadForm/Herm/Genus.jl
@@ -631,7 +631,7 @@ function _genus(::Type{HermLat}, E::S, p::T, data::Vector{Tuple{Int, Int, Int}},
   z.is_dyadic = is_dyadic
   z.is_ramified = is_ramified
   z.is_split = is_split
-  keep = [i for (i, s) in enumerate(data) if s[2] != 0]  # We keep only blocks of non-zero rank
+  keep = Int[i for (i, s) in enumerate(data) if s[2] != 0]  # We keep only blocks of non-zero rank
   z.data = data[keep]
   return z
 end
@@ -791,7 +791,7 @@ function _genus(::Type{HermLat}, E::S, p::T, data::Vector{Tuple{Int, Int, Int}},
   z.is_split = false
   # We test the cheap thing
   @req z.is_dyadic && z.is_ramified "Prime must be dyadic and ramified"
-  keep = [i for (i, s) in enumerate(data) if s[2] != 0]    # We only keep the blocks of non-zero rank
+  keep = Int[i for (i, s) in enumerate(data) if s[2] != 0]    # We only keep the blocks of non-zero rank
   z.norm_val = norms[keep]
   z.data = data[keep]
   z.ni = _get_ni_from_genus(z)
--- a/src/QuadForm/Herm/Genus.jl
+++ b/src/QuadForm/Herm/Genus.jl
@@ -129,7 +129,7 @@ $\mathfrak p$ of $\mathcal O_K$, return the rank of any hermitian lattice whose
 $\mathfrak p$-adic completion has local genus symbol `g`.
 """
 function rank(G::HermLocalGenus)
-  return reduce(+, (rank(G, i) for i in 1:length(G)), init = Int(0))
+  return reduce(+, rank(G, i) for i in 1:length(G); init = Int(0))
 end
 
 @doc raw"""
@@ -161,7 +161,7 @@ The returned value is $1$ or $-1$ depending on whether the determinant is a loca
 norm in `K`.
 """
 function det(G::HermLocalGenus)
-  return reduce(*, (det(G, i) for i in 1:length(G)), init = Int(1))
+  return reduce(*, det(G, i) for i in 1:length(G); init = Int(1))
 end
 
 @doc raw"""
@@ -631,7 +631,7 @@ function _genus(::Type{HermLat}, E::S, p::T, data::Vector{Tuple{Int, Int, Int}},
   z.is_dyadic = is_dyadic
   z.is_ramified = is_ramified
   z.is_split = is_split
-  keep = [i for (i,s) in enumerate(data) if s[2] != 0]  # We keep only blocks of non-zero rank
+  keep = [i for (i, s) in enumerate(data) if s[2] != 0]  # We keep only blocks of non-zero rank
   z.data = data[keep]
   return z
 end
@@ -791,7 +791,7 @@ function _genus(::Type{HermLat}, E::S, p::T, data::Vector{Tuple{Int, Int, Int}},
   z.is_split = false
   # We test the cheap thing
   @req z.is_dyadic && z.is_ramified "Prime must be dyadic and ramified"
-  keep = [i for (i,s) in enumerate(data) if s[2] != 0]    # We only keep the blocks of non-zero rank
+  keep = [i for (i, s) in enumerate(data) if s[2] != 0]    # We only keep the blocks of non-zero rank
   z.norm_val = norms[keep]
   z.data = data[keep]
   z.ni = _get_ni_from_genus(z)
@@ -980,7 +980,7 @@ function ==(G1::HermLocalGenus, G2::HermLocalGenus)
     # in particular normal.
 
     # Thus we only have to check Theorem 3.3.6 4.
-    return all(i -> det(G1, i) == det(G2, i), [i for i in 1:t if iseven(scale(G1,i))])
+    return all(det(G1, i) == det(G2, i) for i in 1:t if iseven(scale(G1, i)))
   end
 
   # Dyadic ramified case
@@ -1354,7 +1354,7 @@ function direct_sum(G1::HermGenus, G2::HermGenus)
   prim = eltype(primes(G1))[]
   P1 = Set(primes(G1))
   P2 = Set(primes(G2))
-  for p in union(P1, P2)
+  for p in union!(P1, P2)
     g1 = G1[p]
     g2 = G2[p]
     g3 = direct_sum(g1, g2)
@@ -1367,7 +1367,7 @@ function direct_sum(G1::HermGenus, G2::HermGenus)
   # For genera of hermitian lattices, are bad all primes dividing 2 and those
   # dividing the discriminant of the field extension (discriminant of the
   # maximal order of the top field in E/K).
-  bd = union(support(2*maximal_order(base_field(E))), support(discriminant(maximal_order(E))))
+  bd = union!(support(2*maximal_order(base_field(E))), support(discriminant(maximal_order(E))))
   # We keep only the local symbols which are defined over bad primes or which
   # are not unimodular.
   # Unimodular <=> There is only one Jordan block, and it is a scale valuation 0
@@ -1423,7 +1423,7 @@ function genus(L::Vector{<:HermLocalGenus}, signatures::Dict{<:InfPlc, Int})
   @req all(g -> rank(g) == r, L) "Local genus symbols must have the same rank"
   E = base_field(first(L))
   @req all(g -> base_field(g) == E, L) "Local genus symbols must be defined over the same extension E/K"
-  bd = union(support(2*maximal_order(base_field(E))), support(discriminant(maximal_order(E))))
+  bd = union!(support(2*maximal_order(base_field(E))), support(discriminant(maximal_order(E))))
   filter!(g -> (prime(g) in bd) || (scales(g) != Int[0]), L)
   return HermGenus(E, r, L, signatures)
 end
@@ -1448,7 +1448,7 @@ Return the global genus symbol `G` of the hermitian lattice `L`. `G` satisfies:
 end
 
 function _genus(L::HermLat)
-  bad = bad_primes(L, discriminant = true, dyadic = true)
+  bad = bad_primes(L; discriminant = true, dyadic = true)
 
   K = base_field(L)
   k = base_field(K)
@@ -1470,7 +1470,7 @@ end
 function _check_global_genus(LGS, signatures)
   _non_norm = _non_norm_primes(LGS, ignore_split = true)
   P = length(_non_norm)
-  I = length([(s, N) for (s, N) in signatures if isodd(mod(N, 2))])
+  I = count(v -> isodd(mod(v[2], 2)), signatures)
   if mod(P + I, 2) == 1
     return false
   end
@@ -1547,7 +1547,7 @@ function _hermitian_form_with_invariants(E, dim, P, N)
       push!(D, el)
     end
   end
-  push!(D, a * prod(D, init = one(E)))
+  push!(D, a * prod(D; init = one(E)))
   Dmat = diagonal_matrix(D)
   dim0, P0, N0 = _hermitian_form_invariants(Dmat)
   @assert dim == dim0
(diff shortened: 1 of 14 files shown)


src/QuadForm/Herm/Lattices.jl
```diff
@@ -227,8 +227,8 @@ function norm(L::HermLat)
   K = base_ring(G)
   R = base_ring(L)
   C = coefficient_ideals(L)
-  to_sum = sum(G[i, i] * C[i] * v(C[i]) for i in 1:length(C))
-  to_sum = to_sum + R * reduce(+, [tr(C[i] * G[i, j] * v(C[j])) for j in 1:length(C) for i in 1:(j-1)], init = ZZ(0)*inv(1*base_ring(R)))
+  to_sum = sum(G[i, i] * C[i] * v(C[i]) for i in 1:length(C); init = 0*R)
```
thofma, line 230: It must be initialised as a fractional ideal.

src/QuadForm/Herm/Lattices.jl
```diff
@@ -359,7 +359,7 @@ function jordan_decomposition(L::HermLat, p)
 
   k = 1
   while k <= n
-    G = S * F * transpose(_map(S, aut))
+    G = S * F * transpose(map(aut, S))
```
thofma, line 362: What did `_map` do?

src/QuadForm/Lattices.jl
```diff
@@ -305,8 +305,8 @@ function generators(L::AbstractLat; minimal::Bool = false)
     d = ncols(St)
     for i in 1:nrows(St)
       if base_ring(L) isa NfOrd
-        I = numerator(St.coeffs[i])
-        den = denominator(St.coeffs[i])
+        I = numerator(coefficient_ideals(St)[i])
```
thofma, line 308: Please check that this does not make a copy

src/QuadForm/Quad/GenusRep.jl
```diff
@@ -155,7 +155,7 @@ function spinor_genera_in_genus(L, mod_out)
   # The smaller the element, the better
   for d in diagonal(Gr)
     if (iszero(spinornorm) && !iszero(d)) || (!iszero(d) && abs(norm(d)) < abs(norm(spinornorm)))
-      spinornorm = d
+      add!(spinornorm, spinornorm, d)
```
thofma, line 158: Why add? Wasn't there a `=` before?

src/QuadForm/Quad/GenusRep.jl
```diff
@@ -167,7 +167,7 @@ function spinor_genera_in_genus(L, mod_out)
       end
     end
     @assert !iszero(Gr[1, i])
-    spinornorm = 2 * Gr[1, i]
+    add!(spinornorm, spinornorm, 2 * Gr[1, i])
```
thofma, line 170: As above

src/QuadForm/Quad/GenusRep.jl
```diff
@@ -511,7 +511,7 @@ function maximal_norm_splitting(L, p)
     end
   end
 
-  @assert all(let G = G; k -> nrows(G[k]) in [1,2]; end, 1:length(G))
+  @assert all(k -> nrows(G[k]) in [1,2], 1:length(G))
```
thofma, line 514: Did you check with `code_warntype`? This is added to work around a julia bug.

src/QuadForm/Quad/GenusRep.jl
```diff
@@ -769,8 +769,8 @@ function _map_idele_into_class_group(mRCG, idele, atinfinity::Vector{Tuple{T, In
   s = s * t
 
   # Check if everything is ok.
-  @hassert :GenRep 1 all(let s = s; k -> isone(quo(R, factors[k])[2](FacElem(s * the_idele[k]))); end, 1:length(the_idele))
-  @hassert :GenRep 1 all(let s = s, the_idele_inf = the_idele_inf; j -> sign(s * the_idele_inf[j], IP[j]) == 1 end, 1:length(IP))
+  @hassert :GenRep 1 all(k -> isone(quo(R, factors[k])[2](FacElem(s * the_idele[k]))), 1:length(the_idele))
+  @hassert :GenRep 1 all(j -> sign(s * the_idele_inf[j], IP[j]) == 1, 1:length(IP))
```
thofma, line 773: As above

src/QuadForm/Quad/Lattices.jl
```diff
@@ -232,8 +232,7 @@ function scale(L::QuadLat)
   end
   G = gram_matrix_of_rational_span(L)
   C = coefficient_ideals(L)
-  to_sum = [ G[i, j] * C[i] * involution(L)(C[j]) for j in 1:length(C) for i in 1:j]
-  s = sum(to_sum)
+  s = sum(G[i, j] * C[i] * involution(L)(C[j]) for j in 1:length(C) for i in 1:j)
```
thofma, line 235: Needs an `init`

src/QuadForm/Quad/Spaces.jl
```diff
@@ -613,13 +612,13 @@ end
 function _quadratic_form_with_invariants(dim::Int, det::nf_elem, finite::Vector, negative::Dict{<:InfPlc, Int})
   @hassert :Lattice 1 dim >= 1
   @hassert :Lattice 1 !iszero(det)
-  K::AnticNumberField = parent(det)
+  K = parent(det)
```
thofma, line 615: This was added for type stability. Have you checked that everything is still fine?

src/QuadForm/Quad/ZGenus.jl
```diff
@@ -135,20 +134,19 @@ function _p_adic_symbol(A::MatElem, p, val)
   q = p^m0
   n = nrows(A)
   A = divexact(A, q)
-  Fp = Native.GF(p)
+  Fp = GF(p)
```
thofma, line 137: No, should be `Native.GF`

src/QuadForm/Quad/ZGenus.jl
```diff
@@ -192,32 +195,31 @@ OUTPUT:
...
   if n == 0
-    return [[0, 0, 1, 0, 0]]
+    return Vector{Int}[Int[0, 0, 1, 0, 0]]
   end
-  m0 = minimum([ valuation(c,2) for c in A if c!=0])
+  m0 = minimum(valuation(c,2) for c in A if c!=0)
   q = ZZ(2)^m0
   A = divexact(A, q)
-  A_2 = change_base_ring(Native.GF(2), A)
+  A_2 = change_base_ring(GF(2), A)
```
thofma, line 207: No

src/QuadForm/Quad/ZGenus.jl
```diff
@@ -2148,11 +2150,11 @@ function _M_p(species, p)
   p = QQ(p)
```
thofma, line 2150: `_p` etc while you are at it

src/QuadForm/Quad/ZGenus.jl
```diff
@@ -2669,13 +2670,7 @@ function represents(G1::ZZGenus, G2::ZZGenus)
...
   end
 
-  primes = [prime(s) for s in local_symbols(G1)]
-  for s in local_symbols(G2)
-    p = prime(s)
-    if !(p in primes)
-      push!(primes, p )
-    end
-  end
+  primes = union!(Hecke.primes(G1), Hecke.primes(G2))
```
thofma, line 2673:
```suggestion
  primes = union!(primes(G1), primes(G2))
```

thofma, line 2673: Does `primes(G1)` make a copy? If not this is wrong

src/QuadForm/Quad/ZLattices.jl
```diff
@@ -437,8 +439,8 @@ function assert_has_automorphisms(L::ZZLat; redo::Bool = false,
   end
 
   # Now gens are with respect to the basis of L
-  @hassert :Lattice 1 all(let gens = gens; i -> change_base_ring(FlintQQ, gens[i]) * GL *
```
thofma, line 440: As above

src/QuadForm/Quad/ZLattices.jl
```diff
@@ -1198,7 +1201,7 @@ function is_maximal_even(L::ZZLat, p)
     return true, L
   end
   G = change_base_ring(ZZ, gram_matrix(L))
-  k = Native.GF(p)
+  k = GF(p)
```
thofma, line 1204: No

src/QuadForm/Torsion.jl
```diff
@@ -1501,7 +1500,7 @@ Return the radical `\{x \in T | b(x,T) = 0 and q(x)=0\}` of the quadratic form
 function radical_quadratic(T::TorQuadModule)
   Kb, ib = radical_bilinear(T)
   G = gram_matrix_quadratic(Kb)*1//modulus_bilinear_form(Kb)
-  F = Native.GF(2, cached=false)
+  F = GF(2; cached=false)
```
thofma, line 1503: No

thofma (commented):
Most of this looks innocent. Most of the `let` statements came from painstakingly working through issues with `@code_warntype`. So please make sure that code stability is not degraded. Same is true for `sum(iterator)` vs `sum(Bla[x for x in iterator])`. There once were some compiler bugs. Not sure if they are around anymore.

StevellM:
I will check all that, yes, thanks for all the comment.

src/QuadForm/Quad/GenusRep.jl
StevellM: `spinornorm` is already initialized to be zero, I thought that doing so would lighten the allocation process (?)
 `add!` will just modify the content of `spinornorm`

thofma: No. Nothing is gained from this. It is actually worse with the change. Before it was just an assignment, which is a noop (no operation; does not cost anything). Now there is an addition, which is not for free.

src/QuadForm/Herm/Lattices.jl
StevellM: The same, but I run several examples using both functions and the underscore one seems to be slightly faster, allocations are the same... I will revert those changed. I first thought that this was a work-around before `map` was fully implemented.

src/QuadForm/Lattices.jl
StevellM: No copy, access directly `St.coeffs`

src/QuadForm/Quad/GenusRep.jl
StevellM: I ran the `code_warntype` and notice wrong appears.. The assertion works also (like if I change some values it returns an error, and does nothing otherwise). So it seems to be fixed (?)

src/QuadForm/Quad/Spaces.jl
StevellM: I will have to re-put it. There are very odd things happening in this function, which I barely understand on the `code_warntype`. The variable `finite0` is defined and used only at the very end of the function in a `hassert`. If we remove the `hassert` and infer the type for `finite0`, for instance
```
local finite0::Vector{<: NfOrdIdl}
finite0 = copy(finite)
```
or
```
local finite0::typeof(finite)
finite0 = copy(finite)
```
or 
```
finite0 = eltype(finite)[p for p in finite]
```
in all cases, if we remove the only line calling this variable, then `code_warntype` will still show `finite0::Any` in red...

src/QuadForm/Quad/ZGenus.jl
StevellM: I will revert them. What is the reason for this ?

src/QuadForm/Quad/ZGenus.jl
```diff
@@ -1693,9 +1695,9 @@ function _gram_from_jordan_block(p::ZZRingElem, block, discr_form=false)
       q = zero_matrix(QQ, 0, 0)
     end
     if discr_form
-        q = q * (1//2)^level
+      map_entries!(x -> x * (1//2)^level, q, q)
```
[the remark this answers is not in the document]
StevellM: It works, one has to take care of the type but if everything is made of `QQFieldElem`, then this overwrites the content of `q`; the tests show that nothing goes wrong.

src/QuadForm/Quad/ZGenus.jl
StevellM: Yes.

src/QuadForm/Quad/GenusRep.jl
thofma: ok

src/QuadForm/Quad/Spaces.jl
thofma: This is all related to https://github.com/JuliaLang/julia/issues/15276. Every time one uses a `p -> begin; x, y, z; end`, every "captured" variable which is used (like `x`, `y`, `z`), might become unstable in the surrounding function. This also effects generators like `(p + z for p in 1:10)`. The `let` is a workaround, also mentioned in the issue above.

StevellM: I see, I completely forgot about this. So maybe the `code_warntype` might not capture the issue. Since it was a recent fix I will put this back too until we make sure the problem is solved (to avoid any performance issue)

thofma:
Not sure if you were finished with the revision?

StevellM:
Sorry, I forgot to mention it - I am done with the revision.

src/QuadForm/Herm/Genus.jl
```diff
@@ -1694,14 +1694,14 @@ function hermitian_local_genera(E, p, rank::Int, det_val::Int, min_scale::Int, m
...
-          push!(dets, [1, -1])
+          push!(dets, Int[1, -1])
         end
         if mod(b[1], 2) == 1
-          push!(dets, [hyperbolic_det^(div(b[2], 2))])
+          push!(dets, Int[hyperbolic_det^(div(b[2], 2))])
         end
       end
 
-      for d in cartesian_product_iterator(dets, inplace = true)# Iterators.product(dets...)
```
thofma, line 1704: Hmm, why?

src/QuadForm/Herm/Genus.jl
```diff
@@ -1748,7 +1748,7 @@ function hermitian_local_genera(E, p, rank::Int, det_val::Int, min_scale::Int, m
       end
     end
 
-    for dn in cartesian_product_iterator(det_norms, inplace = true)
```
thofma, line 1751: Same here, why?

src/QuadForm/Herm/Mass.jl
```diff
@@ -292,7 +292,7 @@ function _standard_mass(L::HermLat, prec::Int = 10)
   local relzeta::arb
 
   while true
-    relzeta = prod(arb[_L_function(E, 1 - i, wprec) for i in 1:2:m])
+    relzeta = prod(_L_function(E, 1 - i, wprec) for i in 1:2:m; init = 1)
```
thofma, line 295:
```suggestion
    relzeta = prod(_L_function(E, 1 - i, wprec) for i in 1:2:m; init = one(ArbField(wprec, cached = false)))
```

thofma:
I added a few things. There are still some `map` instead of `_map`. Do you want to change it?

src/QuadForm/Herm/Genus.jl
StevellM: `inplace = true` is the default for the function so I do not see why keeping it

src/QuadForm/Herm/Mass.jl
StevellM: Ah thanks, that was the thing I was trying to get but couldn't make it work.

StevellM:
Yes I guess so, I just missed them in my last commit. Since we have to make a new commit for your suggestion, I will put everything together.

src/QuadForm/Herm/Genus.jl
thofma: Sorry, I forgot about that, you are right.

thofma:
71 min for the long tests is pretty quick. Not sure it is due to the changes here.

PR #1258, opened 2023-10-23
Improve class group proof

thofma:
- Make some things more thread safe
- Parallel class group proof
- Clean up docstring for `class group`


fieker:
I have a feeling we need to re-visit sime stuff. The proof uses `class_group_small_real_elements_relation_start`, the computation gave up on that ages ago, it is using LLL instead...
`small_real` uses, probably even my ancient, lattice code

thofma:
Yes, I also have some questions. We should discuss this (I wrote you on slack).

PR #1262, opened 2023-10-27
Fix `kronecker_product(::SMat, ::SMat)`

lgoettgens:
Another nasty bug in that function. I added the example that confused me as a test.

I am looking forward to a fast merge and release of this so that I can use it in https://github.com/oscar-system/Oscar.jl/pull/2936.


thofma:
If you bump the version in the PR, it will be even faster.

lgoettgens:
The failed 1.9-macOS job seems to be unrelated to the change

thofma:
Not sure why the "long tests" are failing so often, but once they succeed I will merge.

PR #1263, opened 2023-10-27
add `dot(x, A, y)` for sparse matrices

felix-roehrich:
This PR adds the following dispatches
```julia
dot(x::SRow{T}, A::SMat{T}, y::SRow{T}) where T
dot(x::MatrixElem{T}, A::SMat{T}, y::MatrixElem{T}) where T
dot(x::AbstractVector{T}, A::SMat{T}, y::AbstractVector{T}) where T
```

If you have any suggestions for improvements, let me know.

### DIFF
--- a/src/Sparse/Matrix.jl
+++ b/src/Sparse/Matrix.jl
@@ -725,6 +725,70 @@ end
 
 ################################################################################
 #
+#  Dot product
+#
+################################################################################
+
+function dot(x::SRow{T}, A::SMat{T}, y::SRow{T}) where T
+  v = T(0)
+
+  px = 1
+  py = 1
+  for i in 1:length(A.rows), j in 1:length(A[i].pos)
+    while px <= length(x.pos) && x.pos[px] < A[i].pos[j]
+      px += 1
+    end
+    if px > length(x.pos)
+      break
+    end
+
+    while py <= length(y.pos) && y.pos[py] < A[i].pos[j]
+      py += 1
+    end
+    if py > length(y.pos)
+      break
+    end
+
+    if x.pos[px] == A[i].pos[j] == y.pos[py]
+      v += x.values[px] * A[i].values[j] * y.values[py]
+    end
+  end
+
+  return v
+end
+
+function dot(x::AbstractVector{T}, A::SMat{T}, y::AbstractVector{T}) where T
+  @assert length(x) == length(y)
+
+  v = T(0)
+  for i in 1:length(A.rows), j in 1:length(A[i].pos)
+    if A[i].pos[j] > length(x) || A[i].pos[j] > length(y)
+      error("incompatible matrix dimensions")
+    end
+    v += x[A[i].pos[j]] * A[i].values[j] * y[A[i].pos[j]]
+  end
+
+  return v
+end
+
+# support dot product for vector matrices
+function dot(x::MatrixElem{T}, A::SMat{T}, y::MatrixElem{T}) where T
+  @assert length(x) == length(y)
+  len = length(x)
+
+  v = T(0)
+  for i in 1:length(A.rows), j in 1:length(A[i].pos)
+    if A[i].pos[j] > len || A[i].pos[j] > len
+      error("incompatible matrix dimensions")
+    end
+    v += x[A[i].pos[j]] * A[i].values[j] * y[A[i].pos[j]] # this will throw if x or y is not a vector
+  end
+
+  return v
+end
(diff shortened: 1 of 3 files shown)


src/Sparse/Matrix.jl
```diff
@@ -723,6 +723,98 @@ function *(A::SMat, b)
...
+#
+################################################################################
+
+@doc raw"""
+    dot(x::SRow{T}, A::SMat{T}, y::SRow{T}) where T -> T
+
+Return the generalized dot product `dot(x, A*y)`.
+"""
+function dot(x::SRow{T}, A::SMat{T}, y::SRow{T}) where T
+  v = T(0)
```
thofma, line 738:
```suggestion
  v = zero(T)
```

src/Sparse/Matrix.jl
```diff
@@ -723,6 +723,98 @@ function *(A::SMat, b)
...
+    while px <= length(x.pos) && x.pos[px] < i
+      px += 1
+    end
+    if px > length(x.pos)
+      break
+    elseif x.pos[px] > i
+      continue
+    end
+
+    s = T(0)
```
thofma, line 750:
```suggestion
    s = zero(T)
```

src/Sparse/Matrix.jl
```diff
@@ -723,6 +723,98 @@ function *(A::SMat, b)
...
+
+@doc raw"""
+    dot(x::AbstractVector{T}, A::SMat{T}, y::AbstractVector{T}) where T -> T
+
+Return the generalized dot product `dot(x, A*y)`.
+"""
+function dot(x::AbstractVector{T}, A::SMat{T}, y::AbstractVector{T}) where T
+  @req length(x) == length(A.rows) == length(y) "incompatible matrix dimensions"
+
+  v = T(0)
```
thofma, line 779:
```suggestion
  v = zero(T)
```

src/Sparse/Matrix.jl
```diff
@@ -723,6 +723,98 @@ function *(A::SMat, b)
...
+@doc raw"""
+    dot(x::MatrixElem{T}, A::SMat{T}, y::MatrixElem{T}) where T -> T
+
+Return the generalized dot product `dot(x, A*y)`.
+"""
+function dot(x::MatrixElem{T}, A::SMat{T}, y::MatrixElem{T}) where T
+  @req length(x) == length(A.rows) == length(y) "incompatible matrix dimensions"
+  len = length(x)
+
+  v = T(0)
```
thofma, line 803:
```suggestion
  v = zero(T)
```

src/Sparse/Matrix.jl
```diff
@@ -723,6 +723,98 @@ function *(A::SMat, b)
...
+  for i in 1:length(A.rows)
+    s = T(0)
```
thofma, line 805:
```suggestion
    s = zero(T)
```

src/Sparse/Matrix.jl
```diff
@@ -723,6 +723,98 @@ function *(A::SMat, b)
...
+"""
+function dot(x::AbstractVector{T}, A::SMat{T}, y::AbstractVector{T}) where T
+  @req length(x) == length(A.rows) == length(y) "incompatible matrix dimensions"
+
+  v = T(0)
+  for i in 1:length(A.rows)
+    s = T(0)
+    for j in 1:length(A[i].pos)
+      if A[i].pos[j] > length(y)
+        error("incompatible matrix dimensions")
```
thofma, line 784: Is there a reason we do not check this at the start?

thofma (approved):

thofma:
@lgoettgens maybe you also want to have an artificial look? Since you also implemented a few things recently.

src/Sparse/Matrix.jl
felix-roehrich: Because `A` is sparse, we would can't avoid looping over all rows to check this. So I thought we might as well check this on the fly with the assumption that we won't error.

felix-roehrich: Never mind looking at the some code it looks like `A.c` stores this ? If I read this right, I will change it.

thofma:
Yes, but please use nrows and ncols functions.

thofma:
Do you want a quick release to use in Oscar?

felix-roehrich:
> Do you want a quick release to use in Oscar?

Thank you, but this is not necessary.

PR #1265, opened 2023-10-30
isupper_triangular -> is_upper_triangular, isunit -> is_unit

fingolfin:
The former name has been "deprecated" for a long time now.


fingolfin:
The OscarCI failures seems to be due to an issue with the Julia package servers

PR #1267, opened 2023-10-31
Kaiser

fieker:
the example is not in test as this requires a class group of a degree 24 field... (splitting field is deg 12, plus zeta_3 on top)

Ignore the Miho stuff in examples. It works but is in examples


thofma:
Can you add the Lorenz module example to the tests?

fieker:
if you want...

PR #1268, opened 2023-10-31
Change residue_field for local fields

thofma:
I want to compute $p$-adic regulators for large $p$. This cannot work with the current design, so we change `residue_field` to  use the "new" finite fields.

There is one remaining issue in the computation of the fundamental class. @fieker let's talk about this on Thursday.


src/LocalField/neq.jl
```diff
@@ -796,6 +798,7 @@ function local_fundamental_class_serre(mKL::LocalFieldMor)
   rE, mE = residue_field(E)
   rL, mL = residue_field(L)
   rK, mK = residue_field(K)
+  # how is this supposed to work?
   mrKL = hom(rK, rL, mL(mKL(preimage(mK, gen(rK)))))
```
thofma, line 802: Without promotion magic, this only works if `base_field(rK) === base_field(rL)`, but this is not the case here.

PR #1270, opened 2023-11-01
Replace more uses of isFOO by is_foo

fingolfin:


fingolfin:
There is a failure in `
Run long tests / test (1.9, x64, macOS-latest) (pull_request)` but I can't tell what it is / if it is related to this PR or not. 

@thofma perhaps you could restart it? It would be nice to get this merged so I can proceed with the alias deprecations.

fingolfin:
@thofma it seems you already restarted it twice before and it keeps running into an infinite loop. So perhaps there is a genuine issue here? But what could that be? weird

PR #1271, opened 2023-11-02
Random lines in `LineEnumCtx`

StevellM:
To prepare random Kneser's neighbour method, allow to have access to random lines from the `LineEnumCtx`. It might not be the most optimized way but at least it works.

### DIFF
--- a/src/QuadForm/LineOrbits.jl
+++ b/src/QuadForm/LineOrbits.jl
@@ -7,7 +7,7 @@
 # We iterate through K by adding 1 in the prime field case and by multiplying
 # with a primitive element in the general case.
 
-function LineEnumCtx(K::T, n) where {T}
+function LineEnumCtx(K::T, n::Int) where {T}
   a = primitive_element(K)
   v = Vector{elem_type(K)}(undef, n)
   for i in 1:n
@@ -16,7 +16,7 @@ function LineEnumCtx(K::T, n) where {T}
   depth = n + 1
   dim = n
   q = order(K)
-  length = divexact(BigInt(q)^n - 1, q - 1)
+  length = divexact(ZZ(q)^n - 1, q - 1)
   return LineEnumCtx{T, elem_type(T)}(K, a, dim, depth, v, length)
 end
 
@@ -29,7 +29,7 @@ function LineEnumCtx(K::T, n::Int) where {T <: Union{fpField, FpField}}
   depth = n + 1
   dim = n
   q = order(K)
-  length = divexact(BigInt(q)^n - 1, q - 1)
+  length = divexact(ZZ(q)^n - 1, q - 1)
   return LineEnumCtx{T, elem_type(T)}(K, a, dim, depth, v, length)
 end
 
@@ -54,7 +54,7 @@ function Base.show(io::IO, P::LineEnumCtx)
   end
 end
 
-Base.length(P::LineEnumCtx) = P.length
+Base.length(P::LineEnumCtx) = BigInt(P.length)
 
 Base.eltype(::Type{LineEnumCtx{T, S}}) where {T, S} = Vector{S}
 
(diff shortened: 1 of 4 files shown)


thofma:
Why `BigInt` instead of `ZZ`? `BigInt` will be slower. You can use `IntegerUnion` instead of `Integer` if you want to support both.

StevellM:
I have not thought about this actually, I did not know it could make a difference. You mean for the operations in the `getindex` function ? For the size of the `LineEnumCtx`, I tried putting something not `integer` and some tests did not pass.

src/QuadForm/LineOrbits.jl
```diff
@@ -16,7 +16,7 @@ function LineEnumCtx(K::T, n) where {T}
   depth = n + 1
   dim = n
   q = order(K)
-  length = divexact(q^n - 1, q - 1)
+  length = divexact(BigInt(q)^n - 1, q - 1)
```
thofma, line 19: Here

src/QuadForm/LineOrbits.jl
```diff
@@ -29,7 +29,7 @@ function LineEnumCtx(K::T, n::Int) where {T <: Union{fpField, FpField}}
...
```
thofma, line 32: Here

src/QuadForm/LineOrbits.jl
```diff
@@ -58,12 +58,38 @@ Base.length(P::LineEnumCtx) = P.length
...
+  1 <= i <= length(P) || Base.throw_boundserror(P, i)
+  K = base_field(P)
+  v = fill(zero(K), dim(P))
+  if i == 1
+    v[end] = one(K)
+    return v
+  end
+  p = size(K)
+  j = i-2
+  n = findfirst(n -> sum(BigInt(p)^i for i in 1:n) > j, 1:256)
```
thofma, line 79: Here

src/QuadForm/LineOrbits.jl
```diff
@@ -58,12 +58,38 @@ Base.length(P::LineEnumCtx) = P.length
...
+  v[end-n] = one(K)
+  j = n == 1 ? j : j-sum(BigInt(p)^k for k in 1:(n-1))
```
thofma, line 81: Here

test/QuadForm/LineOrbits.jl
```diff
@@ -0,0 +1,14 @@
+@testset "Random lines" begin
+  P = Hecke.enumerate_lines(GF(5), 20)
+  io = IOBuffer()
+  show(io, MIME"text/plain"(), P)
+  @test length(String(take!(io))) == 98
+  show(IOContext(io, :supercompact => true), P)
+  @test length(String(take!(io))) == 25
+  show(io, P)
+  @test length(String(take!(io))) == 33
```
thofma, line 9: I don't understand this test :)

StevellM: I wanted to increase test coverage, but maybe I can remove the test on the length of the strings and just leave the test calling the show functions (kind of tautological tests but it is "easy" test coverage for the corresponding lines)

src/QuadForm/LineOrbits.jl
StevellM: This I cannot remove, otherwise all functions like `collect`, or `Vector{Int}(undef, length(P))` will not work if we set the length to be a `ZZRingElem`

src/QuadForm/Types.jl
```diff
@@ -444,7 +444,7 @@ mutable struct LineEnumCtx{T, S}
   dim::Int
   depth::Int
   v::Vector{S}
-  length::BigInt
+  length::IntegerUnion
```
fingolfin, line 447: Doing this means you just made anything use `length` type unstable, unless you add a type assertion or a type conversion to every place that accesses `length`. Perhaps that's OK, perhaps not -- you could avoid it by adding the type of `length` as another type parameter to `LineEnumCtx`.

src/QuadForm/LineOrbits.jl
```diff
@@ -54,16 +54,30 @@ function Base.show(io::IO, P::LineEnumCtx)
   end
 end
 
-Base.length(P::LineEnumCtx) = P.length
+Base.length(P::LineEnumCtx) = BigInt(P.length)
```
fingolfin, line 57: You said you cannot change this because otherwise `collect` won't work. But if `P.length` truly does not fit into an `Int`, then `collect` won't work anyway because you just tried to use more RAM than humanity has created in total in the course of its existence.

At the same time, you slow down the common case where it *does* fit.

You could just do this:
```suggestion
Base.length(P::LineEnumCtx) = Int(P.length)
```
That should work for any "small" value of length, and will otherwise give an error. Of course if any code wants the "true" length then it needs to get it directly or use a different accessor.

StevellM: As far as I have seen, the `length` itself is only use in a `_line_orbits` code where we create a vector of length `length`, which should fit in what you say. I guess replacing by `Int` should be fine for that case since it was working in that way before. Thank you for pointing that out.

PR #1304, opened 2023-12-01
Update the banners in examples

StevellM:
On the GitHub page or in the html documentation, the banner of Hecke was not up to date. Similarly for an example on how to call the documentation of a function.

Also, I remarked that the html documentation I have added for `MSet` was not appearing: I forgot to add a line in a file. This should be good now (hopefully)


thofma:
Thanks

PR #1316, opened 2023-12-08
Compute LLL basis in subfields if the field is large

thofma:


thofma:
@fieke any objections?

thofma:
I am happy to have a faster method with does not use linear algebra and the maximal order.

PR #1317, opened 2023-12-08
Fix issue in roots of nonmonic polynomials

thofma:


src/NumField/Elem.jl
```diff
@@ -460,8 +460,7 @@ end
 function roots(f::PolyRingElem{<: NumFieldElem})
   lf = factor(f)
   @assert degree(unit(lf)) == 0
-  scale = inv(coeff(unit(lf), 0))
-  return elem_type(base_ring(f))[-constant_coefficient(x)*scale for x = keys(lf.fac) if degree(x) == 1]
+  return elem_type(base_ring(f))[-constant_coefficient(x)*leading_coefficient(x) for x = keys(lf.fac) if degree(x) == 1]
```
StevellM, line 463: Shouldn't it be a division here ?

fieker: correct, but I think the factors tend to me monic

PR #1322, opened 2023-12-13
sparse_row and SRow accept NCRing instead of Ring 

Lax202:
This is to be able to create Modules in Oscar over non-commutative rings, as discussed with @HechtiDerLachs and @fieker


lgoettgens (changes requested):
There is some of functionality for `SRow`s that expects the coefficient ring to be commutative. These need to be adapted or restricted to elements of subtypes of `Ring`.
Some that I found in a quick look:
https://github.com/Lax202/Hecke.jl/blob/41678ac4bc55f85a04da7074cdf93f7849ef6406/src/Sparse/Row.jl#L397-L399
Is `scale_row!`, `div`, `divexact` based on left- or right-multiplication? This should be decided and added to the docstrings. Furthermore, there should be functions for the other case as well.

fingolfin:
In particular things like `scale_row!` may need two versions, for left and right action...

A crucial aspect of modules over non-commutative rings is that one has to be careful about which actions are left and which are right otherwise things don't work out in the end...

(@mohamed-barakat is an expert on this, perhaps we can consult with him, too ;-) but in principle I think I know the rules, too)

Lax202:
I have added left and right actions in scale_row, add_scaled_row and divexact (not div). There are tests for everything except divexact since I am unable to find a non-commutative ring for which divexact_left is implemented. 
The tests use an exterior algebra. This ring does not exist in Hecke. If someone could suggest an Hecke NCRing that can be used for tests, that would be great.

thofma:
`R = MatrixAlgebra(QQ, 2)` and `a = R([1 2; 3 4])` if you want an element. Or `rand(R, -10:10)` for something random.

fingolfin:
Someone needs to allow the tests to run :-)

src/Sparse/Row.jl
```diff
@@ -419,15 +419,41 @@ end
...
+    return a
+  end
+  for i=1:length(a.pos)
+    a.values[i] = b*a.values[i]
+  end
+  return a
+end
+
+@doc raw"""
+  scale_row_right!(a::SRow, b::NCRingElem) -> SRow
```
fingolfin, line 440:
```suggestion
    scale_row_right!(a::SRow, b::NCRingElem) -> SRow
```
There must always be exactly four spaces for the indentation here

thofma:
As said, the tests will fail, so no need to approve anything. I have to reapprove for the next commit anyway.

mohamed-barakat:
> A crucial aspect of modules over non-commutative rings is that one has to be careful about which actions are left and which are right otherwise things don't work out in the end...
> 
> (@mohamed-barakat is an expert on this, perhaps we can consult with him, too ;-) but in principle I think I know the rules, too)

I am happy to help, of course :) If you have specific questions please let me know.

src/Sparse/Row.jl
```diff
@@ -555,7 +587,33 @@ function divexact(A::SRow{T}, b::Integer) where T
...
     return deepcopy(A)
   end
-  return divexact(A, base_ring(A)(b))
+  return divexact_left(A, base_ring(A)(b))
+end
+
+@doc raw"""
+  divexact_right(A::SRow, b::NCRingElem) -> SRow
+
+Returns the left divexact $A/b$
```
joschmitt, line 596:
```suggestion
Returns the right divexact $A/b$
```

src/Sparse/Row.jl
```diff
@@ -537,13 +564,18 @@ function div(A::SRow{T}, b::Integer) where T
   return div(A, base_ring(A)(b))
 end
 
+@doc raw"""
+  divexact(A::SRow, b::NCRingElem) -> SRow
+
+Returns the left divexact $A/b$
```
joschmitt, line 570:
```suggestion
Returns the left divexact $b\A$
```
(I think???)

joschmitt:
@Lax202 I just pushed a commit to resolve the conflicts.

src/Sparse/Row.jl
lgoettgens: no idea. maybe @thofma or @fingolfin know?

src/Sparse/Row.jl
```diff
@@ -553,7 +585,33 @@ function divexact(A::SRow{T}, b::Integer; check::Bool=true) where T
...
     return deepcopy(A)
   end
-  return divexact(A, base_ring(A)(b); check=check)
+  return divexact_left(A, base_ring(A)(b), check=check)
+end
+
+@doc raw"""
+  divexact_right(A::SRow, b::NCRingElem; check::Bool = true) -> SRow
+
+Returns the right divexact $A/b$
```
fingolfin, line 594: Also change this
```suggestion
Return $C$ such that $A = C \cdot b.
```

src/Sparse/Row.jl
```diff
@@ -513,6 +539,7 @@ function *(A::SRow, b)
   return A*base_ring(A)(b)
 end
 
+#left and right div not implimented
```
fingolfin, line 542:
```suggestion
#left and right div not implemented
```

src/Sparse/Row.jl
fingolfin: I would follow what we say in other docstrings for `divexact` and friends and be explicit about the defining property of the return value:
```suggestion
Return a sparse row $C$ such that $a = b \cdot C$.
```
Or drop the "sparse row" (depending on doc strings for `SRow` do):
```suggestion
Return $C$ such that $a = b \cdot C$.
```

src/Sparse/Row.jl
```diff
@@ -535,13 +562,18 @@ function div(A::SRow{T}, b::Integer) where T
   return div(A, base_ring(A)(b))
 end
 
+@doc raw"""
+  divexact(A::SRow, b::NCRingElem; check::Bool = true) -> SRow
+
+Returns the left divexact $A/b$
+"""
 function divexact(A::SRow{T}, b::T; check::Bool=true) where T
```
fingolfin, line 570: Hmm, IMHO this code is missing a `divexact_left` method. I suggest to rename this one (also in the docstring), and perhaps restrict `T` to `NCRingElem`.
```suggestion
function divexact_left(A::SRow{T}, b::T; check::Bool=true) where T <: NCRingElem
```

And then implement `divexact` by delegating to `divexact_left`... possibly with `T` restricted to `RingElem` -- but I'd also be fine with having it general, but then the docstring for this `divexact` method should explicitly say that it delegates to `divexact_left`.
```
divexact(A::SRow{T}, b::T; check::Bool=true) where T <: RingElem = divexact_left(A, b; check)
```

Of course the delegation could also be done in the other direction, that's how it is done right now in AA for "ring elements", but I think the situation there is a bit different: there `divexact` is already implemented for commutative rings and we just allow `divexact_left` to work; but for non-commutative rings / ring elements, `divexact` just is never defined... So I don't think this pattern matters here, hence my suggestion above.
```
divexact_left(x::T, y::T; check::Bool=true) where T <: RingElement = divexact(x, y; check=check)

divexact_right(x::T, y::T; check::Bool=true) where T <: RingElement = divexact(x, y; check=check)
```

src/Sparse/Row.jl
```diff
@@ -620,7 +679,49 @@ function add_scaled_row!(a::SRow{T}, b::SRow{T}, c::T) where T
...
 
-add_scaled_row!(a::SRow{T}, b::SRow{T}, c::T, tmp::SRow{T}) where T = add_scaled_row!(a, b, c)
+add_scaled_row!(a::SRow{T}, b::SRow{T}, c::T, tmp::SRow{T}) where T = add_scaled_row!(a, b, c) 
+
+add_right_scaled_row(a::SRow{T}, b::SRow{T}, c::T) where {T} = add_right_scaled_row!(a, deepcopy(b), c)
+
+@doc raw"""
+    add_right_scaled_row!(A::SRow{T}, B::SRow{T}, c::T) -> SRow{T}
+
+Returns the right scaled row $c A$ to $B$ by changing $B$ in place.
```
fingolfin, line 689:
```suggestion
Return the right scaled row $c A$ to $B$ by changing $B$ in place.
```

fingolfin:
@Lax202 it would be good if you implemented the suggestions from the feedback here or figure out why it can't be done. Feel free to talk to me about details.

Lax202:
@fingolfin I made the changes you suggested on this PR, and then resolved some merge conflict manually. I have to confess I don't know what's going on now. The tests are not running and I don't understand why.

src/Sparse/Row.jl
```diff
@@ -447,11 +450,32 @@ end
...
+For rows, the standard multiplication is from the left. 
+"""
 function scale_row!(a::SRow{T}, b::T) where T
   @assert !iszero(b)
   if isone(b)
-    return
+    return a
+  end
+  for i=1:length(a.pos)
+    a.values[i] = b*a.values[i]
```
fingolfin, line 465: As we just discussed orally: The problem is that in the meantime PR #1476 was merged which made a fix to `scale_row!` over rings with zero divisors, and this fix now ended up being only in `scale_row_right!` but not in this method.

src/Sparse/Row.jl
```diff
@@ -572,13 +607,20 @@ function div(A::SRow{T}, b::Integer) where T
   return div(A, base_ring(A)(b))
 end
 
-function divexact(A::SRow{T}, b::T; check::Bool=true) where T
+@doc raw"""
+  divexact(A::SRow, b::RingElem; check::Bool = true) -> SRow
```
fingolfin, line 611:
```suggestion
    divexact(A::SRow, b::RingElem; check::Bool = true) -> SRow
```

fingolfin:
Could someone (@thofma @joschmitt) please allow the CI tests to run?

src/Sparse/Row.jl
```diff
@@ -657,7 +726,49 @@ function add_scaled_row!(a::SRow{T}, b::SRow{T}, c::T) where T
...
+
+add_right_scaled_row(a::SRow{T}, b::SRow{T}, c::T) where {T} = add_right_scaled_row!(a, deepcopy(b), c)
+
+@doc raw"""
+    add_right_scaled_row!(A::SRow{T}, B::SRow{T}, c::T) -> SRow{T}
+
+Return the right scaled row $c A$ to $B$ by changing $B$ in place.
+"""
+
+function add_right_scaled_row!(a::SRow{T}, b::SRow{T}, c::T) where T
```
fingolfin, line 739: Instead of duplicating all this code, you could also use `Val` to do something like this:
```suggestion
function add_scaled_row!(a::SRow{T}, b::SRow{T}, c::T, ::Val{left} = Val(true)) where {T, left}
```
and then in the critical part you can do
```
  if left
    t = mul!(t, c, a.values[i])
  else
    t = mul!(t, a.values[i], c)
  end
```

But this can also wait for a future cleanup.
Then finally you can do
```julia
add_left_scaled_row!(a::SRow{T}, b::SRow{T}, c::T) where T = add_scaled_row!(a,b,c,Val(true))
add_right_scaled_row!(a::SRow{T}, b::SRow{T}, c::T) where T = add_scaled_row!(a,b,c,Val(false))
```

test/Sparse/Row.jl
```diff
@@ -173,4 +173,24 @@
   @test dense_row(A, 3) == matrix(FlintZZ, 1, 3, [-5, 0, 2])
   @test dense_row(A, 6) == matrix(FlintZZ, 1, 6, [-5, 0, 2, -10, 1, 0])
   @test sparse_row(dense_row(A, 6)) == A
+
+  SRow{NCRingElem}
```
lgoettgens, line 177:
```suggestion
  # SRow{NCRingElem}
```

fingolfin:
@Lax202 the documentation tests still have some failures you need to address (we discussed these already):
````
ExpandTemplates: expanding markdown templates.
Error: no docs found for 'sparse_row(::ZZRing, ::Vector{Tuple{Int, ZZRingElem}})' in `@docs` block in src/manual/misc/sparse.md:30-34
```@docs
sparse_row(::ZZRing, ::Vector{Tuple{Int, ZZRingElem}})
sparse_row(::ZZRing, ::Vector{Tuple{Int, Int}})
sparse_row(::ZZRing, ::Vector{Int}, ::Vector{ZZRingElem})
```
````

fingolfin:
@thofma @joschmitt @fieker can someone please approve the CI runs? @Lax202 just pushed an update which hopefully will address the documenter issue

fieker (approved):

PR #1324, opened 2023-12-15
Support for arbitrary ring morphisms with number field domains

thofma:


src/Map/NumField.jl
```diff
@@ -322,7 +322,7 @@ function map_data_type(T::Type{<: NfRel}, L::Type{<:Any})
   MapDataFromNfRel{elem_type(L), map_data_type(base_field_type(T), L)}
 end
 
-map_data_type(K::NfRel, L::NumField) = map_data_type(typeof(K), typeof(L))
+map_data_type(K::NfRel, L::Ring) = map_data_type(typeof(K), typeof(L))
```
lgoettgens, line 325: This is duplicate to line 1041

thofma: Hm yes, all the `f(x, y) = f(typeof(x), typeof(y))` could be consolidated. But I don't have time for this clean up now.

PR #1327, opened 2023-12-15
Absolute value in short vectors (iteration)

StevellM:


test/QuadForm/ShortVectors.jl
```diff
@@ -45,7 +45,7 @@
                      2448, 4320, 2982, 2402, 2364, 512, 256, 306 ]
 
   # Floating point bounds
-  L = integer_lattice(gram = QQ[1//2 0; 0 1//3])
```
thofma, line 48: I would duplicate the tests instead of removing the old tests.

thofma:
The "absolute quadratic form" sounds a bit odd, as does the "absolute squared length". But I guess it is what it is. You have an opinion on the terminology @simonbrandhorst?

test/QuadForm/ShortVectors.jl
StevellM: Good point, I will modify this.

src/QuadForm/ShortVectors.jl
```diff
@@ -136,7 +151,8 @@ end
 @doc raw"""
     kissing_number(L::ZZLat) -> Int
 
-Return the Kissing number of the sphere packing defined by `L`.
+Return the Kissing number of the sphere packing defined by the absolute value
+of the quadratic form defined by `L`.
```
simonbrandhorst, line 155:
```suggestion
Return the Kissing number of the sphere packing defined by `L`.
```

simonbrandhorst (approved):

PR #1328, opened 2023-12-15
Add comments on the implementation of local Hilbert symbols

StevellM:
Markus Kirschmer explained me how his code works, so I have added some comments related to this.
Now we have details about the process and why the output is what we want ( + some references for free)


thofma:
Superb!

StevellM:
Do we have to update and re-run all the checks for "nothing" ?

PR #1329, opened 2023-12-18
fix equality for ZZGenus and do not copy symbols

simonbrandhorst:


src/QuadForm/Quad/ZGenus.jl
```diff
@@ -833,8 +833,10 @@ function Base.:(==)(G1::ZZLocalGenus, G2::ZZLocalGenus)
   # This follows p.381 Chapter 15.7 Theorem 10 in Conway Sloane's book
   @req prime(G1) == prime(G2) ("Symbols must be over the same prime "
                                 *"to be comparable")
-  sym1 = filter!(g -> g[2] != 0, symbol(G1))
-  sym2 = filter!(g -> g[2] != 0, symbol(G2))
+
+  # make a copy and enforce sparsity
+  sym1 = [g for g in symbol(G1) if g[2] != 0]
+  sym2 = [g for g in symbol(G1) if g[2] != 0]
```
StevellM, line 839:
```suggestion
  sym2 = [g for g in symbol(G2) if g[2] != 0]
```

StevellM:
I know where the code is breaking in Oscar, I can make a short PR so that CI tests hopefully do not fail

src/QuadForm/Quad/ZGenus.jl
```diff
@@ -1103,10 +1097,10 @@ end
 @doc raw"""
     symbol(S::ZZLocalGenus) -> Vector{Vector{Int}}
 
-Return a copy of the underlying lists of integers for the Jordan blocks of `S`.
+Return a the underlying lists of integers for the Jordan blocks of `S`.
```
lgoettgens, line 1100:
```suggestion
Return the underlying lists of integers for the Jordan blocks of `S`.
```

thofma:
I just restarted the Oscar tests, and they are still failing. Any clue @StevellM?

StevellM:
Yes after testing it seems like `copy` is not enough on the fix in Oscar, we need a `deepcopy`. I will make a new change again.

thofma:
Merci, I restarted CI

PR #1332, opened 2023-12-20
Add missing 'check' argument to several divexact methods

fingolfin:
We need this in order to write generic code


fingolfin:
Turns out there were also a few `divexact` methods which had a `check` argument but it was positional, not a keyword argument... The second commit now fixes these. But I guess this makes this technically speaking a breaking change, in case anyone was using this third argument *sigh*.

fingolfin:
I will add compatibility shims to `src/Deprecations.jl` to avoid the breakage

thofma:
Thanks

PR #1336, opened 2023-12-22
Fixes is_commutative for 1-dim matrix algebras

SirToby25:
Previously, full matrix algebras were assumed to be not commutative. This PR changes this for degree-one matrix algebras.

As I found no tests for this function, I did not write any myself.


thofma:
Nice catch, thanks!

PR #1338, opened 2023-12-30
feat: adjust to finite field changes in Nemo

thofma:


src/NumField/Selmer.jl
```diff
@@ -281,16 +281,16 @@ function Hecke.is_power(a::FacElem{QQFieldElem, QQField}, p::Int)
   return true
 end
 
-function Hecke.is_power_with_root(a::FacElem{QQFieldElem, QQField}, p::Int)
+function is_power_with_root(a::FacElem{QQFieldElem, QQField}, p::Int)
```
lgoettgens, line 284:
```suggestion
function is_power(a::FacElem{QQFieldElem, QQField}, p::Int)
```
This should actually be renamed as well, similar to https://github.com/Nemocas/AbstractAlgebra.jl/pull/1529

src/EllCrv/Finite.jl
```diff
@@ -1148,17 +1148,17 @@ Return a list of generators of the group of rational points on $E$.
 julia> E = elliptic_curve(GF(101, 2), [1, 2]);
 
 julia> gens(E)
-2-element Vector{EllCrvPt{fqPolyRepFieldElem}}:
- Point  (93*o + 10 : 22*o + 69 : 1)  of Elliptic curve with equation
+2-element Vector{EllCrvPt{FqFieldElem}}:
+ Point  (16*o + 42 : 88*o + 97 : 1)  of Elliptic curve with equation
 y^2 = x^3 + x + 2
- Point  (89*o + 62 : 14*o + 26 : 1)  of Elliptic curve with equation
+ Point  (88*o + 23 : 94*o + 22 : 1)  of Elliptic curve with equation
```
lgoettgens, line 1154: Are these changes here mathematically correct?

src/NumField/Selmer.jl
```diff
@@ -281,16 +281,16 @@ function Hecke.is_power(a::FacElem{QQFieldElem, QQField}, p::Int)
   return true
 end
 
-function Hecke.is_power_with_root(a::FacElem{QQFieldElem, QQField}, p::Int)
+function is_power(a::FacElem{QQFieldElem, QQField}, p::Int)
```
lgoettgens, line 284: This signature is now equivalent to the one right above. The above function should probably be removed

src/EllCrv/Finite.jl
thofma: The output is randomized, and the random generator will be different for the new finite field. I will probably double check with tomorrow.

thofma: Yes, it is correct.

src/NumField/Selmer.jl
thofma: Removed it

PR #1339, opened 2023-12-30
Use julia 1.10 as default in CI

lgoettgens:
Merging this needs also an update to the list of required CI jobs.


.github/workflows/CILong.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       fail-fast: false
       matrix:
         julia-version:
-          - 1.9
+          - 1.10
```
lgoettgens, line 25:
```suggestion
          - '1.10'
```

.github/workflows/CILong.yml
```diff
@@ -62,7 +62,7 @@ jobs:
...
```
lgoettgens, line 65:
```suggestion
          - '1.10'
```

lgoettgens:
This now looks good. Somehow, the 1.10 was parsed as 1.1 before.

thofma:
Thanks! I will merge once green. Afterwards I can adjust the "required" thingies.

PR #1342, opened 2024-01-03
Adjust to Nemo changes: 0 is not squarefree any more

joschmitt:
In https://github.com/Nemocas/Nemo.jl/pull/1596 `is_squarefree` was set to return `false` for 0. The Hecke tests wanted `true`.
(Which makes me wonder whether these tests should be in Nemo.)


thofma:
Yes, maybe we actually don't need them anymore. But let's postpone these questions.

lgoettgens:
They should probably be moved to Nemo instead.

PR #1343, opened 2024-01-03
Collect exports in a file

joschmitt:
... sorted with `LC_COLLATE=C sort < exports.jl > dummy` on Linux to make Z > a.


joschmitt:
There are a few functions where I commented out the `export`:
* functions and types defined in `src/Misc/arf.jl` which is not included
* `number_of_small_groups` and `small_groups_limit` which had commented-out exports in `src/Grp/SmallGroups.jl` with the comment that the exports cause errors in Oscar. I don't know what these errors are; both functions do not exist in Oscar.

joschmitt:
And here is a list of functions (!) which are capitalized / camel case
`Amodule, CPS_dvev_complex, CPS_dvev_real, CPS_height_bounds, CPS_non_archimedean, EquationOrder, HyperellipticCurve, MaximalOrder, Order, Qp_soluble, R_soluble, UnitGroup`

PR #1344, opened 2024-01-04
Unify eigenvalue interface

joschmitt:
This is the second half of my proposal to unify the eigenvalue interface throughout OSCAR (https://github.com/oscar-system/Oscar.jl/issues/2017). The other half is https://github.com/Nemocas/Nemo.jl/pull/1611 .

I detail the changes in the Nemo PR. Here, the alias `spectrum` is introduced (mostly for backwards compatibility).


lgoettgens:
The eigenvalue functionality here does not need nothing from Hecke. Just copying it over to Nemo / AA enables eigenvalue computation in arbitrary matrix spaces without loading Hecke. It seems weird to me to just have this available for CC and RR there, but not e.g. QQ. Thus, I would like this generic implementation to be moved to AbstractAlgebra.

joschmitt:
Sounds reasonable. Once we agreed on the signature and return type, I can move it.

thofma:
Would be good to have these functions somewhere, where they can actually be tested (spoiler: it is not AbstractAlgebra).

joschmitt:
The `eigenvalues` and `eigenspaces` are moved to Nemo. I will update this PR once the Nemo side is settled.

lgoettgens:
@joschmitt I am trying to get the compatibility with Nemo 0.40 going in #1362. To mitigate possible errors there due to eigenvalue things, I would like to cherry pick the commits from this PR to there to adapt to the eigenvalue changes in Nemo. As long as I don't have any issues there due to eigenvalues, I can delay the cherry-picking.

Since I think you wanted to make some changes to this PR, I wanted to ping you a bit in advance.

joschmitt:
> @joschmitt I am trying to get the compatibility with Nemo 0.40 going in #1362. To mitigate possible errors there due to eigenvalue things, I would like to cherry pick the commits from this PR to there to adapt to the eigenvalue changes in Nemo. As long as I don't have any issues there due to eigenvalues, I can delay the cherry-picking.
> 
> Since I think you wanted to make some changes to this PR, I wanted to ping you a bit in advance.

This pull request is completely out of date. How urgent is this? I was planning to adjust it in the afternoon.

joschmitt:
I updated this now. You can just cherry-pick the last commit.

PR #1348, opened 2024-01-12
New interface for enumeration of definite integer genera

StevellM:
Implementation of a new interface for enumeration of definite `ZZGenus`. It comes together with some performance. Previously we were using the generic methods for enumeration of `QuadGenus`, but it could be made slightly better by doing everything with regular integers and rational numbers.

The new interface gives the choice between two algorithms, namely `:orbit` and `:random`, with a `:default` choice depending on the rank of the genus, and the prime we choose for Kneser's algorithm.

The main reference is of Schulze-Pillot, and we cover the odd and even cases, as well as the possibility of applying neighbours method above `p = 2`. 

Among the features, we have:
- the use of the mass (which was already available) for the termination of the algorithm;
- `stop_after` to break the process if we have not found new isometry classes after some iterations;
- `max` (which was already available) to return only a certain number of isometry classes;
- `save_partial` to store locally any new isometry classes computed (so we do not lose all partial results after killing a process by hand)

The interface offers the possibility to enumerate a genus from the symbol, a representative, or a list of isometry classes of a given (spinor) genus (for instance if we start again after an incomplete enumeration).


src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,474 @@
...
+###############################################################################
+#
+#  Neighbours
+#
+###############################################################################
+
+function neighbour(L::ZZLat, v::ZZMatrix, p::ZZRingElem)
+  M = gram_matrix(L)*v
+  r, K = left_kernel(matrix(GF(p), rank(L), 1, collect(M)))
+  LL = lattice_in_same_ambient_space(L, lift(view(K, 1:r, :))*basis_matrix(L)) + lattice_in_same_ambient_space(L, 1//p*(transpose(v)*basis_matrix(L))) + p*L
```
StevellM, line 16:
```suggestion
  _B = matrix(QQ, r, ncols(K), ZZRingElem[lift(ZZ, k) for k in view(K, 1:r, :)])
  LL = lattice_in_same_ambient_space(L, _B*basis_matrix(L)) + lattice_in_same_ambient_space(L, 1//p*(transpose(v)*basis_matrix(L))) + p*L
```

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,474 @@
...
+
+  result = typeof(L)[]
+
+  @vprintln :GenRep 3 "$(length(LO)) orbits of lines to try"
+
+  for x in LO
+    vain[] > stop_after && break
+    w = ZZRingElem[lift(ZZ, k) for k in x]
+    a = Tp(abelian_group(Tp)(w))
+    if iszero(quadratic_product(a))
```
StevellM, line 87:
```suggestion
    if !iszero(quadratic_product(a))
```

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,474 @@
...
+      vain[] += 1
+      continue
+    end
+    if has_preimage(jp, a)[1]
+      vain[] += 1
+      continue
+    end
+
```
StevellM, line 95:
```suggestion

w=matrix(ZZ, length(w), 1,w)
```

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,474 @@
...
+  result = typeof(L)[]
+
+  maxlines = isnothing(rand_neigh) ? min(100, length(P)) : min(rand_neigh, length(P))
+
+  @vprintln :GenRep 3 "Try $(maxlines) random lines"
+
+  for i in 1:maxlines
+    vain[] > stop_after && break
+    w = ZZRingElem[lift(QQ, k) for k in rand(P)]
```
StevellM, line 158:
```suggestion
    w = ZZRingElem[lift(ZZ, k) for k in rand(P)]
```

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,474 @@
...
+    a = Tp(abelian_group(Tp)(w))
+    if iszero(quadratic_product(a))
```
StevellM, line 160:
```suggestion
    if !iszero(quadratic_product(a))
```

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,474 @@
...
+      vain[] += 1
+      continue
+    end
+    if has_preimage(jp, a)[1]
+      vain[] += 1
+      continue
+    end
+
```
StevellM, line 168:
```suggestion
w=matrix(ZZ, length(w), 1,w)

```

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,474 @@
...
+    end
+  end
+  resize!(A, count)::typeof(A)
+end
+
+# Could complement with other invariants at some point if we want
+function default_func(L::ZZLat)
+  m = minimum(L)
+  genus = root_lattice_recognition(L)
+  sort!(rlr; lt=(a,b) -> a[1] < b[1] || a[1] == b[1] && a[2] <= b[2])
```
StevellM, line 229:
```suggestion
  
```

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,474 @@
...
+      for M in N
+        push!(tbv, true)
+        push!(res, M)
+      end
+      use_mass && is_zero(missing_mass[]) && break
+      if use_mass
+        @v_do :ZZLatWithIsom 1 perc = Float64(missing_mass[]//_mass) * 100
+        @vprintln :ZZLatWithIsom 1 "Lattices: $(length(res)), Target mass: $(_mass). missing: $(missing_mass[]) ($(perc)%)"
+      else
+        @vprintln :ZZLatWithIsom 1 "Lattices: $(length(res))"
```
StevellM, line 343:
```suggestion
        @v_do :GenRep 1 perc = Float64(missing_mass[]//_mass) * 100
        @vprintln :GenRep 1 "Lattices: $(length(res)), Target mass: $(_mass). missing: $(missing_mass[]) ($(perc)%)"
      else
        @vprintln :GenRep 1 "Lattices: $(length(res))"
```

StevellM (commented):
For my future self.

StevellM:
I am currently using this implementation together with Magma, and it is efficient. Remain to benchmark with a pure Hecke setting.
- [ ] Add documentation for exported functions;
- [ ] Finish the benchmark to choose best algorithm;
- [ ] Implement some fast non-trivial tests;
- [ ] Test all the extra features (vain iteration, save/load, filtering partial list, ...)

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,477 @@
...
+  a = K(div(a, p))
+  v = 2*form*w
+  vK = map_entries(b -> K(b)//a, v)
+  vK = reduce(vcat, dense_matrix_type(K)[identity_matrix(K, 1), vK])
+  _, L = left_kernel(vK)
+  @hassert :GenRep 3 !iszero(view(L, :, 1))
+  j = findfirst(j -> !iszero(L[j, 1]), 1:nrows(L))
+  @hassert :GenRep 3 !isnothing(j)
+  L = map_entries(b -> b//L[j, 1], L)
+  v = lift(L[j, 2:ncols(L)])
```
StevellM, line 34:
```suggestion
  v = ZZRingElem[lift(ZZ, a) for a in L[j, 2:ncols(L)]]
```

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,477 @@
...
+    elseif vain[] > stop_after
+      break
+    end
+  end
+  return res, use_mass ? missing_mass[] : zero(QQ)
+end
+
+function enumerate_definite_genus(
+    L::ZZLat,
+    algorithm::Symbol = _choose_algorithm(_L);
```
StevellM, line 358:
```suggestion
    algorithm::Symbol = _choose_algorithm(L);
```

StevellM (commented):
Another todo

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,950 @@
...
+  count = 1
+  for x in Iterators.drop(A, 1)
+    if all(y -> !is_isometric(x, y), A[1:count])
+      A[it[1]] = x
+      count += 1
+      it = iterate(idxs, it[2])
+    end
+  end
+  resize!(A, count)::typeof(A)
+end
```
simonbrandhorst, line 514: I am having a hard time to read and understand this piece of code. Can you insert some comments as to what happens?
For instance `(iterate(idxs)::T)[2]` seems to be equivalent to `idxs[1]`?

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,950 @@
...
+- save_path -> in the case "save_partial" is true, where to save the lattices;
+- use_mass -> whether to use the mass formula as termination condition;
+- missing_mass -> if "use_mass" and "distinct" are true, and the partial mass of
+                  "known" is known, mention what is the part of the mass missing;
+- vain -> count the number of vain iterations without finding a new lattice in
+          the given spinor genus;
+- stop_after -> the algorithm stops if we have done that amount of iterations
+                without finding a new isometry class;
+- max -> maximum number of lattices to be found.
+======#
```
simonbrandhorst, line 559: remove this? After all the function is documented properly?

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,950 @@
...
+same (spinor) genus, compute new representatives of isometry classes of
+lattices which are not represented by the lattices in `known` using Kneser's
+neighbour algorithm. The output consists of a copy of `known` together with new
+lattices (if any found), and a rational number which corresponds to the portion of
+the mass of the (spinor) genus which is missing.
+
+The second argument gives the choice to which algorithm to use for the enumeration.
+We currently support two algorithms:
+- `:random` which finds new isometry classes by constructing neighbours from random isotropic lines;
+- `:orbit` which makes a smart search by classifying orbits of isotropic lines before constructing neighbours.
```
simonbrandhorst, line 583:
```suggestion
- `:orbit` which computes orbits of isotropic lines before constructing neighbours.
```

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,950 @@
...
+If `algorithm = :default`, the function chooses the most appropriate algorithm
+depending on the rank and determinant of the genus to be enumerated.
+
+The possible extra arguments are as follows:
+- `rand_neigh` -> for random enumeration, how many random neighbours are computed at each iteration;
+- `distinct` -> whether the lattices in `known` are known to be pairwise non-isometric;
+- `invariant_func` -> a function to compute isometry invariants in order to check (faster) whether a new neighbour is already represented by a known isometry class;
```
simonbrandhorst, line 590:
```suggestion
- `invariant_func` -> a function to compute isometry invariants in order to avoid unnecessary isometry tests;
```

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,950 @@
...
+- `save_partial` -> whether one wants to save iteratively new isometry classes (for instance for large genera);
+- `save_path` -> a path to a folder where to save new lattices in the case where `save_partial` is true;
+- `use_mass` -> whether to use the mass formula as termination condition;
+- `missing_mass` -> if `use_mass` and `distinct` are true, and the partial mass of `known` is known, mention what is the part of the mass which is missing;
```
simonbrandhorst, line 594: what do you mean by `mention`?

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,950 @@
...
+- `stop_after` -> the algorithm stops if we have done that amount of iterations without finding a new isometry class;
```
simonbrandhorst, line 595:
```suggestion
- `stop_after` -> the algorithm stops after the specified amount of vain iterations without finding a new isometry class is reached;
```

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,950 @@
...
+- `max` -> the algorithm stops after finding `max` new isometry classes.
+
+If `distinct == false`, the function first compare all the lattices in `known`
```
simonbrandhorst, line 598:
```suggestion
If `distinct == false`, the function first compares all the lattices in `known`
```

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,950 @@
...
+to only keep one representative for each isometry class represented.
+
+If `save_partial == true`, the lattices are stored in a compact way in a `.jl`
```
simonbrandhorst, line 601: Is the file julia code? Only in that case the `.jl` should be used.

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,950 @@
...
+  edg = ZZLat[]
+  LL = _to_number_field_lattice(L)
+  p = _smallest_norm_good_prime(LL)
+  spinor_genera = ZZLat[_to_ZLat(N; K=QQ) for N in spinor_genera_in_genus(LL, typeof(p)[p])]
+  @vprintln :ZGenRep 1 "$(length(spinor_genera)) spinor genera to enumerate"
+
+  for M in spinor_genera
+    vain = Ref{Int}(0)
+    if use_mass
+      _missing_mass = mass(M)//length(spinor_genera)
```
simonbrandhorst, line 788: Please comment what is done here. I am slightly sceptic.

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,950 @@
...
+                                                use_mass,
+                                                stop_after,
+                                                max)
+end
+
+###############################################################################
+#
+#  Compact serialization
+#
+###############################################################################
```
simonbrandhorst, line 897: @thofma  could you give the parsing a short look? I am not really an expert.

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,950 @@
...
+#   representatives for `max` isometry classes in the genus of `L`, in the
+#   outer scope.
+function _neighbours_definite_orbit(L::ZZLat, p::ZZRingElem; callback::Function,
+                                                             inv_dict::Dict,
+                                                             _invariants::Function,
+                                                             save_partial::Bool = false,
+                                                             save_path::Union{Nothing, String} = nothing,
+                                                             use_mass::Bool = true,
+                                                             missing_mass::Base.RefValue{QQFieldElem},
+                                                             vain::Base.RefValue{Int} = Ref{Int}(0),
```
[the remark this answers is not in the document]
StevellM: `vain`, like `missing_mass` is a variable in the outer scope that we pass between functions. So using a reference here make it simpler to modify it locally.

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,950 @@
...
+  # the same `O(L)`-orbit (because they give rise to isometric `p`-neighbours of
+  # `L`).
+  G = automorphism_group_generators(L; ambient_representation=false)
+  if flag
+    gensp = dense_matrix_type(K)[map_entries(K, L0toL*g*LtoL0) for g in G]
+  else
+    gensp = dense_matrix_type(K)[map_entries(K, g) for g in G]
+  end
+  any(isone, gensp) || push!(gensp, identity_matrix(K, rank(L))) # To avoid empty lists for `line_orbits`
+  filter!(m -> isone(m) || !is_diagonal(m), gensp)
```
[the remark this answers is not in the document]
StevellM: Indeed, this was a mistake, it is fixed now.

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,950 @@
...
+# Could complement with other invariants at some point if we want.
+# We use it to compute invariants of lattices in order to make
+# comparison faster (two lattices with distinct invariants cannot
+# be isometric). In such a way, `is_isometric` is triggered at the last
+# moment (since it can be expensive).
+function default_func(L::ZZLat)
+  m = minimum(L)
+  rlr, _ = root_lattice_recognition(L)
+  kn = kissing_number(L)::Int
+  ago = automorphism_group_order(L)::ZZRingElem
```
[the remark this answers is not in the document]
StevellM: For now it is not much of a bottleneck. For this PR I guess we could keep it as is and see for further enhancement with adding new type of invariants.

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,1030 @@
...
+  p == 2 && return false
+
+  # For the odd case, one of them has an isotropic vector but
+  # not the other one
+  sym[end][3] == jacobi_symbol(-1, p) && return false
+  return true
+end
+
+@doc raw"""
+    _is_isotropic(G::ZZGenus, p::ZZRingElem) -> Bool
```
simonbrandhorst, line 499: We have this functionality already.

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,1030 @@
...
+@doc raw"""
+    smallest_kneser_prime(G::ZZGenus) -> ZZRingElem
+
+Given a genus $G$ of integral $\mathbb{Z}$-lattice, return the
+smallest Kneser prime $p$ for $G$.
+
+A prime number $p$ is called Kneser for $G$ if:
+- The lattices in $G$ are isotropic locally at $p$;
+- The lattices in $G$ are maximal bilinear integral locally at $p$;
+- The prime $p$ is not an improper spinor generator of $G$.
```
simonbrandhorst, line 531: Please double check this. Maybe being a generator is too strong?

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,1030 @@
...
+add_assertion_scope(:ZGenRep)
+
+###############################################################################
+#
+#  Neighbours
+#
+###############################################################################
+
+@doc raw"""
+    prime_dual(L::ZZLat, v::ZZMatrix, p::ZZRingElem) -> ZZLat
```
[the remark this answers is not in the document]
StevellM: No, I "created" the name from an alternative definition (this lattice is the dual of another lattice at the end). Schulze-Pillot does not seem to give it an actual name neither.

src/QuadForm/Quad/ZGenusRep.jl
StevellM: As far as I know, we have it for quadratic spaces, but I want to avoid constructing a space where I could only check it from the symbol. So it is the same code at the end.

src/QuadForm/Quad/ZGenusRep.jl
StevellM: I wanted to use the function, but it is not defined everywhere (for example at the bad primes). So maybe I should skip the bad primes, and then only check that `f(p) != 0` ?

simonbrandhorst: yes

simonbrandhorst: well in some cases the bad primes are interesting for us ... but then I am not sure how the spinor genus behaves. 
In case there is a unique one it is no problem anyways.

src/QuadForm/Quad/ZGenusRep.jl
simonbrandhorst: we have `is_isotropic(rational_isometry_class(::ZZGenus))`
please use that. no space is created

StevellM:
For the cases of rank 3 and 4, if we allow the prime `2` for odd genera, or if we also allow bad primes, then we need to be careful. There might be cases for which there exists no admissible vector (it is a bit more rigid than in the good prime case). 
So, we have the choice between doing a case-by-case analysis (on the lattices themselves), and imposing further conditions on the choice of the "Kneser" prime so that we know that each lattice in the genus admits at least one admissible prime. 
I have chosen the last condition because for very low ranks like 3 and 4, we can tolerate to choose a bigger prime for the enumeration (the class numbers are usually small, and we can do orbits of lines with relatively large primes)

simonbrandhorst:
Can you give a proof/reference?

StevellM:
I do not have a reference, but I could provide a proof/argument of what could go wrong in each of the corner cases I have mentioned. Not sure if that fits well in the code directly. 

The drops of rank I mention in the comments come from the possibility of having vector that are `p`-divisible when `p` divides the determinant, or when `p=2` and the lattices are odd (then the radical quadratic of the quotient L/pL might be non-trivial of rank 1, so the regular part loses rank). 
The isotropy solutions are related to Chevalley's theorem (which basically says that if the quotient `L/pL` has (torsion) rank at least 3 then we are good). If you prefer we could discuss it in person.

StevellM:
After investigation, the problem I tried to solve previously was not what I thought: it was just a missing line in the code of `rational_isometry_class`. With that fixed, all the enumerations work as long as we avoid the prime 2 for odd genera.

It was decided that for now we will leave things as it is, and the case of prime 2 for odd genera will be treated in another PR since it would need some more care. Thought I keep for now all the existing infrastructure which might be reuse in the future.

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,1023 @@
...
+
+@doc raw"""
+    is_maximal_integral_bilinear(G::ZZGenus, p::ZZRingElem) -> Bool
+
+Given a genus ``G`` of integral ``\\mathbb{Z}``-lattices and a prime number
+``p``, return whether the lattice ``L_p`` is maximal integral in
+``\\mathbb{Q}L_p``, for the bilinear form on ``L_p``.
+
+Note that this does not depend on a choice of a lattice ``L`` in ``G``,
+and this property can be directly check from the genus symbol.
```
simonbrandhorst, line 466:
```suggestion
and this property can be checked directly from the genus symbol.
```

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -0,0 +1,1023 @@
...
+###############################################################################
+#
+#  User interface
+#
+###############################################################################
+
+@doc raw"""
+    is_maximal_integral_bilinear(G::ZZGenus, p::ZZRingElem) -> Bool
+
+Given a genus ``G`` of integral ``\\mathbb{Z}``-lattices and a prime number
```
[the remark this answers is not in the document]
StevellM: Yes I misread the Documenter documentation. I have fixed that.

simonbrandhorst:
This is a great improvement of our genus enumeration. 
Thanks lot!

StevellM:
@thofma Any idea what is going on with CI? It seems that there are two checks which do not want to pass.

thofma:
Seems to be about the new "book tests" added in Oscar. There seems to be a lot of noise and one "real" error at https://github.com/thofma/Hecke.jl/actions/runs/8673739125/job/23826928025?pr=1348#step:7:8421. @benlorenz is this expected to fail?

simonbrandhorst:
The doctest failure should be unrelated to this PR. It does not touch any code used in F-Theory as far as I can tell.

PR #1358, opened 2024-01-19
feat: add precision argument for completion_easy

thofma:
Also use it in the _p_adic_regulator function

### DIFF
--- a/src/LocalField/Conjugates.jl
+++ b/src/LocalField/Conjugates.jl
@@ -163,7 +163,7 @@ end
 #  flat = true/ false: return (Re, Im) or the complex number
 #TODO: not sure how this would work in the ramified, not-normal case.
 @doc raw"""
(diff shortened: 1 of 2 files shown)


src/LocalField/Conjugates.jl
```diff
@@ -163,7 +163,7 @@ end
 #  flat = true/ false: return (Re, Im) or the complex number
 #TODO: not sure how this would work in the ramified, not-normal case.
 @doc raw"""
-    conjugates(a::nf_elem, C::qAdicConj, n::Int = 10; flat::Bool = false, all:Bool = true) -> []
+    conjugates(a::nf_elem, C::qAdicConj, n::Int = 11; flat::Bool = false, all:Bool = true) -> []
```
thofma, line 166:
```suggestion
    conjugates(a::nf_elem, C::qAdicConj, n::Int = 10; flat::Bool = false, all:Bool = true) -> []
```

PR #1362, opened 2024-01-23
Adjust to Nemo 0.40

lgoettgens:
Includes and thus closes https://github.com/thofma/Hecke.jl/pull/1361.
Includes and thus closes https://github.com/thofma/Hecke.jl/pull/1356.
Includes and thus closes https://github.com/thofma/Hecke.jl/pull/1344.


lgoettgens:
This now contains everything I already knew that has to be adapted. Let's see what else fails...

lgoettgens:
The next error is in https://github.com/thofma/Hecke.jl/blob/26bd0d41246cd5416b9e4c75922e97fed69217c4/src/NumFieldOrd/NfOrd/Ideal/Prime.jl#L1330 due to https://github.com/Nemocas/Nemo.jl/pull/1620. I have no idea how to fix that, can you have a look @thofma?

thofma:
Yes, I will adjust it

PR #1350, opened 2024-01-15
Fix printing of SRow in compound structures

lgoettgens:
before:
```julia
julia> (sparse_row(QQ), sparse_row(QQ))
(Sparse row with positions Int64[] and values QQFieldElem[]
, Sparse row with positions Int64[] and values QQFieldElem[]
)

julia> [sparse_row(QQ), sparse_row(QQ)]
2-element Vector{SRow{QQFieldElem, Vector{QQFieldElem}}}:
 Sparse row with positions Int64[] and values QQFieldElem[]

 Sparse row with positions Int64[] and values QQFieldElem[]


```

after:
```julia
julia> (sparse_row(QQ), sparse_row(QQ))
(Sparse row with positions Int64[] and values QQFieldElem[], Sparse row with positions Int64[] and values QQFieldElem[])

julia> [sparse_row(QQ), sparse_row(QQ)]
2-element Vector{SRow{QQFieldElem, Vector{QQFieldElem}}}:
 Sparse row with positions Int64[] and values QQFieldElem[]
 Sparse row with positions Int64[] and values QQFieldElem[]

```

PR #1363, opened 2024-01-24
Remove some trailing newlines from show methods

lgoettgens:
This makes the types print nicer in e.g. a vector (similar to https://github.com/thofma/Hecke.jl/pull/1350)


fingolfin (approved):
Thanks!

PR #1367, opened 2024-01-26
renaming: `haspreimage`/`has_preimage` -> `has_preimage_with_preimage`

StevellM:
As mentioned in https://hackmd.io/nRnyfrSxTVe5CzXidB1cBQ.

Those only appear in Hecke and Oscar, so this is the first part of the renaming.


StevellM:
@thofma Should we merge it or should we wait to make a bundle of breaking changes ? I have the Oscar counterpart ready (if I haven't missed anything)

thofma:
Yes, we will do (at least) one breaking release before Thursday and we will include this, thanks!

src/AlgAssAbsOrd/Ideal.jl
```diff
@@ -1637,7 +1637,7 @@ function _as_order_of_smaller_algebra(m::AbsAlgAssMor, O::AlgAssAbsOrd, OB::AlgA
   basis_in_A = Vector{elem_type(A)}(undef, dim(B))
   for i = 1:dim(B)
     t = elem_from_mat_row(OB, N, i)
-    b, s = haspreimage(m, elem_in_algebra(t, copy = false))
+    b, s = has_preimage_with_preimage(m, elem_in_algebra(t, copy = false))
```
fieker, line 1640: In a usage pattern like this, would it be better to just call `preimage`? the boolean here is only used to throw an error, so calling preimage directly should do the job?

StevellM: Probably yes, I did not check all changes at once but if there is a `preimage` function defined, it should do the job. Eventually, I would expect the `preimage` function to do exactly the same: call `has_preimage_with_preimage` and return an error if the boolean is `false`. I could add a `preimage` function when missing and change those calls accordingly if you think it is better.

fieker:
On Tue, Jan 30, 2024 at 12:31:26AM -0800, Stevell Muller wrote:
> @StevellM commented on this pull request.
> 
> 
> 
> > @@ -1637,7 +1637,7 @@ function _as_order_of_smaller_algebra(m::AbsAlgAssMor, O::AlgAssAbsOrd, OB::AlgA
>    basis_in_A = Vector{elem_type(A)}(undef, dim(B))
>    for i = 1:dim(B)
>      t = elem_from_mat_row(OB, N, i)
> -    b, s = haspreimage(m, elem_in_algebra(t, copy = false))
> +    b, s = has_preimage_with_preimage(m, elem_in_algebra(t, copy = false))
> 
> For what I have read, it seems that `AbsAlgAssMor` are constructed with a `preimage` function, and it does not seem that there is any check done. Though we have `has_preimage_with_preimage` defined, meaning that they might not always be surjective morphisms. I am not sure whether `preimage` in this context will return an error, or if it does, I am not sure whether this is going to be a helpful error. Any idea on how we should proceed ? 
> 
I'd expect preimage to be essentially calling has_... and throw an error
if this fails. (if this is a helpful error: no idea)

I don't think the having a preimage function is (or should be) linked to
the map being surjective

StevellM:
I agree with that. Should this be a general pattern in your opinion ? Sometimes we only have a `preimage` function (because there is always a preimage ?): should we also have a `has_preimage_with_preimage` by default and fallback to this one for `preimage`, even in those cases ?

thofma:
It is too much of a mess to clean up for the different maps at this point. Let's wait for the deadline and look at this once things have settled down a bit.

PR #1393, opened 2024-02-05
Adapt to name printing changes in AA

lgoettgens:
Refactors everything about name printing to only use the interface discussed in https://github.com/Nemocas/AbstractAlgebra.jl/pull/1594 (and no longer any internals)


lgoettgens:
This needs https://github.com/oscar-system/Oscar.jl/pull/3328 to be merged first.

PR #1396, opened 2024-02-09
Redirect linear solving

joschmitt:
Use the `Solve.` interface. Needs AbstractAlgebra and Nemo update.


src/Hecke.jl
```diff
@@ -111,6 +111,18 @@ import Nemo: acb_struct, Ring, Group, Field, zzModRing, zzModRingElem, arf_struc
...
+const can_solve_with_solution = AbstractAlgebra.Solve.can_solve_with_solution
+
+function kernel(M::MatElem; side::Symbol = :right)
+  return AbstractAlgebra.Solve.kernel(M, side = side)
+end
+
+function kernel(R::Ring, M::MatElem; side::Symbol = :right)
+  return AbstractAlgebra.Solve.kernel(R, M, side = side)
+end
+
```
joschmitt, line 125: This is the "non-trivial" part of this pull request. Is this fine or should I also make dummy functions for `solve` etc.? (It's only temporarily anyways.)

thofma: Yes, this is fine

src/Misc/jordan.jl
```diff
@@ -812,12 +812,12 @@ of a key and a value, then `v` is the eigenspace of `L[i]` w.r.t. the eigenvalue
 If side is `:right`, the right eigenspaces are computed, if it is `:left` then the
 left eigenspaces are computed.
 """
-function common_eigenspaces(L::Vector{<: MatElem{T}}; side::Symbol = :right) where T <: FieldElem
-  eigs = [ eigenspaces(M, side = side) for M in L ]
+function common_eigenspaces(L::Vector{<: MatElem{T}}; side::Symbol = :left) where T <: FieldElem
+  eigs = [ eigenspaces(M; side = :left) for M in L ]
```
joschmitt, line 816:
```suggestion
  eigs = [ eigenspaces(M; side = side) for M in L ]
```
I don't think anything should be changed here?

PR #1398, opened 2024-02-13
Remove old deprecations

lgoettgens:
Removes all deprecations that were present in the last (0.14.0) Oscar release to reduce the number of deprecations for 1.0.
Removing the other ones as well would be inconvenient for users of Oscar since they miss the deprecation stage of those type names.

x-ref: https://github.com/oscar-system/Oscar.jl/issues/3356


lgoettgens:
@thofma can you please mark this as "breaking"?

lgoettgens:
Most parts of https://github.com/thofma/Hecke.jl/blob/lg/deprecations-Hecke/src/Aliases.jl should probably go as well, even though they are not marked as deprecated currently, but only as `@alias`. What do you think?

cc @benlorenz

lgoettgens:
The errors are about usages of `is_equivalent` that are no longer possible due to the removal of `@deprecate is_equivalent is_isometric`. However, there are methods for `is_isometric`:
https://github.com/thofma/Hecke.jl/blob/a6b4f9ca2bbe26fd18d859b80a40f98e8a70177e/src/QuadForm/QuadBin.jl#L409

Can someone familiar with the code have a look, please?

thofma:
Can you elaborate? The function should be called `is_equivalent` and and has a `is_isometric` "alias" a few lines above.

thofma:
Yes, the method for `QuadBin` types should stay.

lgoettgens:
Most parts of [lg/deprecations-Hecke/src/Aliases.jl](https://github.com/thofma/Hecke.jl/blob/lg/deprecations-Hecke/src/Aliases.jl?rgh-link-date=2024-02-13T11%3A08%3A49Z) should probably go as well, even though they are not marked as deprecated currently, but only as `@alias`. What do you think?

cc @benlorenz

thofma:
I'd rather turn them first into proper deprecations.

lgoettgens:
> I'd rather turn them first into proper deprecations.

OK, I can prepare a PR for this tomorrow

PR #1407, opened 2024-02-18
fix: multivariate gcd over number fields

thofma:
- handle the case of non-nice defining polynomial better


fieker (approved):
well spotted

PR #1409, opened 2024-02-19
Add a few more ZZIdl methods

fingolfin:


src/NumField/QQ.jl
```diff
@@ -152,6 +152,12 @@ isone(I::ZZIdl) = isone(I.gen)
 iszero(I::ZZIdl) = iszero(gen(I))
 is_maximal(I::ZZIdl) = is_prime(gen(I))
 is_prime(I::ZZIdl) = is_zero(I) || is_maximal(I)
+is_primary(I::ZZIdl) = is_zero(I) || is_prime_power_with_data(gen(I))[1]
+
+is_subset(I::ZZIdl, J::ZZIdl) = is_divisible_by(gen(J), gen(I))
+
+radical(I::ZZIdl) = ideal(ZZ, radical(gen(I)))
+primary_decomposition(I::ZZIdl) = [ (ideal(ZZ, p^k), ideal(ZZ, p)) for (p,k) in factor(gen(I)) ]
```
thofma, line 160: I think this will fail for the zero ideal?

fingolfin: Indeed, fixed both `radical` and `primary_decomposition` for the zero ideal. Also fixed `intersect` for more or fewer than 2 arguments, and added test cases triggering all that.

PR #1410, opened 2024-02-20
fix: purge FakeFmpqMat from user interface

thofma:


Project.toml
```diff
@@ -28,7 +28,6 @@ GAPExt = "GAP"
 PolymakeExt = "Polymake"
 
 [compat]
-AbstractAlgebra = "^0.39.0"
```
lgoettgens, line 31: why remove the AA compat? that will block AutoMerge in the general registry

thofma: Oops, thanks for catching it

PR #1416, opened 2024-02-23
feat: improve radical code

thofma:
- move code to radical.jl and disentangle
- avoid using FqField in the prime field case


src/AlgAss/radical.jl
```diff
@@ -0,0 +1,438 @@
...
+    if d == 0
+      return elem_type(A)[]
+    end
+    C = transpose(B)*C
+  end
+
+  return elem_type(A)[ elem_from_mat_row(A, C, i) for i = 1:nrows(C) ]
+end
+
+#function _radical(A::AbstractAssociativeAlgebra{T}) where { T <: Union{ fpFieldElem, FpFieldElem } }
```
StevellM, line 242: Is this necessary to keep this commented block ? It seems to be somehow similar to what is written above.

src/ModAlgAss/Lattices/Morphisms.jl
```diff
@@ -254,6 +254,7 @@ end
 ################################################################################
 
 function _is_isomorphic_with_isomorphism_same_ambient_module(L::ModAlgAssLat, M::ModAlgAssLat, with_iso::Type{Val{T}} = Val{true}) where {T}
+  @info "Computing hom space as ideal"
```
StevellM, line 257: Maybe here and line 271 it would be better a `@vprintln`, or is `@info` kept on purpose for warning the users ?

thofma:
Sorry, I noticed that I still have some debug statements left and some old code. I have to overhaul this a bit!

PR #1413, opened 2024-02-21
Remove outdated linear solving code

joschmitt:

PR #1417, opened 2024-02-26
Adjust to Nemo changes

joschmitt:
I expected this to be more complicated... This should make Hecke ready for Nemo 0.43.1.


fingolfin:
There are doctest failures, though, due to some doctest code now resulting in errors.

joschmitt:
> There are doctest failures, though, due to some doctest code now resulting in errors.

Yes, all the failures are expected, this needs Nemo 0.43.1 now.
(The "cleanest" thing would have been to first release Nemo and then use the `SolveCtx` stuff in Hecke, but it is too late for this.)

joschmitt:
Closing this now. Can/needs to be done in one go when we get #1413 back in.

thofma:
When we revert the revert, we can cherry-pick it.

PR #1429, opened 2024-03-15
More tests for Integer.jl 

Sreevatsa2k:
Added ten test codes related to support and radical function, going to add tests for io functions as well


thofma:
The tests `string(...) == "..."` are not very robust and should be avoided.

PR #1438, opened 2024-03-29
feat: make valuation ring elements of local fields more robust

thofma:
- do not allow arbitrary nonsensical coercions
- provide absolute_coordinates(Qp, ...) with predetermined parent


thofma:
@fieker: The `absolute_coordinates(Qp, x)` is not optimal, since it mixes precision and parent. What I really want is `absolute_coordinates(x; precision = p)`, but absent a global $\mathbf{Q}_p$ this is not a useful interface.

PR #1444, opened 2024-04-03
fix: use more any_order

thofma:


thofma:
wait, this is wrong

thofma:
OK, I had to untangle the maximal orders of relative simple extensions, but should be fine now. There are more broken things, but at least the new test is working.

fieker (approved):

PR #1446, opened 2024-04-04
Simplify some FinGenAbGroup function by using elementary_divisors

fingolfin:
Also fix `is_cyclic` for trivial `FinGenAbGroup`


thofma:
Most of those should just call `elementary_divisors` I think. We probably did not have it back then.

fingolfin:
I can use `elementary_divisors` if you prefer, but that will then lead to additional allocations as it does `return copy(G.snf)`

thofma:
I think it is still worthwhile. The less we use `.snf` the better.

fingolfin:
Is it guaranteed that there is no 1 in the vector returned by `elementary_divisors`? Its docstring doesn't really say so directly (well, I guess it must be implicitly true for the "uniqueness" claim to hold...)

fingolfin:
Update to use `elementary_divisors` and I also tweaked the `elementary_divisors` docstring (to say that 1 does not occur, and to document that 0 may occur)

thofma:
Thanks. Yeah, it is ambiguous about the `1`. I guess for elementary divisors of matrices, one probably includes the `1`, while for modules, one does not to make it unique.

PR #1447, opened 2024-04-05
Update GAP.jl to 0.11.0 resp. to GAP 4.13.0

fingolfin:


fingolfin:
There is a failure in `Run long tests / test (1.10, x64, macOS-latest) ` but it looks unrelated?

PR #1455, opened 2024-04-17
Fix uses of `Val` inside Hecke

lgoettgens:
This is similar to https://github.com/Nemocas/AbstractAlgebra.jl/pull/1664.

`Val{T}` is a type and `Val(T)` its singleton object. This is the way it is described in the julia docs (https://docs.julialang.org/en/v1/manual/types/#%22Value-types%22).

I went through all the uses and corrected them.

For all exported functions where the parameter occurred in a docstring, I already added deprecations. If you want more deprecations, please let me know.

### DIFF
--- a/src/AlgAss/AlgGrp.jl
+++ b/src/AlgAss/AlgGrp.jl
@@ -312,23 +312,23 @@ function _merge_elts_in_gens!(left::Vector{Tuple{Int, Int}}, mid::Vector{Tuple{I
 end
 
 @doc raw"""
-    gens(A::GroupAlgebra, return_full_basis::Type{Val{T}} = Val{false})
+    gens(A::GroupAlgebra, return_full_basis::Val = Val(false))
       -> Vector{GroupAlgebraElem}
 
 Returns a subset of `basis(A)`, which generates $A$ as an algebra over
 `base_ring(A)`.
-If `return_full_basis` is set to `Val{true}`, the function also returns a
+If `return_full_basis` is set to `Val(true)`, the function also returns a
 `Vector{AbstractAssociativeAlgebraElem}` containing a full basis consisting of monomials in
 the generators and a `Vector{Vector{Tuple{Int, Int}}}` containing the
 information on how these monomials are built. E. g.: If the function returns
 `g`, `full_basis` and `v`, then we have
 `full_basis[i] = prod( g[j]^k for (j, k) in v[i] )`.
 """
-function gens(A::GroupAlgebra, return_full_basis::Type{Val{T}} = Val{false}) where T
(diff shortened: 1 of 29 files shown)


src/AlgAss/AlgGrp.jl
```diff
@@ -312,23 +312,23 @@ function _merge_elts_in_gens!(left::Vector{Tuple{Int, Int}}, mid::Vector{Tuple{I
...
-If `return_full_basis` is set to `Val{true}`, the function also returns a
+If `return_full_basis` is set to `Val(true)`, the function also returns a
 `Vector{AbstractAssociativeAlgebraElem}` containing a full basis consisting of monomials in
 the generators and a `Vector{Vector{Tuple{Int, Int}}}` containing the
 information on how these monomials are built. E. g.: If the function returns
 `g`, `full_basis` and `v`, then we have
 `full_basis[i] = prod( g[j]^k for (j, k) in v[i] )`.
 """
-function gens(A::GroupAlgebra, return_full_basis::Type{Val{T}} = Val{false}) where T
+function gens(A::GroupAlgebra, :Val{return_full_basis} = Val(false)) where return_full_basis
```
thofma, line 327:
```suggestion
function gens(A::GroupAlgebra, ::Val{return_full_basis} = Val(false)) where return_full_basis
```

lgoettgens:
I'll continue to add deprecations for everything that fails in the downstream tests

lgoettgens:
Somehow the "fake" doctests in Oscar emit deprecation warnings that make them fail. Do you have an idea why depwarns are enabled there?

thofma:
Hm, I don't remember how those fake doctests are run. @benlorenz do you remember from the top of your head?

thofma:
Maybe add the old method back in @lgoettgens?

lgoettgens:
> Deprecating exported functionality is considered breaking and should be done with a minor version bump instead of a patch-level bump.

This is just stated for versions after 1.0.0 (ref https://semver.org/spec/v2.0.0.html#spec-item-7), in the 0.y.z phase everything may change at any time (https://semver.org/spec/v2.0.0.html#spec-item-4)

To reduce friction here, let me add the deprecations as "real functions" (i.e. without a `@deprecate`) and we change this in the next minor release instead.

thofma:
Julia does not follow SemVer for 0.*: https://pkgdocs.julialang.org/v1/compatibility/#compat-pre-1.0.

lgoettgens:
> (Almost) everything that is exported by Hecke is also exported by Oscar and thus breaking for Oscar 1.0.x.

Thanks for reminding me. However, this is no longer an issue due to my last changes. The deprecations will only be introduced in Hecke 0.31.0, which will then only be made compatible on the Oscar master branch, and not the release-1.0.

PR #1456, opened 2024-04-17
move code from `Oscar.jl/experimental/GModule/Misc.jl` here

ThomasBreuer:
The idea is to add this code to Hecke.jl,
and then to remove it from Oscar.jl.

Once we agree what shall be really added here,
tests and documentation has to be added.
(I am quite sure that I have put certain parts into the wrong files, and that some functions have to be changed.
For example:
- The function `dual` belongs to Hecke.jl, but the method in question deals with objects from AbstractAlgebra.jl; should `dual` better get defined in AbstractAlgebra.jl?
- `pseudo_inv` is defined in AbstractAlgebra.jl, the method in question uses `MapFromFunc`, which belongs to Hecke.jl; what is the relation between Hecke.jl's `MapFromFunc` and AbstractAlgebra.jl's `FunctionalMap`?)


fingolfin:
Can we get this merged? Or is there any serious blocker (perhaps `dual` should be moved to `AA` -- or not. But surely we can decide that later? In the meantime having it here is already better than having it only in Oscar).

@thofma @fieker @lgoettgens any thoughts on this?

lgoettgens:
> It is not in a form to get merged

(in https://github.com/thofma/Hecke.jl/pull/1482#issuecomment-2088397584 by @thofma)

src/GrpAb/GrpAbFinGen.jl
```diff
@@ -662,6 +692,27 @@ function direct_sum(G::FinGenAbGroup...; task::Symbol = :sum, kwargs...)
   return _direct_product(:sum, G...; task = task, kwargs...)
 end
 
+@doc raw"""
+    direct_sum(G::FinGenAbGroup, H::FinGenAbGroup, V::Vector{<:Map{FinGenAbGroup, FinGenAbGroup}})
+
+For groups `G = prod G_i` and `H = prod H_i` as well as maps `V_i: G_i -> H_i`,
+build the induced map from `G -> H`.
+"""
+function direct_sum(G::FinGenAbGroup, H::FinGenAbGroup, V::Vector{<:Map{FinGenAbGroup, FinGenAbGroup}})
```
lgoettgens, line 701: This one is already added in https://github.com/thofma/Hecke.jl/pull/1488 under the name `hom_direct_sum`. Once https://github.com/thofma/Hecke.jl/pull/1488 gets merged, this should get removed here.

fingolfin:
There is no urgency in this from my POV, it just seems prudent to move it along eventually... But I think @ThomasBreuer is busy working on other things right now, and hopefully can back to this once those other things are done.

lgoettgens:
As a first step, the code could be moved from `experimental/GModule/` to https://github.com/oscar-system/Oscar.jl/blob/master/src/Misc/MoveToHecke.jl. Then we can have the discussion about the contents elsewhere, and without the pressure of doing a breaking Hecke release right after the merge

lgoettgens:
> @lgoettgens I can move the code, but I do not catch the point. If the code gets moved inside Oscar first then still there is the problem that we get warnings about duplicate names after a release of Hecke.jl.

The benefit of first moving everything in Oscar to a central place and do discussions there is that the moving itself is a trivial action that does not need to involve that many people to cooperate, but it is just a simple copy-paste that e.g. Tommy can just do when the next breaking Hecke release is planned

ThomasBreuer:
@lgoettgens O.k.

PR #1458, opened 2024-04-18
Sort out deprecations and aliases

lgoettgens:
This needs https://github.com/Nemocas/Nemo.jl/pull/1704 to be available, and the Nemo compat to be adapted accordingly.

This is an approach to sort deprecations and aliases as discussed with @fingolfin and already started in https://github.com/Nemocas/AbstractAlgebra.jl/pull/1537 and https://github.com/Nemocas/Nemo.jl/pull/1704.
There are some workarounds added to make this non-breaking for Oscar 1.0, that should get removed in the next minor release.

Things that have been changed:

- put all permanent aliases into aliases.jl, remove duplicates with AA (`ishermitian`)
- collect all deprecations in three lists in deprecations.jl (temp aliases, deprecate_bindings, and deprecations)
- no longer use path to import stuff from Nemo, instead call `@include_deprecated_bindings()` from both AA and Nemo. The initial problem both of these approaches fix is that types that are deprecated using `@deprecate_binding` are not re-exported (but should). @fingolfin and I found a macro the easiest way to execute the deprecations again in the Hecke scope (and once everything is updated in the Oscar scope as well).


lgoettgens:
Bump @thofma.
If there are any concerns with this please let me know. It would be great to get this out in a non-breaking release before the large chain of breaking releases reaches Hecke, to make the handling downstream a bit easier, and to be able to already prepare for the following breaking changes.

lgoettgens:
Thanks for merging. Could you release this as a patch release so I can already start adapting Oscar?

PR #1459, opened 2024-04-18
Avoid some uses of Meta.parse

fingolfin:


thofma:
I assume you tried those locally? (No need to add tests, just wanted to confirm.)

fingolfin:
The import/export test is exercised each time the package is loaded (and also I copied it from Oscar). 

But I failed to test `find_cache`, which I've done now, by calling `Hecke.print_cache()`, which promptly revealed an issue that I now resolved.

thofma:
Merci

PR #1460, opened 2024-04-19
Use `CacheDictType`

joschmitt:


lgoettgens:
Thanks! What I find a bit inconsistent is that you only renamed the global dict object for GroupAlgebra. Imo the other ones should be renamed as well to match the type name

src/GrpAb/Dual.jl
```diff
@@ -5,7 +5,7 @@
 #
 ######################################################################
 
-struct QmodnZ <: GrpAb
+mutable struct QmodnZ <: GrpAb
```
joschmitt, line 8: I had to make this `struct` mutable because the `WeakValueDict` wants to add a finalizer.

joschmitt:
Note to self: tests fail because of this:
```
julia> G = small_group(8, 4)
QG1 =Generic group of order 8 with multiplication table

julia> QG1 = QQ[G]
Group algebra
  of generic group of order 8 with multiplication table
  over rational field

julia> H = small_group(8, 4)
Generic group of order 8 with multiplication table

julia> QG2 = QQ[H]
Group algebra
  of generic group of order 8 with multiplication table
  over rational field

julia> G === H
false

julia> QG1 === QG2
true

julia> QG2(H[1])
ERROR: KeyError: key (1) not found
Stacktrace:
 [1] getindex
   @ ./dict.jl:498 [inlined]
 [2] GroupAlgebraElem
   @ ~/.julia/dev/Hecke/src/AlgAss/Types.jl:219 [inlined]
 [3] (::GroupAlgebra{QQFieldElem, MultTableGroup, MultTableGroupElem})(c::MultTableGroupElem)
   @ Hecke ~/.julia/dev/Hecke/src/AlgAss/Elem.jl:621
 [4] top-level scope
   @ REPL[8]:1
```
I'll look into it, but not today...

thofma:
I remember this one. Forgot what the problem is.

src/GrpAb/Dual.jl
thofma: OK

thofma:
What you see is a consequence of `==` not being `===` for these groups. Before the cache used `===`, but with the changes it uses `==`.

thofma:
#1463 should help with that

joschmitt:
> #1463 should help with that

Ah, great, thank you!

PR #1462, opened 2024-04-20
test: fix GAP and Polymake injection on CI

thofma:


thofma:
@benlorenz maybe you have an idea what I am doing wrong with the environments.

I am adding GAP and Polymake here: https://github.com/thofma/Hecke.jl/blob/32aba11fa811aa53bb3785e0c8ac3902ba50499c/.github/workflows/CILong.yml#L48
Then during the test run I try to load it here:
https://github.com/thofma/Hecke.jl/blob/master/test/runtests.jl#L53-L57
But it is not found:
https://github.com/thofma/Hecke.jl/actions/runs/8763890098/job/24053293456?pr=1462#step:7:125
Any idea what could be the problem or how to fix this?

thofma:
Thanks for the explanation. I will go with the same trick of injecting the dependencies into the Project.toml.

thofma:
Yeah, I saw it locally. Thanks for the fix

PR #1466, opened 2024-04-22
Add `coefficients` and `coeff` for `FinAbGenGroupElem`

joschmitt:
I assume these were not left out for a reason?


thofma:
Yes, they were left out for a reason, since it is not clear whether it should be
1)`Vector{ZZRingElem}` instead of `ZZMatrix`; or
2)  called `coordinates`.

PR #1470, opened 2024-04-25
Global precision p-adics and q-adics

joschmitt:
I copied the p-adic and q-adic functionality form Nemo and removed the precision of the field.

There is something wrong in the completion code which I have not been able to debug so far.


thofma:
we will pursue a different approach

PR #1476, opened 2024-04-28
scale_row! fixed

BD-00:
Scaling with zero divisors led to zeros saved in the value-array.


thofma:
Would be great if you could add a test at the bottom of https://github.com/thofma/Hecke.jl/blob/master/test/Sparse/Row.jl

BD-00:
> Would be great if you could add a test at the bottom of https://github.com/thofma/Hecke.jl/blob/master/test/Sparse/Row.jl

There's already another test for the same function. Should I add it into the same block?

PR #1480, opened 2024-04-30
Make compatible with AA 0.41 and Nemo 0.44

lgoettgens:
Includes and thus closes https://github.com/thofma/Hecke.jl/pull/1471.


lgoettgens:
Some familiar with the Hecke types should look through the newly added `base_ring_type` dispatches. I am not always sure what I am doing.

thofma:
Looks reasonable, but there is a duplicate.

PR #1486, opened 2024-05-03
fix: polynomial factorization over rational function field

thofma:
- when making factors primitive, the unit needs to be adjusted
  using the correct power.


HechtiDerLachs:
Great! Trying it out just now...

PR #1488, opened 2024-05-03
Try to sort out the homs

lgoettgens:
cf. https://github.com/oscar-system/Oscar.jl/pull/3667#issuecomment-2091079080

Alternative to and thus closes #1483.


thofma:
why is Aqua unhappy?

lgoettgens:
> why is Aqua unhappy?

The persistent task test does not work with deved dependecies

thofma:
so just ignore it?

lgoettgens:
Jup, should be fine. And even if not, we are not changing anything for oscar master here

PR #1489, opened 2024-05-03
fix: missing ring interface function for local fields

thofma:
- `canonical_unit` was missing, which made the printing
  of rational functions fail


thofma:
CC: @simonbrandhorst

PR #1491, opened 2024-05-04
Fix docs

lgoettgens:
I don't know why the docs job in https://github.com/thofma/Hecke.jl/pull/1488 was green, but https://github.com/oscar-system/Oscar.jl/pull/3667 failed due to the docs not being changed in #1488.


thofma:
Oscar docs are probably built with `warnonly = false` or `strict = true`. I will also add it here later.

PR #1495, opened 2024-05-05
Add `sparse_row_type` and `sparse_matrix_type`

fingolfin:
Also use them to improve type stability in a few places.


src/NumField/NfAbs/NormRelation/Setup.jl
```diff
@@ -344,7 +344,7 @@ function induce_action_from_subfield(N::NormRelation, i, s, FB, cache)
   S = FB.ideals
   ZK = order(S[1])
 
-  z = SMat{ZZRingElem}[zero_matrix(SMat, FlintZZ, 0, length(S)) for i in 1:degree(field(N))]
+  z = [zero_matrix(SMat, FlintZZ, 0, length(S)) for i in 1:degree(field(N))]
```
thofma, line 347:
```suggestion
  z = sparse_matrix_type(ZZ)[zero_matrix(SMat, FlintZZ, 0, length(S)) for i in 1:degree(field(N))]
```
Lets keep helping the compiler.

src/Sparse/Matrix.jl
```diff
@@ -23,6 +23,37 @@ parent(A::SMat) = SMatSpace(base_ring(A), A.r, A.c)
 
 base_ring(A::SMat{T}) where {T} = A.base_ring::parent_type(T)
 
+base_ring_type(::Type{SMat{T}}) where {T} = parent_type(T)
```
lgoettgens, line 26:
```suggestion
base_ring_type(::Type{<:SMat{T}}) where {T} = parent_type(T)
```

src/Sparse/ZZRow.jl
```diff
@@ -249,7 +251,7 @@ function sparse_row(R::ZZRing, A::Vector{Tuple{Int64, ZZRingElem}}; sort::Bool =
   return SRow(R, l, a)
 end
 
-function sparse_row(R::ZZRing, pos::Vector{Int64}, val::AbstractVector{ZZRingElem}; sort::Bool = true)
+function sparse_row(R::ZZRing, pos::Vector{Int64}, val::AbstractVector{T}; sort::Bool = true) where T
```
lgoettgens, line 254: This needs to stay `ZZRingElem`, as it uses `ZZRingElem_Array()` in the body!

lgoettgens (changes requested):
Both things above will break something and should be adjusted

src/Sparse/ZZRow.jl
lgoettgens: Maybe false alarm for this: the `push!(::ZZRingElem_Array, -)` takes care of other input types, at least I found a version for `Int`. So this change makes `sparse_row(ZZ, ::Vector{Int}, ::Vector{Int})` work while moving errors for unsupported types further down.

fingolfin: Yes that was the intention, to get feature parity with the generic `sparse_row(R::Ring, J::Vector{Int}, V::Vector{T})` which explicitly coerces into the ring:
```julia
function sparse_row(R::Ring, pos::Vector{Int}, val::AbstractVector{T}; sort::Bool = true) where T
  if sort
    p = sortperm(pos)
    pos = pos[p]
    val = val[p]
  end
  if T === elem_type(R)
    return SRow(R, pos, val)
  else
    mapval = map(R, val)::Vector{elem_type(R)}
    return SRow(R, pos, mapval)
  end
end
```

src/HeckeTypes.jl
```diff
@@ -331,6 +332,7 @@ mutable struct SRow{T, S} # S <: AbstractVector{T}
       end
     end
     r = new{elem_type(R), typeof(S)}(R, S, p)
+    @assert r isa sparse_row_type(R)
```
fingolfin, line 335: This is a bit of a paranoia check that I inserted to make sure I got everything right (and it was helpful, it caught at least one bug :-)). We could now
1. keep it
2. remove it
3. turn it into a type assertion

src/HeckeTypes.jl
```diff
@@ -359,6 +361,7 @@ mutable struct SRow{T, S} # S <: AbstractVector{T}
 
   function SRow{T, S}(A::SRow{T, S}; copy::Bool = false) where {T, S}
     copy || return A
+    @assert Vector{T} == S
```
fingolfin, line 364: Another paranoia check that was valuable. It should get optimized away, I think?

Of course we could also try to make this function work generically, by doing something like this:
```julia
  return new{T, S}(base_ring(A), S(A.values), copy(A.pos))
```
(and adding a suitable copy constructor `ZZRingElem_Array(A:: ZZRingElem_Array)`

src/NumFieldOrd/NfOrd/Clgp/Rel_add.jl
```diff
@@ -161,7 +161,7 @@ function class_group_add_relation(clg::ClassGrpCtx{SMat{ZZRingElem, Hecke.ZZRing
   return class_group_add_relation(clg, a, R)
 end
 
-function class_group_add_relation(clg::ClassGrpCtx{SMat{ZZRingElem, Hecke.ZZRingElem_Array_Mod.ZZRingElem_Array}}, a::FacElem{AbsSimpleNumFieldElem, AbsSimpleNumField}, R::SRow{ZZRingElem}; always::Bool = true, add_orbit = true)
+function class_group_add_relation(clg::ClassGrpCtx{<:SMat{ZZRingElem}}, a::FacElem{AbsSimpleNumFieldElem, AbsSimpleNumField}, R::SRow{ZZRingElem}; always::Bool = true, add_orbit = true)
```
fingolfin, line 164: Motivation here is to hide the implementation detail `ZZRingElem_Array` as much as possible -- with this PR the only references to it now are in `ZZRow.jl`

src/Sparse/Matrix.jl
```diff
@@ -23,6 +23,37 @@ parent(A::SMat) = SMatSpace(base_ring(A), A.r, A.c)
 
 base_ring(A::SMat{T}) where {T} = A.base_ring::parent_type(T)
 
+base_ring_type(::Type{<:SMat{T}}) where {T} = parent_type(T)
```
fingolfin, line 26: This is of course not necessary for this PR, but I noticed it was "missing", and thought I should just add it while I am here...

src/HeckeTypes.jl
thofma: Just keep it

src/HeckeTypes.jl
fingolfin: Indeed, using `===` is a good idea here.
```suggestion
    @assert Vector{T} === S
```

lgoettgens (approved):

PR #1497, opened 2024-05-06
Use Varname to simplify some code

### LINKED ISSUE #171: Strings vs Symbols
Make sure that all "structure" constructors that want strings (or arrays of strings) also
 - take symbols
 - take nothing (and have defaults, different ones for IJulia?)
 
 Possibly also have an interface (at least) to obtain the string(s) back as strings, rather than the Symbols only.

fingolfin:
Resolves #171

### DIFF
--- a/src/NumField/NfAbs/NonSimple.jl
+++ b/src/NumField/NfAbs/NonSimple.jl
@@ -900,7 +900,7 @@ end
 ################################################################################
 
 @doc raw"""
-    number_field(f::Vector{QQPolyRingElem}, s::String="_\$") -> AbsNonSimpleNumField
+    number_field(f::Vector{QQPolyRingElem}, s::VarName="_\$") -> AbsNonSimpleNumField
 
 Let $f = (f_1, \ldots, f_n)$ be univariate rational polynomials, then
 we construct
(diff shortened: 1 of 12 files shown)


src/NumField/NfAbs/NonSimple.jl
```diff
@@ -917,7 +917,11 @@ function number_field(f::Vector{QQPolyRingElem}, s::String="_\$"; cached::Bool =
   return number_field(f, lS, cached = cached, check = check)
 end
 
-function number_field(f::Vector{QQPolyRingElem}, s::Vector{String}; cached::Bool = false, check::Bool = true)
+function number_field(f::Vector{QQPolyRingElem}, s::VarName; cached::Bool = false, check::Bool = true)
+  return number_field(f, String(s); cached = cached, check = check)
```
lgoettgens, line 921:
```suggestion
  return number_field(f, Symbol(s); cached = cached, check = check)
```
I think this is the way you do it most other times

fingolfin: No, this code deliberately uses the `String` case as base case.

fingolfin: Specifically, the base case starts in line 910 and is hidden here, so that's what it looks like:
```julia
function number_field(f::Vector{QQPolyRingElem}, s::String="_\$"; cached::Bool = false, check::Bool = true)
  n = length(f)
  if occursin('#', s)
    lS = Symbol[Symbol(replace(s, "#"=>"$i")) for i=1:n]
  else
    lS = Symbol[Symbol("$s$i") for i=1:n]
  end
  return number_field(f, lS, cached = cached, check = check)
end
```
So it performs string manipulation on `s`.

lgoettgens: Ah ok, thanks

fingolfin:
One OscarCI tested died due to timeout (not caused by this PR here, it's an issue in Oscar), hence automerge does not work.

PR #1500, opened 2024-05-07
Padics done right

joschmitt:
* Add a constructor `unramified_extension(::PadicField, ...)` for `QadicField`.
* Remove some 'illegal' `setprecision` calls on parent objects. I did not touch the various `setprecision!(::CompletionMap, ...)` methods which call `setprecision` on a parent object several times. I don't understand whether the precision of domain and codomain is important for the map or not. The only place I found where such a method is called is in the tests.

If we allow the construction of a `QadicField` both via `QadicField(::Int, ::Int)` and `unramified_extension(::PadicField, ::Int)` (so without and with the base field given), we have the following problem:
```
julia> K = PadicField(2)
Field of 2-adic numbers

julia> L, _ = QadicField(2, 2)
(Unramified extension of 2-adic numbers of degree 2, (2^0 + O(2^64))*a)

julia> base_field(L) === K
false

julia> L2, _ = unramified_extension(K, 2)
(Unramified extension of 2-adic numbers of degree 2, (2^0 + O(2^64))*a)

julia> L === L2 # the QadicField is cached
true

julia> base_field(L2) === K # unramified_extension set the base field in the attribute to K
true

julia> base_field(L) === K # the base field changed
true
```

### DIFF
--- a/src/Deprecations.jl
+++ b/src/Deprecations.jl
@@ -269,3 +269,6 @@ end
 
 @deprecate hom(G::FinGenAbGroup, H::FinGenAbGroup, A::Matrix{ <: Map{FinGenAbGroup, FinGenAbGroup}}) hom_direct_sum(G, H, A)
 @deprecate hom(G::FinGenAbGroup, H::FinGenAbGroup, A::Vector{ <: Map{FinGenAbGroup, FinGenAbGroup}}) hom_tensor(G, H, A)
+
+@deprecate lift(a::QadicRingElem) lift(ZZ, a)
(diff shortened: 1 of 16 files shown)


joschmitt:
Another thing to think about: Should `with_precision(::LocalField, ...)` also adjust the precision of the base field (or more generally, all fields in a tower)? Right now, one has to do
```
with_precision(K, N) do
  return with_precision(base_field(K), N) do
    # do something
  end
end
```
if one for example works with the coefficients of a `QadicFieldElem`.

joschmitt:
I close this for now so that Nemo doesn't try to test against this branch anymore. This needs a big update anyway.

PR #1502, opened 2024-05-07
sparse_row: don't sort 1-element lists

fingolfin:
This is obviously pointless and incurs allocations. At the same time it
is a relatively common case in Oscar, so optimizing this pays off.

### DIFF
--- a/src/Sparse/Row.jl
+++ b/src/Sparse/Row.jl
@@ -73,7 +73,6 @@ function sparse_row(R::Ring)
   return SRow(R)
 end
 
-const _sort = sort
 @doc raw"""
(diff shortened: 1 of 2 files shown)


lgoettgens:
Wouldn't it be even more efficient to use `sort!` instead of `sort`? In the one-length case we don't do a copy now anymore, so why do that in other cases?

fingolfin:
That would change the semantics: `sparse_row` currently does not modify the vectors passed to it as arguments.

One could perhaps implement this by performing the sort inside the `SRow` constructors / performing it on the already constructed `SRow` object. But clearly that's a bigger change. Feel free to make a follow up PR to this PR exploring this venue.

PR #1503, opened 2024-05-10
Repair _torsion_units_lattice_enum

fingolfin:
I verified it works again via this:
```
julia> K, a = cyclotomic_field(13, cached = false);

julia> O = maximal_order(K);

julia> G, mG = torsion_unit_group(O)
(Z/26, Map: G -> O)

julia> length(Hecke._torsion_units_lattice_enum(O)[1])
26
```
Either something like that should be turned into a test -- or alternatively this undocumented function could be deleted, as nothing calls it.


thofma:
I will add a test later

Issue #746, opened 2022-07-08
Speeding up "time to first result"?


fingolfin:
On my M1 Mac with Julia 1.8-rc1:
```
julia> @time @eval maximal_order(quadratic_field(-1)[1]);
  5.577463 seconds (2.85 M allocations: 140.472 MiB, 0.34% gc time, 99.97% compilation time)

julia> @time @eval maximal_order(quadratic_field(-1)[1]);
  0.000288 seconds (121 allocations: 4.948 KiB)
```
This is not a simple matter of caching as it also is fast with different arguments, e.g.
```
julia> @time @eval maximal_order(quadratic_field(-3)[1]);
  0.003174 seconds (635 allocations: 30.745 KiB)
```

Most of this time seems to be spent in compiling. Indeed, starting Julia with `-O1` things are much faster:
```
julia> @time @eval maximal_order(quadratic_field(-1)[1]);
  2.370229 seconds (554.39 k allocations: 25.480 MiB, 99.94% compilation time)

```

I've used [`@snoopl`](https://timholy.github.io/SnoopCompile.jl/stable/reference/#SnoopCompileCore.@snoopl) from SnoopCompile to get some insights into this. So I did
```julia
@snoopl "func_names.csv" "llvm_timings.yaml" begin
    using Hecke
    maximal_order(quadratic_field(-1)[1])
end
```

The resulting data files can be loaded separately:

```
julia> times, info = SnoopCompile.read_snoopl("func_names.csv", "llvm_timings.yaml");

julia> sort(times)[end][1]   # where is most time spent?
0.428102083

julia> sort(times)[end-1][1]  # second worst place?
0.068432708

julia> sum(k for (k,v) in times)
5.172362951000003
```

So yeah, we spend a lot of time in compilation. It compiles 999 method instances, it seems. 

Unfortunately there don't seem to be many functions to help analyze this further...

Attached is the list of method instances, sorted alphabetically; perhaps this already gives some ideas...
[methods.txt](https://github.com/thofma/Hecke.jl/files/9073043/methods.txt)

fingolfin:
As @thofma pointed out: it may well be that in the end, we can't do anything about this other than building a custom baseimage. Quite true, but perhaps `@noinline` and `@nospecialize` or some other tricks can also help at least a bit?

thofma:
@fingolfin @fieker There is a PR over at julia, and someone (don't want to ping them) used Hecke as a benchmark:
[...]

PR #1504, opened 2024-05-10
Change _zero_algebra -> zero_algebra

fingolfin:
The function _zero_algebra was renamed to zero_algebra some time ago, but not all places calling it were updated.

I found this because it caused a bunch of invalidations in code found while investigating #746 (indirectly, of course, as any actual call to `_zero_algebra` could not have worked).

### DIFF
--- a/src/AlgAss/AlgAss.jl
+++ b/src/AlgAss/AlgAss.jl
@@ -190,7 +190,7 @@ end
 # Does anyone actually use this?
 function StructureConstantAlgebra(R::Ring, d::Int, arr::Vector{T}) where {T}
   if d == 0
-    return _zero_algebra(R)
+    return zero_algebra(R)
   end
   mult_table = Array{T, 3}(undef, d, d, d)
   n = d^2
@@ -302,7 +302,7 @@ function StructureConstantAlgebra(O::Union{AbsNumFieldOrder, AlgAssAbsOrd}, I::U
   Fp = GF(p, cached = false)
 
   if r == 0
-    A = _zero_algebra(Fp)
+    A = zero_algebra(Fp)
 
     local _image_zero
 
@@ -430,7 +430,7 @@ function StructureConstantAlgebra(I::Union{ AbsNumFieldOrderIdeal, AlgAssAbsOrdI
   Fp = GF(p, cached = false)
 
   if r == 0
-    A = _zero_algebra(Fp)
+    A = zero_algebra(Fp)
 
     local _image_zero
 
@@ -574,7 +574,7 @@ function StructureConstantAlgebra(O::Union{ RelNumFieldOrder{T, S}, AlgAssRelOrd
   r = length(basis_elts)
 
   if r == 0
-    A = _zero_algebra(Fp)
+    A = zero_algebra(Fp)
 
     local _image_zero
 
@@ -729,7 +729,7 @@ function StructureConstantAlgebra(I::Union{ RelNumFieldOrderIdeal{T, S}, AlgAssR
   r = length(basis_elts)
 
   if r == 0
-    A = _zero_algebra(Fp)
+    A = zero_algebra(Fp)
 
     local _image_zero
 
@@ -995,7 +995,7 @@ function subalgebra(A::StructureConstantAlgebra{T}, e::AssociativeAlgebraElem{T,
   r = size(mult_table, 1)
 
   if r == 0
-    eA = _zero_algebra(R)
+    eA = zero_algebra(R)
     return eA, hom(eA, A, zero_matrix(R, 0, n))
   end
 
--- a/src/AlgAssAbsOrd/LocallyFreeClassGroup.jl
+++ b/src/AlgAssAbsOrd/LocallyFreeClassGroup.jl
@@ -336,7 +336,7 @@ function _unit_group_generators(A::MatAlgebra{<:FinFieldElem})
     @assert is_unit(det(g2))
     push!(res, g1, g2)
   end
-  return map(A, res)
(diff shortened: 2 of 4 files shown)


thofma:
Makes inference worse in some "unrelated" place, thus making the tests fail. (I have seen this happening before. I still don't know why.)

lgoettgens:
Running these failing tests in the REPL does not fail, so this kind of weird to debug.

fingolfin:
Here's a theory why: code which calls `_zero_algebra` by definition cannot work (because there is no `_zero_algebra`). LLVM may then classify this code as having UB (undefined behavior) and that enables many "optimizations" (basically you can replace calls to such code with anything). That in turn may make it possible to "prove" that certain functions are type stable, when they really are not.

If this guess is right, then anything that now fails inference should be checked as to whether it directly or indirectly leads to calls to `_zero_algebra`

fingolfin:
I can reproduce this in the REPL:
```
julia> A = matrix_algebra(GF(2), 2)
Matrix algebra of dimension 4 over Prime field of characteristic 2

julia> @inferred Hecke._unit_group_generators(A)
ERROR: return type Vector{MatAlgebraElem{FqFieldElem, FqMatrix}} does not match inferred return type Any
```

This goes away if I change the last line of `Hecke._unit_group_generators` from `return map(A, res)` to `return res` or `return A, res`.

But `@code_warntype` reports nothing suspicious and indeed even displays the correct and concrete return type. Moreover, if I use the modified version with `return res`, and then do `@inferred map(A,res)` on the top level, there are no complaints...

Could this be a bug in `@inferred` then?!

thofma:
Yeah, more (random) inference errors.

fingolfin:
Fascinating. With this patch,
```
    E = elliptic_curve([1, 2, 3, 4, 5])
    a, b = @inferred CPS_height_bounds(E)
```
fails because inside `CPS_height_bounds` the return type of `CPS_dvev_real` can no longer be inferred. 

OK, I can reduce this to
```
    E = elliptic_curve([1, 2, 3, 4, 5])
    @inferred Hecke.CPS_dvev_real(E, PosInf(), 110)
```

There, it fails to infer the full return type for `Hecke._roots`. With this PR I see
```
│   %156 = Core.kwcall(%155, Hecke._roots, f, v)::Tuple{Vector{AcbFieldElem}, Any, Vector{AcbFieldElem}}
```
without the PR I see
```
│   %156 = Core.kwcall(%155, Hecke._roots, f, v)::Tuple{Vector{AcbFieldElem}, Vector{ArbFieldElem}, Vector{AcbFieldElem}}
```


Narrowing it down:
```
Kx, x = polynomial_ring(QQ, "x")
b2, b4, b6, b8 = (9, 11, 29, 35)
f = 4*x^3 + b2*x^2 + 2*b4*x + b6
@inferred Hecke._roots(f, PosInf(), prec = 110)
```

Will dig more later.

lgoettgens:
I found a very simple example that differs between master and this branch:
```julia
julia> y = [AcbField(64)(1)]
1-element Vector{AcbFieldElem}:
 1.0000000000000000000

julia> @code_warntype real.(y)
MethodInstance for (::var"##dotfunction#234#8")(::Vector{AcbFieldElem})
  from (::var"##dotfunction#234#8")(x1) @ Main none:0
Arguments
  #self#::Core.Const(var"##dotfunction#234#8"())
  x1::Vector{AcbFieldElem}
Body::Any
1 ─ %1 = Base.broadcasted(Main.real, x1)::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1}, Nothing, typeof(real), Tuple{Vector{AcbFieldElem}}}
│   %2 = Base.materialize(%1)::Any
└──      return %2
```
on this branch, while on master:
```julia
julia> @code_warntype real.(y)
MethodInstance for (::var"##dotfunction#225#1")(::Vector{AcbFieldElem})
  from (::var"##dotfunction#225#1")(x1) @ Main none:0
Arguments
  #self#::Core.Const(var"##dotfunction#225#1"())
  x1::Vector{AcbFieldElem}
Body::Vector{ArbFieldElem}
1 ─ %1 = Base.broadcasted(Main.real, x1)::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1}, Nothing, typeof(real), Tuple{Vector{AcbFieldElem}}}
│   %2 = Base.materialize(%1)::Vector{ArbFieldElem}
└──      return %2
```

This then makes inference fail in https://github.com/thofma/Hecke.jl/blob/b3c226459fc2a6118c9de4c800d3e22ed98768bf/src/Misc/Poly.jl#L731 leading to the problems observed by @fingolfin

fingolfin:
@lgoettgens thanks. Unfortunately that seems to be a dead end:
```julia
code_warntype(Base.materialize, (Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1}, Nothing, typeof(real), Tuple{Vector{AcbFieldElem}}}, ))
```
prints perfectly inferred information in both branches.


I tried using `JET` but `@report_opt broadcast(real, y)` reports no issues (note that `@code_warntype broadcast(real, y)` shows the same behavior as your `@code_warntype real.(y)`

And according to `@descend broadcast(real, y)` everything is fine, too.


Aaargh!

fingolfin:
Hah, I just tried with Julia 1.11.0-rc2 and there it works (both Lars example and also the elliptic curve example).

But it's not as if all was great: with that Julia, I get an inference failure with
```
g = @inferred supersingular_polynomial(2)
```
which works fine in Julia 1.10 (both with my branch and with master).


Seems we just can't have nice things *sigh*

PR #1505, opened 2024-05-11
feat: make gcd over number fields faster

thofma:


thofma:
Not terminating for the following example:
```
julia> Qx, x = QQ["x"];

julia> K, a = number_field(x^16 + 36*x^12 - 120*x^10 + 392*x^8 - 432*x^6 + 216*x^4 - 48*x^2 + 4, "a");

julia> R, (x, y, t) = polynomial_ring(K, ["x", "y", "t"]);

julia> gcd((x + 1)^10 * (x + y)^5, (x + 1)^5 * (x + y)^10 * t^10)
```
Because `coeff(gl, 0) == 0` and so `glp == 0` for every `p`.

thofma:
Let's do a mini multivariate polynomial triage and decide:
1. Should we always use the trick to work over $\mathbf{F}_p[X]/(f)$ with squarefree $f$? But this does not seem to scale well. Or stick to the idea of using prime ideals of degree $1$? Or mix both?
2. Should we recognize rational polynomials and do the computation with `fmpq_mpoly`? (We could speed up the `finish!` in this case).
3. Do we agree that for towers, we should pass to an `absolute_simple_field`? (Of course caching it together with both maps).

thofma:
Should be good to go. I noticed no regression. Even for `rationals_as_number_field()`, passing to the rationals is still faster.

thofma:
@benlorenz Some non-canonical output changed for the booktest (but the result is isomorphic). What is the recommended way? Do a breaking release and then adjust it?

Issue #1509, opened 2024-05-14
setprecision!(f::CompletionMap)


simonbrandhorst:
```
julia> K,toK = completion(F, 2*OF);

julia> F,_ = cyclotomic_field(3)
(Cyclotomic field of order 3, z_3)

julia> OF = maximal_order(F);

julia> K,toK = completion(F, 2*OF);

julia> setprecision!(toK,5)
```
it seems buggy also for the other methods in the same file

By the way, I found it surprising that the completion map can have a different precision than its codomain.

joschmitt:
(That is also one of the questions I have in #1500.)

simonbrandhorst:
For the record, another bug in the ramified case:

```
  Qx, x = QQ["x"];
  f = x^9 - 828*x^7 - 4371*x^6 + 226071*x^5 + 2371023*x^4 - 14243253*x^3 - 318645900*x^2 - 1637156439*x - 2754662093;

  K, a = number_field(f, "a");
  OK = maximal_order(K);
  p = 3;
  P = prime_ideals_over(OK, p)[1];
  C, mC = completion(K, P, 8);


  setprecision!(mC, 20)
  setprecision!(mC, 21)
  setprecision!(mC, 20) #throws
```

joschmitt:
Looks like there is an `assert` that effectively prevents one from setting the precision back to an earlier value.
I also just realized that the field `.precision` of the map is never updated, although it is used in the `setprecision!` function. 

@thofma These `setprecision!` methods need to be revised. I could try to understand what they are supposed to be doing. Or is this written down somewhere?

thofma:
@fieker we forgot (did not have time) to discuss this. What is up with this `@assert`?

simonbrandhorst:
#1821 fixes my personal problem. The second one 

```
julia> Qx, x = QQ["x"];

julia> f = x^9 - 828*x^7 - 4371*x^6 + 226071*x^5 + 2371023*x^4 - 14243253*x^3 - 318645900*x^2 - 1637156439*x - 2754662093;

julia> K, a = number_field(f, "a");

julia> OK = maximal_order(K);

julia> p = 3;

julia> P = prime_ideals_over(OK, p)[1];

julia> C, mC = completion(K, P, 8);

julia> setprecision!(mC, 20)

julia> setprecision!(mC, 21)

julia> setprecision!(mC, 20) #throws
ERROR: AssertionError: !(new_prec in keys(Kp.def_poly_cache))

[...]

PR #1511, opened 2024-05-15
Maybe fix 1509

joschmitt:
My educated guess is that `setprecision!(::CompletionMap, ...)` should change the precision of the codomain (if any).

Related to #1509.

### DIFF
--- a/test/LocalField/Completions.jl
+++ b/test/LocalField/Completions.jl
@@ -43,6 +43,7 @@ end
   F, _ = cyclotomic_field(3)
   OF = maximal_order(F);
   K, toK = completion(F, 2*OF);
+  @test iszero(preimage(toK, toK(F(0))))
   setprecision!(toK, 10)
   @test precision(toK(F(1))) == 10
   setprecision!(toK, 70)
--- a/test/LocalField/Completions.jl
+++ b/test/LocalField/Completions.jl
@@ -48,6 +48,19 @@ end
   @test precision(toK(F(1))) == 10
   setprecision!(toK, 70)
   @test precision(toK(F(1))) == 70
+
+  K, toK = Hecke.unramified_completion(F, 2*OF)
+  setprecision!(toK, 10)
+  @test precision(toK(F(1))) == 10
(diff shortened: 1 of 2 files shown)


test/LocalField/Completions.jl
```diff
@@ -39,6 +39,14 @@ end
   @test precision(bb) >= 20
 end
 
+@testset "Issue 1509" begin
+  F, _ = cyclotomic_field(3)
+  OF = maximal_order(F);
+  K, toK = completion(F, 2*OF);
+  setprecision!(toK, 100)
+  @test precision(toK(F(1))) == 100
```
joschmitt, line 47: This test doesn't trigger the issue.
The (default) precision of `toK` is 64 and the issue only appears if the new precision is lower than the old one.

simonbrandhorst:
Note that constructing stuff with `unramified_extension` the `setprecision!` method still errors ... but it is internal and maybe I am not using it correctly so I would say that is a different issue.
```
@testset "Issue 1509" begin
  F, _ = cyclotomic_field(3)
  OF = maximal_order(F);
  K, toK = Hecke.unramified_completion(F, 2*OF);
  setprecision!(toK, 200)
  @test precision(toK(F(1))) == 200
end

test/LocalField/Completions.jl
simonbrandhorst: ok will fix it

joschmitt:
> Note that constructing stuff with `unramified_extension` the `setprecision!` method still errors ... but it is internal and maybe I am not using it correctly so I would say that is a different issue.
> 
> ```
> @testset "Issue 1509" begin
>   F, _ = cyclotomic_field(3)
>   OF = maximal_order(F);
>   K, toK = Hecke.unramified_completion(F, 2*OF);
>   setprecision!(toK, 200)
>   @test precision(toK(F(1))) == 200
> end
> ```

I don't think you are using it wrong... I try to fix it.

joschmitt:
All of this is just guess work from my side. Obviously these functions did not work for some time, maybe they never did.

PR #1513, opened 2024-05-15
Simplify MatAlgebraElem type params

fingolfin:
The second type param is fully determined by the first and third,
so get rid of it.

### DIFF
--- a/src/AlgAss/AlgMat.jl
+++ b/src/AlgAss/AlgMat.jl
@@ -18,7 +18,7 @@ basis(A::MatAlgebra) = A.basis
 
 has_one(A::MatAlgebra) = true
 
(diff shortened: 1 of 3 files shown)


thofma:
Do you plan to do this for all the types in Oscar?

fingolfin:
No plans of any kind is behind. Rather, I hit upon this will wading through invalidations in Hecke, and trying to understand the various type parameters of types involved in that. It was surprising that `MatAlgebra{T, S}` and `MatAlgebraElem{T, S, Mat}` mean different things by `S`. I had to understand what is going on and eventually figured it out, and then thought "why is it like that???" and ended up with this change.

The actual invalidation fix doesn't need this, but I thought I should follow the good old campsite rule...

fingolfin:
(I am also not aware of other instances of such redundancies in the OSCAR ecosystem)

thofma:
Thanks, makes sense.

PR #1514, opened 2024-05-15
Avoid some type instability issues

fingolfin:
... resp. trigger points for method invalidations

### DIFF
--- a/src/AlgAss/AlgMat.jl
+++ b/src/AlgAss/AlgMat.jl
@@ -12,7 +12,7 @@ base_ring(A::MatAlgebra{T, S}) where {T, S} = A.base_ring::parent_type(T)
 
 base_ring_type(::Type{MatAlgebra{T, S}}) where {T, S} = parent_type(T)
 
-coefficient_ring(A::MatAlgebra{T, <:MatrixElem{S}}) where {T, S} = A.coefficient_ring::base_ring_type(S)
+coefficient_ring(A::MatAlgebra{T, S}) where {T, S} = A.coefficient_ring::base_ring_type(S)
 
 basis(A::MatAlgebra) = A.basis::Vector{elem_type(A)}
 
--- a/src/AlgAss/AlgMat.jl
+++ b/src/AlgAss/AlgMat.jl
@@ -12,9 +12,9 @@ base_ring(A::MatAlgebra{T, S}) where {T, S} = A.base_ring::parent_type(T)
 
 base_ring_type(::Type{MatAlgebra{T, S}}) where {T, S} = parent_type(T)
 
-coefficient_ring(A::MatAlgebra) = A.coefficient_ring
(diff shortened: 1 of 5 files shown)


src/AlgAss/AlgMat.jl
```diff
@@ -12,9 +12,9 @@ base_ring(A::MatAlgebra{T, S}) where {T, S} = A.base_ring::parent_type(T)
 
 base_ring_type(::Type{MatAlgebra{T, S}}) where {T, S} = parent_type(T)
 
-coefficient_ring(A::MatAlgebra) = A.coefficient_ring
+coefficient_ring(A::MatAlgebra{T, <:MatrixElem{S}}) where {T, S} = A.coefficient_ring::base_ring_type(S)
```
lgoettgens, line 15:
```suggestion
coefficient_ring(A::MatAlgebra{T, <:MatrixElem{S}}) where {T, S} = A.coefficient_ring::parent_type(S)
```
Or alternatively take the base_ring_type of the second type param of MatAlgebra

fingolfin: Ah, well spotted, that was me revising my patch in completely: this code used to read
```suggestion
coefficient_ring(A::MatAlgebra{T, S}) where {T, S} = A.coefficient_ring::base_ring_type(S)
```
which however requires a newer AA release adding that `base_ring_type`. So I meant to rewrite it, and apparently didn't finish (I was working on this stuff on an off for over a week, always half an hour here or there, always being interrupted, and last night I just decided to just submit it, no matter what. There is plenty more to be done, but I am not sure when I'll get to it)

src/QuadForm/Misc.jl
```diff
@@ -879,7 +879,7 @@ function _find_quaternion_algebra(b, P, I)
     end
     M = vcat(M, v)
     push!(elts, f(L[i])) # cache
-    fl, w = can_solve_with_solution(M, target, side = :left)
+    fl, _ = can_solve_with_solution(M, target, side = :left)
     if fl
```
lgoettgens, line 883:
```suggestion
    if can_solve(M, target, side = :left)
```
no need to compute the solution and then discard it

src/AlgAss/AlgMat.jl
lgoettgens: I prepared a new AA release in https://github.com/Nemocas/AbstractAlgebra.jl/pull/1708 to make the fix available

src/QuadForm/Misc.jl
```diff
@@ -862,9 +862,9 @@ function _find_quaternion_algebra(b, P, I)
 
   found = false
   U, h = unit_group(R)
-  sign_vector = g -> begin
+  function sign_vector(g)
```
thofma, line 865: why not anonymous?

lgoettgens: My guess is that anonymous functions don't get a recognizable name in Jet analysis, but no real idea

PR #1516, opened 2024-05-18
feat: adjust to local field changes in Nemo

thofma:


thofma:
@joschmitt there is one `prime_field` call remaining in `test/LocalField/neq.jl:52`, but to fix it one needs the `completion(K, P, Qp)` version.

PR #1520, opened 2024-05-20
Fix `irreducible_components(::ZZLat)`

StevellM:
@simonbrandhorst I guess this is the good fix. 

For the example that I added in the test, depending on the ordering of `sv` (short vectors list) in `_irreducible_components_short_vectors`, I could have different outputs. 

The original issue was that the output of `_irreducible_components_short_vectors` after full iteration was a list of two lattices whose ranks did not sum up to the rank of the original lattice. This was mainly because the algorithm discarded too many short vectors. Indeed, if I understand well, the algorithm was discarding all unvisited short vectors orthogonal to the first vector of `B`, instead of checking whether it was orthogonal to the all `B`.
Hence, the final basis `B` was two "small" and we did not get a honest splitting (in fact there was a non-trivial gluing, and then the upper bound for the iteration was too small for the orthogonal complement, resulting in an empty lattice after iteration in line 1915).


simonbrandhorst:
Thank you!

PR #1522, opened 2024-05-21
Replace some `setprecision!(::ParentObject)`

joschmitt:
This replaces many `setprecision!(::ParentObject)` calls by `with_precision` blocks.
Also removes `prime_field` in the context of local fields.

### DIFF
--- a/src/LocalField/Completions.jl
+++ b/src/LocalField/Completions.jl
@@ -336,7 +336,7 @@ function setprecision!(f::CompletionMap{LocalField{QadicFieldElem, EisensteinLoc
     gq = _increase_precision(gq, pol_gq, div(f.precision+e-1, e), ex, P)
     f.inv_img = (gq, f.inv_img[2])
 
-    Zp = maximal_order(prime_field(Kp))
+    Zp = maximal_order(absolute_base_field(Kp))
     Qq = base_field(Kp)
 
     setprecision!(Qq, ex)
--- a/src/LocalField/Completions.jl
+++ b/src/LocalField/Completions.jl
@@ -173,9 +173,9 @@ The map giving the embedding of $K$ into the completion, admits a pointwise
 preimage to obtain a lift. Note, that the map is not well defined by this
 data: $K$ will have $\deg P$ many embeddings.
 
-The map is guaranteed to yield a relative precision of at least `preciscion`.
+The map is guaranteed to yield a relative precision of at least `precision`.
 """
-function completion(K::AbsSimpleNumField, P::AbsNumFieldOrderIdeal{AbsSimpleNumField, AbsSimpleNumFieldElem}, precision::Int = 64)
+function completion(K::AbsSimpleNumField, P::AbsNumFieldOrderIdeal{AbsSimpleNumField, AbsSimpleNumFieldElem}, precision::Int = 64, Qp::Union{PadicField,Nothing} = nothing)
   #to guarantee a rel_prec we need to account for the index (or the
   #elementary divisor of the trace mat): the map
   #is for the field (equation order), the precision is measured in the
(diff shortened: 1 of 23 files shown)


src/LocalField/Completions.jl
```diff
@@ -189,9 +189,11 @@ function completion(K::AbsSimpleNumField, P::AbsNumFieldOrderIdeal{AbsSimpleNumF
   f = degree(P)
   e = ramification_index(P)
   prec_padics = div(precision+e-1, e)
-  Qp = PadicField(minimum(P), prec_padics, cached = false)
+  if isnothing(Qp)
+    Qp = padic_field(minimum(P), precision = prec_padics, cached = false)
+  end
   Zp = maximal_order(Qp)
-  Qq, gQq = QadicField(minimum(P), f, prec_padics, cached = false)
+  Qq, gQq = unramified_extension(Qp, f, precision = prec_padics, cached = false)
```
fingolfin, line 196: I have no detail about the maths in here, so I just trust this change (which unlike the others seems not completely "obvious") is correct

src/LocalField/Conjugates.jl
```diff
@@ -53,15 +51,17 @@ function newton_lift(f::ZZPolyRingElem, r::LocalFieldElem, precision::Int = pare
...
-    setprecision!(Q, p)
     setprecision!(qf, p)
     setprecision!(qfs, p)
-    r = r - qf(r)*o
-    if Nemo.precision(r) >= n
-      setprecision!(Q, n)
+    r = with_precision(Q, p) do
+      return r - qf(r)*o
+    end
+    if precision(r) >= n
```
fingolfin, line 59: It seems to me that before, the precision of `r` resp. `Q` was always restricted to `n` if `precision(r) >= n`, while now it might stay at a larger value. Presumably this is fine, though, or even an improvement?

Unfortunately this particular case does not seem to be covered by CI.

src/LocalField/Conjugates.jl
```diff
@@ -93,10 +93,11 @@ function roots(C::qAdicRootCtx, n::Int = 10)
   lf = factor_mod_pk(Array, C.H, n)
   rt = QadicFieldElem[]
   for Q = C.Q
-    Q.prec_max = n
     for x = lf
       if is_splitting(C) || degree(x[1]) == degree(Q)
-        append!(rt, roots(Q, x[1], max_roots = 1))
+        with_precision(Q, n) do
```
fingolfin, line 98: would it be better to move the `with_precision` outside the `for x = lf` loop?

fingolfin (commented):
This PR is big and has been sitting here since May. I tried to give it a review, but the `precision` changes seem rather delicate to me, and there are so many changes in here, most unrelated, that I gave up halfway through :-(. Sorry about that (I am not the most qualified person for this anyway, so perhaps not a big loss).

But I noticed that the "change constructors to snake case" part is huge, yet mostly trivial. Perhaps that could be split into a separate PR (and perhaps some of the other precision related changes into yet another) which then might make each individual PR easier and quicker to review?

joschmitt:
> This PR is big and has been sitting here since May. I tried to give it a review, but the `precision` changes seem rather delicate to me, and there are so many changes in here, most unrelated, that I gave up halfway through :-(. Sorry about that (I am not the most qualified person for this anyway, so perhaps not a big loss).

Oh, sorry, I thought this was marked as draft. I didn't really proceed with this: there is a lot of code that apparently assumes that the precision of some field is set correctly at some point before. Since I don't know half of the involved theory, most of this was educated guessing and trial and error with the CI tests.
This also breaks OSCAR for the same reason: functions in OSCAR seem to assume that some functions in Hecke set the precision "correctly". I have a local branch where tried to fix *that*, but gave up at some point.

Splitting off the trivial renamings and such sounds like a good idea. I'll look into it.

PR #1528, opened 2024-05-27
feat: make has_preimage_with_preimage faster for abelian groups

thofma:
- Use to a SolveContext to cache the implicit normal form
  computations


fieker:
Good idea - I solved this mostly, for my applications, by providing an
interface to compute many preimages at once (predating the Ctx), but
this is more neat!

PR #1529, opened 2024-05-28
Add residue ring of local fields

joschmitt:


test/LocalField.jl
```diff
@@ -4,4 +4,5 @@
   include("LocalField/LocalField.jl")
   include("LocalField/neq.jl")
   include("LocalField/Completions.jl")
+  include("LocalField/ResidueRing.l")
```
thofma, line 7:
```suggestion
  include("LocalField/ResidueRing.jl")
```

src/LocalField/ResidueRing.jl
```diff
@@ -0,0 +1,357 @@
...
+  @req parent(a) === parent(b) "Parents do not match"
+  @req !is_zero(b) "Division by 0"
+  if is_zero(a)
+    return zero(parent(a))
+  end
+  @req valuation(data(a)) >= valuation(data(b)) "Division not possible"
+  c = divexact(data(a), data(b))
+  setprecision!(c, _exponent(parent(a)))
+  return parent(a)(c, copy = false, check = false)
+end
```
thofma, line 251: @fieker Can you confirm that this makes sense?

thofma:
Splendid! Can you hook up the solving/kernel to use the Howell form?

fieker:
the precision of the zero is wrong, I think. It should(?) depend on the precision of a and b, not of the parent...

src/LocalField/ResidueRing.jl
```diff
@@ -0,0 +1,357 @@
...
+
+################################################################################
+#
+#  Basic functionality
+#
+################################################################################
+
+characteristic(R::LocalFieldValuationRingResidueRing) = prime(_field(R))^_exponent(R)
+
+zero(R::LocalFieldValuationRingResidueRing) = R()
```
fieker, line 37: unless the zerois exact, it needs a precision....

thofma: This is `Zp/p^k`, so the zero is the exact zero of this ring.

joschmitt: I mean, this one is exact because it is the zero of the residue ring?

fieker (approved):
I'm not 100% sure all the precision is correct, but otherwise it is looking good

PR #1531, opened 2024-05-29
Small improvement in genus symbol enumerations

StevellM:


thofma:
It has been some time, but I think it looks good.

src/QuadForm/Herm/Genus.jl
```diff
@@ -1835,7 +1835,8 @@ function hermitian_genera(E::Hecke.RelSimpleNumField, rank::Int, signatures::Dic
   @req rank >= 0 "Rank must be a non-negative integer"
   K = base_field(E)
   OE = maximal_order(E)
-  bd = union!(support(2*maximal_order(K)), support(discriminant(OE)))
+  bds = union!(support(2*maximal_order(K)), support(discriminant(OE)))
+  bd = deepcopy(bds)
```
simonbrandhorst, line 1839: @StevellM 
please explain this change
`bd` is modified afterwards and `bds` then used below in line 1871. This should be a change in behavior of the function.
Why is the output still the same?

StevellM: I can explain: `bds` is used at the end to "filter out" the unimodular local symbols at the good primes. Before, we were perturbating this list by adding new primes and thus, we could still keep unimodular local symbols at good primes (I remarked it by the printing).

So we can keep track of `bds` for filtering at the end, and we use `bd` for enumerating possible local genera.

PR #1532, opened 2024-05-31
Add missing parentheses in complex printing and fix `free_resolution(::FinGenAbGrp)`

lgoettgens:
The first commit is part of https://github.com/oscar-system/Oscar.jl/pull/3815.

The second commit makes `free_resolution(::FinGenAbGrp)` work again, and also the printing of its return value. This seems to be some fallout from the `rank` -> `torsion_free_rank` thing. And I added a dummy test to keep this working in the future (but without real test).


fingolfin:
Triage is OK with this being merged.

PR #1534, opened 2024-06-04
Use mat_entry_ptr in more places

fingolfin:


examples/NfMat.jl
```diff
@@ -837,15 +837,13 @@ function Hecke.coeff(M::NfMatElem, i::Int)
   return map(x->coeff(x, i), M)
 end
 
-function getindex_raw(M::QQMatrix, i::Int, j::Int)
-  return ccall((:fmpq_mat_entry, Nemo.libflint), Ptr{QQFieldElem}, (Ref{QQMatrix}, Int, Int), M, i-1, j-1)
-end
+getindex_raw(M::QQMatrix, i::Int, j::Int) = mat_entry_ptr(M, i, j)
```
thofma, line 840:
```suggestion
```
you seem to replace all uses of this function?

fingolfin: There are still 56 occurrences of `getindex_raw` in this file and in `src/LargeField/misc2.jl`, and I did not want to vet every single one and risk screwing that up, so I decided to leave this in for now.

Issue #1526, opened 2024-05-24
Random test failures in `NfOrd/PicardGroup.jl`


lgoettgens:
observed in the downstream tests in AA in https://github.com/Nemocas/AbstractAlgebra.jl/actions/runs/9224253991/job/25379194623?pr=1715#step:7:169

```julia
Error During Test at /home/runner/.julia/packages/Hecke/SZzwT/test/NfOrd/PicardGroup.jl:61
  Got exception outside of a @test
  AssertionError: $(Expr(:escape, :(nbits(pp) < 10000)))
  Stacktrace:
    [1] macro expansion
      @ ~/.julia/packages/Hecke/SZzwT/src/Assertions.jl:507 [inlined]
    [2] evaluate_mod(a::FacElem{AbsSimpleNumFieldElem, AbsSimpleNumField}, B::AbsSimpleNumFieldOrderFractionalIdeal)
      @ Hecke ~/.julia/packages/Hecke/SZzwT/src/NumField/NfAbs/CompactRepresentation.jl:361
    [3] macro expansion
      @ ~/.julia/packages/Hecke/SZzwT/src/Hecke.jl:476 [inlined]
    [4] compact_presentation(a::FacElem{AbsSimpleNumFieldElem, AbsSimpleNumField}, nn::Int64; decom::Dict{AbsSimpleNumFieldOrderIdeal, ZZRingElem}, arb_prec::Int64, short_prec::Int64)
      @ Hecke ~/.julia/packages/Hecke/SZzwT/src/NumField/NfAbs/CompactRepresentation.jl:217
    [5] compact_presentation
      @ ~/.julia/packages/Hecke/SZzwT/src/NumField/NfAbs/CompactRepresentation.jl:10 [inlined]
    [6] macro expansion
      @ ~/.julia/packages/Hecke/SZzwT/src/Hecke.jl:476 [inlined]
    [7] _ispower(a::FacElem{AbsSimpleNumFieldElem, AbsSimpleNumField}, n::Int64; with_roots_unity::Bool, decom::Dict{AbsSimpleNumFieldOrderIdeal, ZZRingElem}, trager::Bool)
      @ Hecke ~/.julia/packages/Hecke/SZzwT/src/NumField/NfAbs/CompactRepresentation.jl:435
[...]

PR #1538, opened 2024-06-05
Newton Lifting for points of elliptic curves over function fields

simonbrandhorst:
Adds a multivariate p-adic newton iteration to lift points on Elliptic curves over function fields of finite fields 
to the function field of a number field.


simonbrandhorst:
@HechtiDerLachs  here is the lifting

src/EllCrv/NewtonLift.jl
```diff
@@ -0,0 +1,295 @@
...
+
+function _can_reconstruct_with_reconstruction(E::EllipticCurve, padic_point, F_to_K, prec)
+  @vprint :NewtonIteration 2 "trying rational reconstruction\n"
+  K = codomain(F_to_K)
+  # prec0 = precision(K)
+  #setprecision!(F_to_K, prec) # leads to difficult bugs
+
+  xnF, ynF, dnF = [map_coefficients(x->preimage(F_to_K,K(x); small_lift=true), a) for a in padic_point]
+
+  #setprecision!(F_to_K, prec0)
```
simonbrandhorst, line 92: @joschmitt  thanks to your work, I could get this pr running. 

In this bit the `setprecision!` method fails and I am out of my depth fixing it. 
would you have a look? @fieker 

(merely removing the assertion does not help.)

src/EllCrv/NewtonLift.jl
```diff
@@ -0,0 +1,295 @@
...
+
+  fl || return false, point
+
+  #@show  minimum(_valuation.(M*v - EQc))
+  @assert minimum(_valuation.(M*v - EQc)) >= 2*prec
+  @assert minimum(_valuation.(v)) >= prec
+  t = gen(OKt)
+
+  xne = sum(v[i+1]*t^i for i in 0:xd; init=zero(OKt))
+  yne = sum(v[i+xd+2]*t^i for i in 0:yd; init=zero(OKt))
```
fieker, line 247: Really? Computing powers of `t` this way? I agree that in this context this is not a runtime problem, but from a complexity point of view, you trade exactly `yd` multiplications with definitely more, possibly up to `yd^2/2`.
Assuming that `OKt` is a poly ring, a univariate one, then `OKt(v[xd+2:xd+yd])` or s.th. very similar, shuold do the job with 0 multiplications....

simonbrandhorst: Right, thanks for the reminder. Changed accordingly.

simonbrandhorst:
Not sure what to do with the failing oscar CI tests. Failures should be unrelated.

thofma:
@benlorenz or @lgoettgens: https://github.com/oscar-system/Oscar.jl/blob/master/experimental/Experimental.jl#L76C10-L76C46 this does not seem to be what we want it to be: https://github.com/thofma/Hecke.jl/actions/runs/9396919418/job/25879076324?pr=1538? Any idea how we should fix this in the downstream tests?

lgoettgens:
No idea how this even works. All I can add: the downstream tests in Nemo/AA don't seem to have this type of failure.

src/EllCrv/NewtonLift.jl
```diff
@@ -0,0 +1,302 @@
...
+    pt = pt[1]
+    x0 = -constant_coefficient(pt)
+    y0 = F(0)
+    push!(sing, [x0,y0,t0])
+  end
+  return sing
+end
+
+@doc r"""
+    newton_lift(E::EllipticCurve, P::EllipticCurvePoint, reduction_map; max_iterations=16) -> Bool, EllipticCurvePoint
```
joschmitt, line 43: The OSCAR tests fail because of this doc string.
```suggestion
@doc raw"""
    newton_lift(E::EllipticCurve, P::EllipticCurvePoint, reduction_map; max_iterations=16) -> Bool, EllipticCurvePoint
```
seems to be better.

src/EllCrv/NewtonLift.jl
```diff
@@ -0,0 +1,302 @@
+#import Hecke.AbstractAlgebra
+export newton_lift
+add_verbosity_scope(:NewtonIteration)
+
+_valuation(x::RingElem) = iszero(x) ? QQ(precision(parent(x))) : valuation(x)
+
+
+r"""
```
joschmitt, line 8:
```suggestion
raw"""
```
Also here. `r""` produces a regex apparently.

joschmitt:
Now there is a failure https://github.com/thofma/Hecke.jl/actions/runs/9400468865/job/25890160049?pr=1538 looking similar to #1526. Should I just restart?

thofma:
Yes, ignore it

src/EllCrv/NewtonLift.jl
simonbrandhorst: thanks for spotting this! `r` was the python way.

PR #1540, opened 2024-06-05
Clean up `LocalFieldValuationRing` a bit

joschmitt:
Mostly moves around code in `src/LocalField/Ring.jl` and adds conformance tests.

Also adds a bit nicer printing, makes `valuation_ring` an attribute of the field and fixes `canonical_unit`.

### DIFF
--- a/src/LocalField/Ring.jl
+++ b/src/LocalField/Ring.jl
@@ -6,117 +6,151 @@
 ################################################################################
 # CHECK precision!!!
 
-function Base.show(io::IO, Q::LocalFieldValuationRing)
-  print("Integers of ", Q.Q)
-end
+################################################################################
+#
+#  Parent type etc
+#
+################################################################################
 
-function MaximalOrder(Q::QadicField)
-  return LocalFieldValuationRing{QadicField, QadicFieldElem}(Q)
-end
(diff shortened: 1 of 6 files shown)


src/LocalField/Ring.jl
```diff
@@ -6,117 +6,151 @@
...
-end
+is_unit(a::LocalFieldValuationRingElem) = !iszero(a) && valuation(a) == 0
 
-function AbstractAlgebra.expressify(a::LocalFieldValuationRingElem; context=nothing)
-  return expressify(a.x; context=context)
+function canonical_unit(a::LocalFieldValuationRingElem)
+  iszero(data(a)) && return setprecision(parent(a)(1), precision(a))
+  v = ZZ(absolute_ramification_index(_field(parent(a)))*valuation(data(a)))
+  pi = uniformizer(_field(parent(a)))
+  return LocalFieldValuationRingElem(parent(a), data(a)//pi^v)
```
joschmitt, line 62: The biggest non-trivial change is in `canonical_unit`.
* Before this would return the inverse of what it does now? So an element `u` such that `a*u` is "nicer". The way I understand the AbstractAlgebra documentation, it should be `a//u`.
* The function used the prime instead of the uniformizer, so this wouldn't work for `LocalField` where the valuation might be a fraction.

PR #1554, opened 2024-06-25
Remove unused to_array and to_fmpz_mat methods

fingolfin:
Nothing uses them and the names are suboptimal (so if someone
needs this functionality we'd be better of implementing these
as `convert` methods or constructors or something like that).


fingolfin:
@thofma OK to merge this?

PR #1558, opened 2024-07-06
fix: bug in short_elem of trivial ideal

thofma:
- cache only if the weights are trivial


thofma:
changing the book output, so let's postpone this

PR #1576, opened 2024-08-13
feat: implement solve interface for rings of integers

thofma:
Todo:
- [x] Reuse the pseudo Hermite form in the kernel call
- [x] More tests

@joschmitt Am I doing this correctly? Do I need to handle both `:right` and `:left` variants?


joschmitt:
> @joschmitt Am I doing this correctly? Do I need to handle both `:right` and `:left` variants?

Looks good. (I realized I made a mistake in the documentation: `side` should always be a keyword argument apparently.)
And one needs to handle both `:left` and `:right`, yes.

thofma:
OK, thanks. And I also saw the `;side` thing in AA. I have a fix locally.

PR #1592, opened 2024-09-03
Power series as valuation rings

joschmitt:
Some functionality for valuation rings of local fields in positive characteristic


joschmitt:
That one OSCAR CI job keeps crashing in the book tests. I can't imagine that this is related to the changes here?

PR #1604, opened 2024-09-11
Use `padic_field`/`qadic_field` and prepare deprecations

joschmitt:
The "trivial" part of #1522.


fingolfin (approved):
Excellent!

lgoettgens:
@thofma I think it would make sense to get this merged before doing the heavy work in Nemo that then needs https://github.com/thofma/Hecke.jl/pull/1522.

joschmitt:
Nothing should need #1522. I think you mean #1605.

lgoettgens:
> Nothing should need #1522. I think you mean #1605.

Yes, exactly. Thanks for pointing this out.

Issue #1428, opened 2024-03-14
LocalFundamentalClass Serre tests fail randomly


lgoettgens:
```julia
LocalFundamentalClass Serre: Test Failed at /home/runner/.julia/packages/Hecke/1fdag/test/LocalField/neq.jl:57
  Expression: iszero(a) || valuation(a) > 20
Stacktrace:
 [1] macro expansion
   @ ~/.julia/packages/Hecke/1fdag/test/LocalField/neq.jl:57 [inlined]
 [2] macro expansion
   @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1[151](https://github.com/Nemocas/AbstractAlgebra.jl/actions/runs/8281920352/job/22661589864?pr=1636#step:7:152) [inlined]
 [3] top-level scope
   @ ~/.julia/packages/Hecke/1fdag/test/LocalField/neq.jl:45
```
https://github.com/Nemocas/AbstractAlgebra.jl/actions/runs/8281920352/job/22661589864?pr=1636#step:7:147

I already observed this a few times, but it usually goes away when restarting the job.

thofma:
Different error, same file:
```
Error During Test at /home/runner/work/AbstractAlgebra.jl/AbstractAlgebra.jl/oscar-dev/Hecke/test/LocalField/neq.jl:44
  Got exception outside of a @test
  ArgumentError: Unable to solve linear system
  Stacktrace:
    [1] #solve#10
      @ ~/work/AbstractAlgebra.jl/AbstractAlgebra.jl/src/Solve.jl:568 [inlined]
    [2] #solve#9
      @ ~/work/AbstractAlgebra.jl/AbstractAlgebra.jl/src/Solve.jl:563 [inlined]
    [3] artin_schreier_equation(X::Hecke.ArtinSchreierSolveCtx{FqMatrix, Vector{FqFieldElem}}, c::FqFieldElem)
      @ Hecke ~/work/AbstractAlgebra.jl/AbstractAlgebra.jl/oscar-dev/Hecke/src/LocalField/neq.jl:542
    [4] frobenius_equation2(c::Hecke.LocalFieldElem{Hecke.LocalFieldElem{QadicFieldElem, Hecke.EisensteinLocalField}, Hecke.UnramifiedLocalField}, F::PadicField; frobenius::Hecke.MapEvalCtx, start::Nothing)
[...]

PR #1617, opened 2024-09-20
Fixes Serre (improves #1428)

fieker:
Given an element of norm 1, Hilbert 90 assures it can be written as sigma(x)/x
Generically, this is solves via Lagrage resolvent and randomization. (frobenius_equation). If this fails, a much slower method is called\that constructs the element digit-by-digit (frobenius_equation2). The 1st function has to increase (temporarily) the precision. However then the norm 1 condition is no longer true... and thus the input for the second function illegal hence the failure.

### DIFF
--- a/src/LocalField/neq.jl
+++ b/src/LocalField/neq.jl
@@ -488,6 +488,8 @@ end
 
     Find an element `x` in `parent(c)` such that `frobenius(x, d) = x*c`.
     If the norm of `c` is one, this is supposed to work.
+
+    (Hilbert 90)
 """
 function frobenius_equation(d::Int, c::FinFieldElem)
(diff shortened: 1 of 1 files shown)


src/LocalField/neq.jl
```diff
@@ -520,6 +522,7 @@ end
     artin_schreier_equation(d::Int, c::Union{fpFieldElem, fqPolyRepFieldElem})
 
     Find an element `x` in `parent(c)` such that `frobenius(x, d) -x = c`.
+    Additive Hilber 90.
```
fingolfin, line 525:
```suggestion
    Additive Hilbert 90.
```

PR #1619, opened 2024-09-20
Add unsafe operations for FinGenAbGroupElem

fingolfin:


src/GrpAb/Elem.jl
```diff
@@ -224,6 +224,65 @@ end
...
+#function set!(x::FinGenAbGroupElem, y::FinGenAbGroupElem)
+#  set!(x.coeff, y.coeff)
+#  # no reduction necessary
+#  return x
+#end
+
+function add!(x::FinGenAbGroupElem, y::FinGenAbGroupElem, z::FinGenAbGroupElem)
+  add!(x.coeff, y.coeff, z.coeff)
+  assure_reduced!(parent(x), x.coeff)
+  return x
```
lgoettgens, line 255:
```suggestion
  return reduce!(x)
```
I think I would find this even easier to understand (and less code duplication). Same for the other functions. But this is just personal preference

lgoettgens (approved):
Apart from the comment, lgtm!

src/GrpAb/Elem.jl
fingolfin: Yeah makes sense. (This patch grew organically and `reduced!` was only added at the very end :-))

PR #1622, opened 2024-09-21
Add `Base.conj` for `AbsSimpleNumFieldElem`

SoongNoonien:
As discussed with @fieker, @lgoettgens and @joschmitt on Friday at the OSCAR workshop I've added a `Base.conj` method for `AbsSimpleNumFieldElem`. The main purpose of this is to provide an easy way to conjugate cyclotomic numbers interactively (for actual implementations `complex_conjugation` should be better). After looking at the code I'm not sure why checking the parent field should be necessary. I think we can leave all the checks to `complex_conjugation`.

### DIFF
--- a/src/NumField/NfAbs/Conjugates.jl
+++ b/src/NumField/NfAbs/Conjugates.jl
@@ -487,67 +487,66 @@ end
 @doc raw"""
     complex_conjugation(K::AbsSimpleNumField)
 
-Given a totally complex normal number field, this function returns an
-automorphism which is the restriction of complex conjugation at one embedding.
+Given a normal number field, this function returns an automorphism which is the
+restriction of complex conjugation at one embedding.
 """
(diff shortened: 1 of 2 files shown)


thofma:
I think it is reasonable to have this, thanks. How about we cache the result of `complex_conjugation(K)` using `@attr`? Then we can remove the `!!! note`.

SoongNoonien:
I'm not very familiar with this `@attr`. Do you mean something like this: `@attr NumFieldHom{AbsSimpleNumField, AbsSimpleNumField, MapDataFromAnticNumberField{AbsSimpleNumFieldElem}, MapDataFromAnticNumberField{AbsSimpleNumFieldElem}, AbsSimpleNumFieldElem} function complex_conjugation(...`?

thofma:
I noticed that we cannot use `@attr`, since it does not work with keyword arguments. I pushed a commit adding the caching and merge once everything is green.

PR #1628, opened 2024-09-25
bump version number, in order to allow for GAP.jl 0.12

ThomasBreuer:
If Oscar.jl wants to use GAP.jl 0.12.0 then the Hecke.jl dependencies must allow GAP.jl 0.12 (see oscar-system/Oscar.jl/pull/4086).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Hecke"
 uuid = "3e1990a7-5d81-5526-99ce-9ba3ff248f21"
-version = "0.34.3"
+version = "0.34.4"
 
 [deps]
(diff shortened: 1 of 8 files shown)


thofma:
This indeed needs adjustments. What is the substitute for `julia_to_gap`?

ThomasBreuer:
One can call `GapObj` instead of `julia_to_gap`. I will adjust my pull request accordingly.
(Installing new methods for `GapObj` seems to be not needed in Hecke.jl.)

thofma:
Found it in the meantime and pushed a fix.

ThomasBreuer:
As far as I see, there is one more `GAP.julia_to_gap` call in `src/FieldFactory/CM_recognition.jl`.

thofma:
thanks, I also missed another one

PR #1631, opened 2024-09-27
Add missing Abelian groups docs

paemurru:


thofma:
Yes and no, see https://github.com/thofma/Hecke.jl/pull/1466

PR #1632, opened 2024-09-27
Add `components` method to Abelian group

paemurru:
@simonbrandhorst suggested that instead of accessing the property using the dot notation, a proper getter would be useful. I named the method `components` since this is how it is referred to in the docs.

### DIFF
--- a/docs/src/manual/abelian/elements.md
+++ b/docs/src/manual/abelian/elements.md
@@ -19,7 +19,7 @@ parent(x::FinGenAbGroupElem)
 ### Access
 
 ```@docs
-components(x::FinGenAbGroupElem)
+getindex(x::FinGenAbGroupElem, v::AbstractVector{Int})
 getindex(x::FinGenAbGroupElem, i::Int)
 ```
 
--- a/docs/src/manual/abelian/elements.md
+++ b/docs/src/manual/abelian/elements.md
@@ -19,6 +19,7 @@ parent(x::FinGenAbGroupElem)
 ### Access
 
 ```@docs
+components(x::FinGenAbGroupElem)
 getindex(x::FinGenAbGroupElem, i::Int)
 ```
 
--- a/src/GrpAb/Elem.jl
+++ b/src/GrpAb/Elem.jl
@@ -158,20 +158,10 @@ function getindex(x::FinGenAbGroupElem, v::AbstractVector{Int})
   return [x.coeff[1, i] for i in v]
 end
 
-@doc raw"""
-    firstindex(x::FinGenAbGroupElem) -> Int64
-
-Returns the first component of the element $x$.
-"""
 function Base.firstindex(x::FinGenAbGroupElem)
-  return Int64(1)
+  return Int(1)
 end
 
-@doc raw"""
-    lastindex(x::FinGenAbGroupElem) -> Int64
-
-Returns the last component of the element $x$.
-"""
 function Base.lastindex(x::FinGenAbGroupElem)
   return ngens(parent(x))
 end
--- a/src/GrpAb/Elem.jl
+++ b/src/GrpAb/Elem.jl
@@ -138,6 +138,15 @@ end
 ################################################################################
 
 @doc raw"""
+    components(x::FinGenAbGroupElem) -> Vector{ZZRingElem}
+
+Returns the components of the element $x$.
+"""
+function components(x::FinGenAbGroupElem)
+  return x.coeff
+end
+
+@doc raw"""
     getindex(x::FinGenAbGroupElem, i::Int) -> ZZRingElem
 
 Returns the $i$-th component of the element $x$.
--- a/src/GrpAb/Elem.jl
+++ b/src/GrpAb/Elem.jl
@@ -138,21 +138,42 @@ end
 ################################################################################
 
 @doc raw"""
-    components(x::FinGenAbGroupElem) -> Vector{ZZRingElem}
-
-Returns the components of the element $x$.
-"""
-function components(x::FinGenAbGroupElem)
-  return x.coeff
-end
-
-@doc raw"""
     getindex(x::FinGenAbGroupElem, i::Int) -> ZZRingElem
 
 Returns the $i$-th component of the element $x$.
 """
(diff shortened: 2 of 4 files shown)


thofma:
Yes and no, see https://github.com/thofma/Hecke.jl/pull/1466

simonbrandhorst:
I see. Maybe we just make an arbitrary default decision and go for it? Throw a dice?
We can also add 
`coefficients(::Type{ZZMatrix}, x)` 
`coefficients(::Type{Vector{ZZRingElem}, x)`

fingolfin:
What do you want to do? Knowing your usecase would help decide how to best address this. Maybe a more mathematical interface would be possible.

Note that you already can do `x[1]` to get the first, second, etc. entry.

If we really want this, we could also allow something like `Vector{ZZRingElem}(x)` ...

paemurru:
A toric geometry function outputs an algebra graded by a finitely generated abelian group. I want to make a new algebra with a slightly different grading. Namely, the last "column" replaced by ones.

Ideally I would want to be able to write `x[1:end]` to get the vector of entries. So I can construct the vector `[x[1:end-1]; 1]`.

fieker:
On Wed, Oct 09, 2024 at 05:05:18AM -0700, Erik Paemurru wrote:
> A toric geometry function outputs an algebra graded by a finitely generated abelian group. I want to make a new algebra with a slightly different grading. Namely, the last "column" replaced by ones.
> 
> Ideally I would want to be able to write `x[1:end]` to get the vector of entries. So I can construct the vector `[x[1:end-1]; 1]`.
> 
Is this math - or developing/ testing to get examples.
In the current world you "could" do
  n = ngens(parent(x))
  x - (x[n] + 1)*parent(x)[n]
to achieve that

paemurru:
> Is this math - or developing/ testing to get examples.
Math.

I think it would be rather nice if we could write `g[1:end]` to get the vector of entries, considering `g[i]` already gives an entry. Is there an easy way to add this functionality?

simonbrandhorst:
It is just about the interface. We know how to achieve what we want. But do others know it as well?
I find both of the following not optimal. The first because one has to go back to the parent and the  second because it is not using the interface .
```
julia> A = abelian_group([0,1,2,3]);

julia> x = A[1]
Abelian group element [1, 0, 0, 0]

julia> [x[i] for i in 1:ngens(A)]
4-element Vector{ZZRingElem}:
 1
 0
 0
 0

julia> x.coeff
[1   0   0   0]
```

fieker:
On Wed, Oct 09, 2024 at 07:15:12AM -0700, Erik Paemurru wrote:
> > Is this math - or developing/ testing to get examples.
> Math.
> 
> I think it would be rather nice if we could write `g[1:end]` to get the vector of entries, considering `g[i]` already gives an entry. Is there an easy way to add this functionality?
> 
well... yes no... it means adding getindex(GrpAbFinGenElem, ???)
where ??? needs thought.
 - internally, at this point, the elems are using ZZMatrix hence
   converting to a vector is bad (memory wise)
   This is mostly more efficient that Vector{ZZRingElem}
 - Tommy wants to move towards Vector{ZZRingElem} - which has other
   problems
 - looking at most usage patterns, one would like a nice name and decent
   version of
   vcat([x.coeff for x in L]...) or similar to get the coeffs into a
   matrix.
   Possibly paired by a way of converting a matrix into a lost of group
   elems
 - not completely convinced that, apart from nice, there is a strong use
   case for any of that - that justifies the work...

fieker:
On Wed, Oct 09, 2024 at 07:25:53AM -0700, Simon Brandhorst wrote:
> It is just about the interface. We know how to achieve what we want. But do others know it as well?
> I find both of the following not optimal. The first because one has to go back to the parent and the  second because it is not using the interface .
> ```
> julia> A = abelian_group([0,1,2,3]);
> 
> julia> x = A[1]
> Abelian group element [1, 0, 0, 0]
> 
> julia> [x[i] for i in 1:ngens(A)]
> 4-element Vector{ZZRingElem}:
>  1
>  0
>  0
>  0
> 
> julia> x.coeff
> [1   0   0   0]
> ```

I spend time thinking about it... and I think it is really important to
get down to the math. In general, there will be no element in the group
with some specific component set to 1. It might work here due to free
(and in particular torsion free) groups in optimal presentation.

However, in general
G([0,0,1])[3] != 1
is expected:

julia> A = abelian_group([0 0; 0 1])
Finitely generated abelian group
  with 2 generators and 2 relations and relation matrix
  [0   0]
  [0   1]

julia> A([0,1])
Abelian group element [0, 0]

simonbrandhorst:
I can see your point. As a mathematician I would do the following:
```
julia> B = abelian_group([2 2 0; 2 2 0; 1 2 3])
Finitely generated abelian group
  with 3 generators and 3 relations and relation matrix
  [2   2   0]
  [2   2   0]
  [1   2   3]

julia> b = B([7,5,31])
Abelian group element [0, 1, 40]
```
Now the goal is to express `b` as a `Z`-linear combination of the generators. 
As a mathematician I am well aware that this combination is in general not unique,
because the presentation
$$\mathbb{Z}^3 \to A$$
has a kernel.
Hence I would do the following.
```
julia> phi = hom(A,B,gens(B))
Map
  from Z^3
  to finitely generated abelian group with 3 generators and 3 relations

julia> a = preimage(phi,b)
Abelian group element [0, 1, 40]
```
And now I would wonder, how I get the coefficients `[0,1,40]`. 
Then I would try a bunch of things coming to mind e.g. 
`collect(a)` , `coefficients(a)`, `coordinates(a)`, maybe `matrix(a)` .. then
I would probably give up and type `a.coeff`.
Possibly someone would try
```
julia> F = free_module(ZZ,3)
Free module of rank 3 over ZZ

julia> hom(F,B,gens(B))
ERROR: MethodError: 
```

paemurru:
Added
```
getindex(x::FinGenAbGroupElem, v::AbstractVector{Int}) -> Vector{ZZRingElem}
```
with a note that it is inefficient.

Now the function `components` (or whatever it should be called) is not needed since we can call `x[begin:end]` instead.

thofma (approved):

PR #1633, opened 2024-10-02
Make `kernel(::SMat)` better infereable

lgoettgens:
Without this change, the compiler only infers `Union{dense_matrix_type(C), Nothing}`, as it does not know that the if/elseif clause get always taken due to the check in the beginning. Adding this explicit error  removes `Nothing` from the possible return types.

### DIFF
--- a/src/Sparse/Rref.jl
+++ b/src/Sparse/Rref.jl
@@ -205,8 +205,6 @@ function kernel(M::SMat{T}; side::Symbol = :left) where T <: FieldElement
   Solve.check_option(side, [:right, :left], "side")
   if side == :right
     return _right_kernel(M)[2]
-  elseif side == :left
-    return _left_kernel(M)[2]
   end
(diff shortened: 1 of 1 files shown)


src/Sparse/Rref.jl
```diff
@@ -208,4 +208,5 @@ function kernel(M::SMat{T}; side::Symbol = :left) where T <: FieldElement
   elseif side == :left
     return _left_kernel(M)[2]
   end
+  error("unreachable")
```
joschmitt, line 211:
```suggestion
  end
  return _left_kernel(M)[2]
```
Could we do this? Then there is less danger that at some point somebody will remove the unreachable error.

lgoettgens: This is how `solve` and friends handle this. They only have one of the cases in an if-clause, and the other below.

PR #1638, opened 2024-10-08
Update abelian group docs

paemurru:
To match with the other documentation.
Also fix one typo.


thofma:
I change the convention every other day

PR #1639, opened 2024-10-09
Fix factor

fieker:
if neccessary I can remove the spurious import. Should fix Oscar #4170

PR #1640, opened 2024-10-09
fix Oscar #4170 - too low precision in factoring

thofma:


thofma:
https://github.com/thofma/Hecke.jl/pull/1639 without the other commit

thofma:
@fieker do you have the condensed example?

fieker:
On Wed, Oct 09, 2024 at 03:08:16AM -0700, Tommy Hofmann wrote:
> @fieker do you have the condensed example?
Its still the degree 96 over quad. Don't know if we want to add it

thofma:
The `NfAbs/Elem.jl` test finished, but they took 1700 seconds compared to 230 seconds without the changes here.

fieker:
I checked out the th/facccc and it terminates on my machine in 150sec - now twice

fieker:
OK, 13sec in the previous version. I see the problem

PR #1648, opened 2024-10-11
Update .gitignore

paemurru:
Building documentation creates temporary files which should be ignored.


thofma:
Archiv, I had thos locally but forgot to open the PR. Thanks

PR #1650, opened 2024-10-17
Use set!, zero!, etc. in YET more places

fingolfin:
Also switch to Nemo.denominator! and Nemo.numerator!,
and remove add_two! which now is equivalent to `add!(x, 2)`.


lgoettgens (approved):

thofma:
I guess some of the downstream tests changed their name?

lgoettgens:
there just was another OscarDevTools.jl release that reverted this name change

thofma:
Thanks @benlorenz

PR #1653, opened 2024-10-18
Add some comments in conjugate_quad(a::AbsSimpleNumFieldElem)

fingolfin:


src/NumField/NfAbs/Elem.jl
```diff
@@ -920,13 +920,22 @@ function conjugate_quad(a::AbsSimpleNumFieldElem)
...
+    b_ptr = reinterpret(Ptr{Int}, pointer_from_objref(b))
     p_ptr = reinterpret(Ptr{Int}, k.pol_coeffs)
     s = sizeof(Ptr{ZZRingElem})
 
-    ccall((:fmpz_mul, libflint), Cvoid, (Ref{ZZRingElem}, Ptr{Int}, Ptr{Int}), q, p_ptr+s, a_ptr +s)
-    ccall((:fmpz_sub, libflint), Cvoid, (Ptr{ZZRingElem}, Ptr{ZZRingElem}, Ref{ZZRingElem}), reinterpret(Ptr{Int}, pointer_from_objref(b)), a_ptr, q)
-    ccall((:fmpz_neg, libflint), Cvoid, (Ptr{ZZRingElem}, Ptr{ZZRingElem}), reinterpret(Ptr{Int}, pointer_from_objref(b))+1*s, a_ptr + s)
-    ccall((:fmpz_set, libflint), Cvoid, (Ptr{ZZRingElem}, Ptr{ZZRingElem}), reinterpret(Ptr{Int}, pointer_from_objref(b))+3*s, a_ptr+3*s)
+    # The C type `nf_elem_t` (corresponding to `AbsSimpleNumFieldElem`) starts with
+    # an `fmpq_poly_t` (our `QQPolyRingElem`).
```
thofma, line 928: I am not sure that this is true, since the quadratic case is handled differently: https://github.com/flintlib/flint/blob/main/src/nf_elem.h#L49.

fingolfin: Indeed, it may be wrong. I'll update it once I find the time.

(Just to explain how this PR came about: I converted `ccall`s in Hecke to use the new Nemo helpers like `set!`, but the code in here really confused the hell out of me, so I tried to untangle it... at some point I decided it's not worth it right now to try to convert it, but thought that perhaps my comments would still be helpful.)

src/NumField/NfAbs/Elem.jl
```diff
@@ -919,14 +919,28 @@ function conjugate_quad(a::AbsSimpleNumFieldElem)
   b = k()
   q = ZZRingElem()
   GC.@preserve b begin
```
thofma, line 921:
```suggestion
  GC.@preserve a b k begin
```
while we are at

thofma:
thanks

PR #1655, opened 2024-10-18
feat: group algebras with sparse representation

thofma:
Due to popular demand, group algebras with sparse representation (aka "don't crash with large groups"). Needs some work/clean up:
```
julia> G = SymmetricGroup(10)
Full symmetric group over 10 elements

julia> QG = group_algebra(QQ, G)
Group algebra
  of full symmetric group over 10 elements
  over rational field

julia> a = QG(rand(G)); b = QG(rand(G));

julia> c = (a + b) * (a - b)
(1,2,4,8,7)(3,9,10,5,6) - (1,4,9)(2,10,3,6,7,5,8) + (1,7,3,8,6,10,4)(2,9,5) - (1,2,6)(3,9)(4,5)
```

@royess this might interest you. It is similar to your approach.

Small todo:
- [x] Add a `FG(::Dict)` constructor.
- [x] Add tests.
- [x] Add documentation.

### DIFF
--- a/docs/src/manual/algebras/groupalgebras.md
+++ b/docs/src/manual/algebras/groupalgebras.md
@@ -1,0 +1,49 @@
+# Group algebras
+
+```@meta
+CurrentModule = Hecke
+DocTestSetup = quote
+  using Hecke
+end
+```
+
+As is natural, the basis of a group algebra $K[G]$ correspond to the elements of $G$ with respect
+to some arbitrary ordering.
+
+## Creation
+
+```@docs
+group_algebra(::Field, ::Group)
+```
+
+## Elements
+
+Given a group algebra `A` and an element of a group `g`, the corresponding group algebra element
+can be constructed using the syntax `A(g)`.
+
+```jldoctest grpalgex1
+julia> G = abelian_group([2, 2]); a = G([0, 1]);
+
+julia> QG = group_algebra(QQ, G);
+
+julia> x = QG(a)
+[0, 0, 1, 0]
+```
+
+Vice versa, one can obtain the coordinate of a group algebra element `x` with respect to a group
+element `a` using the syntax `x[a]`.
+
+```jldoctest grpalgex1
+julia> x[a]
+1
+```
+
+It is also possible to create elements by specifying for each group element the corresponding coordinate either by a list of pairs or a dictionary:
+
+```jldoctest grpalgex1
+julia> QG(a => 2, zero(G) => 1) == 2 * QG(a) + 1 * QG(zero(G))
+true
+
+julia> QG(Dict(a => 2, zero(G) => 1)) == 2 * QG(a) + 1 * QG(zero(G))
+true
+```
--- a/src/AlgAss/AlgGrp.jl
+++ b/src/AlgAss/AlgGrp.jl
@@ -32,8 +32,8 @@ group(A::GroupAlgebra) = A.group
 
 has_one(A::GroupAlgebra) = true
 
-function (A::GroupAlgebra{T, S, R})(c::Vector{T}; copy::Bool = false) where {T, S, R}
-  length(c) != dim(A) && error("Dimensions don't match.")
+function (A::GroupAlgebra{T, S, R})(c::Union{Vector{T}, SRow{T}}; copy::Bool = false) where {T, S, R}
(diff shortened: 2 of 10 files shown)


lgoettgens:
Does this change allow me to work with group algebras of infinite groups as well?

thofma:
Yes:
```
julia> F = free_group(2)
Free group of rank 2

julia> x, y = gens(F);

julia> QF = group_algebra(QQ, F; sparse = true)
Group algebra
  of free group of rank 2
  over rational field

julia> (QF(x) + QF(y))^2
f1^2 + f1*f2 + f2*f1 + f2^2
```

Fe-r-oz:
This is wonderful! I will shift to `sparse = true` as soon as this is merged! Thank you.

src/AlgAss/Types.jl
```diff
@@ -130,6 +130,13 @@ end
 @attributes mutable struct GroupAlgebra{T, S, R} <: AbstractAssociativeAlgebra{T}
   base_ring::Ring
   group::S
+  # We represent elements using a coefficient vector,
```
fingolfin, line 133:
```suggestion
  # If `sparse` is false we represent elements using a coefficient vector,
```

src/AlgAss/Types.jl
```diff
@@ -226,8 +272,22 @@ mutable struct GroupAlgebraElem{T, S} <: AbstractAssociativeAlgebraElem{T}
...
+    z.coeffs_sparse = coeffs
+    return z
+  end
 end
 
+__elem_index(A, g) = get!(A.group_to_base, g) do
+        A.ind += 1
+        resize!(A.base_to_group, max(A.ind, length(A.base_to_group)))
+        A.base_to_group[A.ind] = g
+        return A.ind
```
fingolfin, line 288: So when would `length(A.base_to_group)` ever exceed `A.ind`? Right now I am wondering why you even need `A.ind` and can't just do this:
```suggestion
        push!(A.base_to_group, g)
        return length(A.base_to_group)
```

src/AlgAss/Types.jl
```diff
@@ -146,77 +153,116 @@ end
...
+        end
+        A.group_to_base[el] = 1
+        A.base_to_group[1] = el
+        A.sparse_one = sparse_row(K, [1], [one(K)])
+        A.ind = 1 # index - 1 for the next group element
       end
-      v[1] = one(K)
 
-      A.one = v
+      if !A.sparse
```
fingolfin, line 199: Merge this with preceding `end` into an `else`?

fingolfin (commented):
Great! Some quick comments and questions (all nitpicks, not meant to block anything)

src/AlgAss/Types.jl
thofma: I mean vector in the mathematical sense. They comment there is valid both for sparse and non-sparse representations.

src/AlgAss/Types.jl
thofma: Yes, I think I wanted to do work with A.base_to_group thing differently, but it does not make any sense now. Will do as you suggest.

docs/src/manual/algebras/groupalgebras.md
```diff
@@ -0,0 +1,69 @@
...
+```
+
+It is also possible to create elements from dictionaries:
+
+```jldoctest
+julia> G = abelian_group([2, 2]); a = G([0, 1]);
+
+julia> QG = group_algebra(QQ, G);
+
+julia> QG(Dict(a => 2, zero(G) => 1)) == 2 * QG(a) + 1 * QG(zero(G))
```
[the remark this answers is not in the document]
thofma: added

src/AlgAss/AlgGrp.jl
```diff
@@ -59,17 +59,25 @@ end
 ################################################################################
 
 @doc raw"""
-    group_algebra(K::Ring, G; op = *) -> GroupAlgebra
+    group_algebra(K::Ring, G::Group; sparse::Bool = false,
```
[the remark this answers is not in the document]
thofma: I got rid of it by adding another layer of indirection.

thofma:
Thanks for the comments. I think I addressed them all. I also removed the `sparse` option from the user-facing version. It is now chosen automatically (as in Magma).

PR #1661, opened 2024-10-25
fix: saturation at large primes

thofma:
Use baby-step-giant-step algorithm for large primes to make
the complexity not linear in the order.

Before (saturating at ~2^25):
```julia
96.520001 seconds (56.86 k allocations: 120.018 GiB, 17.87% gc time)
```
With this PR:
```julia
2.726522 seconds (70.38 M allocations: 1.101 GiB, 30.02% gc time)
```

### DIFF
--- a/src/NumFieldOrd/NfOrd/Clgp/Saturate.jl
+++ b/src/NumFieldOrd/NfOrd/Clgp/Saturate.jl
@@ -12,10 +12,11 @@ function mod_p(R::Vector{FacElem{AbsSimpleNumFieldElem, AbsSimpleNumField}}, Q::
   pp, e = Hecke.ppio(oF, p)
   #We now want the discrete logarithm of the images of the elements of R in F
   #We don't need the full group, just the quotient by p
-  #We compute a generator and cache the powers in dl
-  #Then we will compute the discrete logarithms by checking the values in dl
+  #We compute a generator and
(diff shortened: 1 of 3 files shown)


src/NumFieldOrd/NfOrd/Clgp/Saturate.jl
```diff
@@ -28,12 +29,21 @@ function mod_p(R::Vector{FacElem{AbsSimpleNumFieldElem, AbsSimpleNumField}}, Q::
...
+    for i = 0:pp-1
+      dl[y] = i
+      y *= x
+    end
+  end
+  ma = Vector{Int}(undef, length(R))
+  for i in 1:length(R)
+    imgd = image(mF1, R[i], D[i], cached, pp)^e
+    ma[i] = get!(dl, imgd) do
+      ma[i] =  Hecke.baby_step_giant_step(x, pp, imgd, dl) % p
```
fieker, line 43: it feels like one of the `ma[i] = ` should be superflous. dies the `get` also update the `dl` cache?

thofma: oops, you are right

thofma:
@fieker happy?

fieker (approved):

PR #1663, opened 2024-10-27
Fix @GC.preserve -> GC.@preserve

fingolfin:
This old syntax has been deprecated for a long time

### DIFF
--- a/examples/IsPower.jl
+++ b/examples/IsPower.jl
@@ -312,11 +312,11 @@ function Base.getindex(H::Hecke.HenselCtx, s::Symbol, i::Int)
     return unsafe_load(H.link, i)
   elseif s == :v
     f = Hecke.Globals.Zx()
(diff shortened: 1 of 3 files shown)


examples/IsPower.jl
```diff
@@ -312,11 +312,11 @@ function Base.getindex(H::Hecke.HenselCtx, s::Symbol, i::Int)
     return unsafe_load(H.link, i)
   elseif s == :v
     f = Hecke.Globals.Zx()
-    @GC.preserve f H ccall((:fmpz_poly_set, Nemo.libflint), Cvoid, (Ref{ZZPolyRingElem}, Ptr{Hecke.fmpz_poly_raw}), f, H.v+(i-1)*sizeof(Hecke.fmpz_poly_raw))
+    GC.@preserve f H ccall((:fmpz_poly_set, Nemo.libflint), Cvoid, (Ref{ZZPolyRingElem}, Ptr{Hecke.fmpz_poly_raw}), f, H.v+(i-1)*sizeof(Hecke.fmpz_poly_raw))
```
fieker, line 315: Is there actually a point in using preserve on a single ccall? (I know its my code, but..)

thofma: yes, I think so

PR #1665, opened 2024-10-29
Remove comment and add new ambiguous example

StevellM:
Follows  #1664

We can remove the comment now, I have checked what was needed. I found another ambiguous example in my data, so I added it.
While playing with data, I found a misbehaving argument check which did not take in account the case of genus enumeration in the non full rank case for binary form over QQ.


simonbrandhorst (approved):
Thanks.

thofma:
Can you add the example of the non-full rank lattices?

PR #1668, opened 2024-10-30
Don't export `coprime_base_insert` & `subalgebra`

fingolfin:
They don't exist


thofma:
downstream tests slowly turning useless

PR #1670, opened 2024-11-01
feat: subgroup enumeration with no subgroups

thofma:


thofma:
P.S.: Before
```
julia> G = abelian_group([2]);

julia> length(collect(subgroups(G; quotype = [3])))
1
```

PR #1675, opened 2024-11-06
Update Hecke banner

paemurru:
I only recently found out that Hecke has its own manual. I included the manual link in the banner, similarly to the Julia and OSCAR banners.


thofma:
I did not count, but does this fit a terminal with 80 characters?

paemurru:
Yes, it is 73 characters wide

PR #1677, opened 2024-11-08
Pretty printing for elliptic curves and their points

simonbrandhorst:


thofma:
I think of "short" form should not contain line breaks? I vaguely remember something like that

simonbrandhorst:
You are right. Did I overlook a linebreak somewhere?

src/EllCrv/EllCrv.jl
```diff
@@ -778,8 +778,23 @@ end
...
+  print(io, Indent())
+  _print_equation(io, E)
+  print(io, Dedent())
+end
+
 function show(io::IO, E::EllipticCurve)
-  print(io, "Elliptic curve with equation\n")
+  io = pretty(io)
+  print(io, "Elliptic curve with equation ")
+  print(io, Indent())
```
thofma, line 792: I think I got confused by this. What does this do without linebreak?

simonbrandhorst: Nothing (useful). I overlooked it. Thanks for pointing it out. Removed it.

thofma:
This 'breaks' the booktests, since the output changed (https://github.com/thofma/Hecke.jl/actions/runs/11754631063/job/32757483758?pr=1677#step:7:19595). So this change becomes breaking unfortunately. I will label it as such and we merge it once we want to make a breaking release.

StevellM:
@simonbrandhorst Why don't you make the base field part of the detailed printing ? Since a priori one could define elliptic curves over any (nice?) field, wouldn't it sense to tell whether it is defined over $\mathbb{Q}$ or $\mathbb{F}_{121}$.

simonbrandhorst:
We could do that. But I don't want to fix the doctests again xD.

thofma:
What would the right order be? First the equation or first the base field?

simonbrandhorst:
I would put the base field first. 
```
Affine algebraic set
  in affine 2-space over QQ with coordinates [x1, x2]
defined by ideal (x1*x2)
```

thofma:
How about the following? If we agree, I can push the changes.
```
julia> E = elliptic_curve(GF(11), [1, 2])
Elliptic curve
  over prime field of characteristic 11
with equation
  y^2 = x^3 + x + 2

julia> E = elliptic_curve(K, [1, 2])
Elliptic curve
  over number field of degree 4 over QQ
with equation
  y^2 = x^3 + x + 2
```

StevellM:
It looks good like that!

PR #1678, opened 2024-11-08
Improve README example

paemurru:
Change FlintQQ to QQ.


thofma:
Thanks, but it seems that was contained in #1674

PR #1680, opened 2024-11-11
Remove strange checkbounds method

fingolfin:
It seems to be unused


fieker:
I thin that was an attempt to support @inbounds/ boundscheck in Julia in Nemo/Hecke.
No idea how far we are in that direction

PR #1684, opened 2024-11-14
fix: don't use polynomial rings over zero rings

thofma:


src/NumField/NfAbs/Elem.jl
```diff
@@ -157,6 +157,8 @@ end
 
 #In this version, n is supposed to be a prime power
 function is_norm_divisible_pp(a::AbsSimpleNumFieldElem, n::ZZRingElem)
+  Main.a = a
+  Main.n = n
```
lgoettgens, line 161: Just a reminder to remove these debugging artifacts again once you are finished tracking down all uses of zero rings

PR #1685, opened 2024-11-18
fix: group algebras of additive groups with sparse rep

lgoettgens:
These were the necessary changes to make (sparse) group algebras with an (infinite) additive group that is not `FinGenAbGroup`.

To find these things, I defined a thin wrapper around `FinGenAbGroup` and tried to make that work with the existing things.

```julia
struct FooGroup <: AbstractAlgebra.AdditiveGroup
  G::FinGenAbGroup
end

struct FooGroupElem <: AbstractAlgebra.AdditiveGroupElem
  x::FinGenAbGroupElem
end

elem_type(::Type{FooGroup}) = FooGroupElem

parent_type(::Type{FooGroupElem}) = FooGroup

parent(x::FooGroupElem) = FooGroup(parent(x.x))

is_abelian(G::FooGroup) = is_abelian(G.G)

is_finite(G::FooGroup) = is_finite(G.G)

order(T, G::FooGroup) = order(T, G.G)

order(G::FooGroup) = order(G.G)

length(G::FooGroup) = length(G.G)

function Base.iterate(G::FooGroup)
  it = iterate(G.G)
  isnothing(it) && return nothing
  x, state = it
  return FooGroupElem(x), state
end

function Base.iterate(G::FooGroup, state)
  it = iterate(G.G, state)
  isnothing(it) && return nothing
  x, state = it
  return FooGroupElem(x), state
end

zero(G::FooGroup) = FooGroupElem(zero(G.G))

iszero(x::FooGroupElem) = iszero(x.x)

+(x::FooGroupElem, y::FooGroupElem) = FooGroupElem(x.x + y.x)

Base.hash(x::FooGroupElem, h::UInt) = hash(x.x, h)

Base.:(==)(x::FooGroupElem, y::FooGroupElem) = x.x == y.x

Base.rand(G::FooGroup) = FooGroupElem(rand(G.G))

export FooGroup, FooGroupElem
```

PR #1688, opened 2024-11-19
Update show for GroupAlgebraElem with sparse rep

lgoettgens:
Let me show the current issue with an example. To reproduce, you need the branch https://github.com/oscar-system/Oscar.jl/tree/lg/character-ring together with a dev'ed Hecke that includes https://github.com/thofma/Hecke.jl/pull/1685.

```julia
julia> R = root_system(:A, 2);

julia> P = weight_lattice(R); # this is basically ZZ^2, but with nice printing

julia> ZP = group_algebra(ZZ, P);

julia> z = zero(ZP)
0

julia> o = one(ZP)
0

julia> z == o # but they print the same
false

julia> 42 * one(ZP) # that should be something like `42 * 0` or `42 * e(0)`
0

julia> a = ZP(gen(P, 1)) + ZP(gen(P, 2))
w_1 + w_2

julia> b = ZP(gen(P, 1) + gen(P, 2))
w_1 + w_2

julia> a == b # but they print the same
false
```
Thus I propose to instead of just writing the group elem as the basis elem, to do something like `e(g)` as often done in the literature. With this change enabled, the problematic lines from above change to the following, which IMO is a clear readability and disambiguation improvement.
```julia
julia> z = zero(ZP)
0

julia> o = one(ZP)
e(0)

julia> 42 * one(ZP) # that should be something like `42 * 0` or `42 * e(0)`
42*e(0)

julia> a = ZP(gen(P, 1)) + ZP(gen(P, 2))
e(w_1) + e(w_2)

julia> b = ZP(gen(P, 1) + gen(P, 2))
e(w_1 + w_2)
```

### DIFF
--- a/src/AlgAss/Elem.jl
+++ b/src/AlgAss/Elem.jl
@@ -752,7 +752,7 @@ function show(io::IO, a::AbstractAssociativeAlgebraElem)
         for (i, ci) in a.coeffs_sparse
           push!(sum.args,
                 Expr(:call, :*, AbstractAlgebra.expressify(ci, context = io),
-                                AbstractAlgebra.expressify(parent(a).base_to_group[i], context = IOContext(io, :compact => true))))
+                                Expr(:call, :e, AbstractAlgebra.expressify(parent(a).base_to_group[i], context = IOContext(io, :compact => true)))))
         end
       end
       print(io, AbstractAlgebra.expr_to_string(AbstractAlgebra.canonicalize(sum)))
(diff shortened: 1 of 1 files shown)


fingolfin:
Seems sensible to me

fingolfin:
@thofma any thoughts on this PR?

thofma:
Have not made up my mind yet unfortunately.

PR #1689, opened 2024-11-20
Fix show method for pseudo-matrices

SirToby25:
Previously, generating a pseudo-matrix over an algebraic number field with zero rows was possible, e.g.,

`K, _ = quadratic_field(-5)`
`P = pseudo_matrix(identity_matrix(K, 0), [])`

and the pseudo-matrix P was created correctly. However, an error was raised, due to an undefined reference and this P could not be printed. This commit just adds an extra output for pseudo-matrices with zero rows.


thofma:
Thanks

PR #1691, opened 2024-11-22
is_empty for pseudo-matrices

SirToby25:
It seems natural to have an 'is_empty'-method for pseudo-matrices, so this commit adds one. I also added a couple of tests.


thofma:
makes sense, thanks

thofma:
merci

PR #1693, opened 2024-11-26
Minor cleanup related to FacElemMon and FacElem

fingolfin:
FacElemMon is a monoid, not a ring


thofma:
I wonder if there was a reason for this. Do you remember @fieker?

PR #1703, opened 2024-12-13
Bump compat of AA and Nemo

lgoettgens:


lgoettgens:
The failure in `Run long tests` is due to an old version of `GAP.jl` getting used here, as the current release is not compatible with the new AA yet (waiting for backport of https://github.com/oscar-system/GAP.jl/pull/1091 to the 0.12.x release series). We should just wait for that right now and then restart the test.

lgoettgens:
@thofma CI is happy now that the new GAP.jl is released.

PR #1707, opened 2024-12-19
Add some ring / field conformance tests

fingolfin:
... and fix some issues this revealed. In particular "missing" `base_ring` methods
which the conformance tests kind of assume, but here in some cases "only" `base_field` methods were available.

Contains #1706 because of overlap.

### DIFF
--- a/src/LocalField/LocalField.jl
+++ b/src/LocalField/LocalField.jl
@@ -37,6 +37,8 @@ prime(K::LocalField) = prime(base_field(K))
 #
 ################################################################################
 
+base_ring_type(::Type{LocalField{S, T}}) where {S <: FieldElem, T <: LocalFieldParameter} = parent_type(S)
+
 base_field_type(::Type{LocalField{S, T}}) where {S <: FieldElem, T <: LocalFieldParameter} = parent_type(S)
 
 elem_type(::Type{LocalField{S, T}}) where {S <: FieldElem, T <: LocalFieldParameter} = LocalFieldElem{S, T}
@@ -47,8 +49,8 @@ elem_type(::Type{LocalField{S, T}}) where {S <: FieldElem, T <: LocalFieldParame
 #
 ################################################################################
 
-is_domain_type(::Type{S}) where S <: LocalField = true
-is_exact_type(::Type{S}) where S <: LocalField = false
+is_domain_type(::Type{<: LocalFieldElem}) = true
+is_exact_type(::Type{<: LocalFieldElem}) = false
 isfinite(K::LocalField) = isfinite(base_field(K))
 
 ################################################################################
@@ -129,6 +131,10 @@ end
 #  Subfields
 #
 ################################################################################
+
+function base_ring(L::LocalField)
+  return base_ring(defining_polynomial(L))
+end
 
 function base_field(L::LocalField)
   return base_ring(defining_polynomial(L))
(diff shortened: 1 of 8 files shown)


thofma:
I probably misunderstood something, but `base_ring` is not part of the ring interface. How can a missing `base_ring` method make the conformance tests fail?

src/LocalField/LocalField.jl
```diff
@@ -134,6 +132,10 @@ end
 #
 ################################################################################
 
+function base_ring(L::LocalField)
```
fieker, line 135: shouldn't this be `base_ring(L::LocalField) = base_field(L)` instead?
But `base_ring` was never (intended to be) part of any useful interface, it was just indicating, in recursively defined rings, the previous step. Strictly here, `base_ring` might even have been the polynomial ring...
`base_field`, `coefficient_ring/field` were intended to be more useful

test/LocalField/LocalField.jl
```diff
@@ -1,3 +1,16 @@
+# the following is copied from the `random_elem` function
+function test_elem(L::Union{QadicField, Hecke.LocalField})
+  b = basis(L)
+  n = degree(L)
+  r = [rand(1:5*n) for i in 1:n]   # Choose small coordinates
+  return sum( [r[i]*b[i] for i in 1:n])
+end
+
+# TODO/FIXME: implement isapprox so we can get rid of the following HACK:
```
fieker, line 9: why? From the docs:
### Approximation (floating point and ball arithmetic only)

```julia
isapprox(f::MyElem, g::MyElem; atol::Real=sqrt(eps()))
```

so it is not part of the normal ring/field interface I'd say

fingolfin:
No I don't think you missed anything -- the general conformance tests should not require `base_ring`. So this is a bug in them. It may still makes sense to test it if available (specifically: if `base_ring` is there, then so should `base_ring_type`, and they should agree).

We'll fix the conformance test suite, and once that's there, I can adjust this PR (there are still a few valid changes in here, like a missing `hash` function; and there are also some more rings defined in Hecke for which it would be IMHO useful to run conformance tests)

test/Misc/RelFinField.jl
```diff
@@ -1,5 +1,12 @@
 @testset "RelFinField" begin
 
+  @testset "conformance" begin
+    F = Native.finite_field(3, 3, cached = false)[1]
```
fieker, line 4: @thofma I thought we'd rather get rid of the RelFinField?

thofma: I only replaced the ones in the prime decomposition of relative number fields. They might be used somewhere else still.

thofma:
Thanks for the clarification. I am also happy to help with missing methods.

test/LocalField/LocalField.jl
fingolfin: I never really questioned the use of `isapprox` in the conformance tests, I assumed that Dan had checked this was the right thing to use when he added them.

I'd be happy to use something else then, but what? We currently have this helper in the conformance tests:
```julia
# helper
function equality(a, b)
   if is_exact_type(typeof(a)) && is_exact_type(typeof(b))
      return a == b
   else
      return isapprox(a, b)
   end
```

We use it to implement tests like these:
```
            @test equality(a^4, a*a*a*a)
            @test equality((a + b) + c, a + (b + c))
            @test equality(a + b, b + a)
            @test equality(a - c, a + (-c))
```
What would be a way to compare two `QAdicFieldElem` in a way that would satisfy such tests that deals with the problem that e.g. `a^4` and `a*a*a*a` may produce slightly different results (e.g. perhaps `a^4` is "more" accurate)

thofma: `==`

PR #1713, opened 2024-12-27
Small fix of is_divisible (and divexact) + tests

SirToby25:
Given an instance `A` of a subtype of `Hecke.AbstractAssociativeAlgebra{T}`, trying

```julia

divexact(zero(A), zero(A))

```
would result in a `BoundsError` instead of the usual `ErrorException`, due to `r == 0` in the Elem.jl.

This commit fixes this an also adds two small tests.

### DIFF
--- a/src/AlgAss/Elem.jl
+++ b/src/AlgAss/Elem.jl
@@ -463,7 +463,7 @@ function is_divisible(a::AbstractAssociativeAlgebraElem, b::AbstractAssociativeA
   Ma = hcat(M, va)
   r = rref!(Ma)
 
-  if all(iszero, [ Ma[r, i] for i = 1:dim(A) ])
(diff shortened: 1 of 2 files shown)


thofma:
Hm, I think the answer for `is_divisible(zero(A), zero(A))` should be `true` and `divexact(zero(A), zero(A))` should not throw.

SirToby25:
I agree that `is_divisible(zero(A), zero(A))` should return `true`.

However, the following already throw a DivideError, so I think `divexact(zero(A), zero(A))` throwing would be preferable for consistency.
```julia
divexact(zero(ZZ), zero(ZZ))
divexact(zero(QQ), zero(QQ))
```

thofma:
Right, let's keep it consistent for now.

PR #1715, opened 2025-01-01
fix: minor space typo in the `AbsAlgAssMor` output after decomposition of `group_algebra`

Fe-r-oz:
It seems there is a minor space typo as there is no space after order 5 in the output below:
```
julia> C = abelian_group(5);

julia> GA = group_algebra(GF(3), C);

julia> d = decompose(GA);

julia> d[1][2]
Map
  from structure constant algebra of dimension 1 over GF(3)
  to group algebra of group of order 5over GF(3)
```

I think the first output line is printed correctly from: https://github.com/thofma/Hecke.jl/blob/3f9f55b0c1e1c2c013f3ecd60d0000aba86d121a/src/AlgAss/StructureConstantAlgebra.jl#L362-L363

It seems the typo is due to this line, no space before over, it should be `" over " `as in the earlier clause

https://github.com/thofma/Hecke.jl/blob/3f9f55b0c1e1c2c013f3ecd60d0000aba86d121a/src/AlgAss/AlgGrp.jl#L185

Please help review this PR, thank you!

### DIFF
--- a/src/AlgAss/AlgGrp.jl
+++ b/src/AlgAss/AlgGrp.jl
@@ -178,11 +178,11 @@ function show(io::IO, A::GroupAlgebra)
   else
     print(io, "Group algebra of group ")
     if is_finite(group(A))
-      print(io, "of order ", order(group(A)))
+      print(io, "of order ", order(group(A)), " ")
     else
       print(io, "of infinite order ")
     end
-    print(io, " over ")
+    print(io, "over ")
     print(terse(io), base_ring(A))
   end
 end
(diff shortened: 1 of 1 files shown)


lgoettgens (changes requested):
This unfortunately creates a double-space in the case of an infinite group. To resolve this, I think the cleanest solution would be to revert your change here and instead add a trailing space to the print in line 181

Fe-r-oz:
Thank you for your suggestion! Hopefully this works okay.

PR #1717, opened 2025-01-07
Allow GAP.jl 0.13

lgoettgens:
Draft as GAP.jl 0.13.0 is not released yet.


lgoettgens:
GAP.jl 0.13.0 is now released. @thofma could you please merge and release this bump, so we can proceed with https://github.com/oscar-system/Oscar.jl/pull/4421

Cc @fingolfin

PR #1718, opened 2025-01-08
fix: aliasing bug in mul! of LocalFieldElem

thofma:


fieker (approved):
how long did it take to find this? Well done!

PR #1742, opened 2025-02-10
Explicitly qualify constructors that are extended

lgoettgens:
to silence the warnings introduced in https://github.com/JuliaLang/julia/pull/57311.


thofma:
merci

PR #1746, opened 2025-02-12
feat: swap type parameter order for orders in algebras

thofma:
@fieker now it is the same as for `GenOrd`. First the field/algebra, then the "base ring".


fieker (approved):
if you're happy..

thofma:
if it is bad, it should be consistently bad

PR #1755, opened 2025-02-15
fix: corect argument order for minpoly/charpoly

### LINKED ISSUE #1727: Argument order in `minpoly` and `charpoly` methods
Hecke defines multiple `minpoly` and `charpoly` methods with two arguments but in the wrong order, e.g.:
```julia
function minpoly(a::AbsNumFieldOrderElem, Zx::ZZPolyRing = ZZPolyRing(ZZ, :x, false))
  return Zx(minpoly(elem_in_nf(a)))
end
```
or
```julia
function minpoly(a::T, Rx::PolyRing = polynomial_ring(base_field(parent(a)), "x", cached = false)[1]) where T <:RelFinFieldElem
...
```

Correct would be to have the polynomial ring first.

To address this in a backwards compatible I suggest the following:

1. We add two methods like this (possibly with a deprecation warning in it?)
```
minpoly(a::RingElem, R::PolyRing) = minpoly(R, a)
charpoly(a::RingElem, R::PolyRing) = charpoly(R, a)
```
2. we warn in the release notes that these may go one day?
3. we add the missing functions, e.g..
```julia
function minpoly(Zx::ZZPolyRing, a::AbsNumFieldOrderElem)
  return Zx(minpoly(elem_in_nf(a)))
end
```
4. If we implement https://github.com/oscar-system/Oscar.jl/issues/4518 (CC @johnaabbott) before we are now done, otherwise also add a single-arg fallback

Of course code in Hecke using the "wrong order" should ideally also be adjusted but that's a separate concern...

thofma:
- closes #1727
- implements the strategy of #1727


thofma:
@fingolfin is this what you had in mind?

src/Deprecations.jl
```diff
@@ -280,3 +280,7 @@ end
 @deprecate prime_field(L::Union{QadicField, LocalField}) absolute_base_field(L)
 @deprecate coefficient_ring(k::LocalField) base_field(k)
 @deprecate coefficient_field(k::QadicField) base_field(k)
+
+# Deprecated in 0.35.10
+@deprecate minpoly(a::RingElem, R::PolyRing) minpoly(R, a)
+@deprecate charpoly(a::RingElem, R::PolyRing) charpoly(R, a)
```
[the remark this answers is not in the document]
thofma: added less generic versions

src/Deprecations.jl
```diff
@@ -280,3 +280,8 @@ end
 @deprecate prime_field(L::Union{QadicField, LocalField}) absolute_base_field(L)
 @deprecate coefficient_ring(k::LocalField) base_field(k)
 @deprecate coefficient_field(k::QadicField) base_field(k)
+
+# Deprecated in 0.35.10
+@deprecate minpoly(a::Union{LocalFieldElem, QadicFieldElem, RelFinFieldElem, AbsNumFieldOrderElem}, R::PolyRing) minpoly(R, a)
+@deprecate minpoly(a::AbsNumFieldOrderElem, R::PolyRing) minpoly(R, a)
+@deprecate charpoly(a::RingElem, R::PolyRing) charpoly(R, a)
```
lgoettgens, line 287: for charpoly holds the same, i.e. better to use a union here. and for minpoly, the single one in line 286 is also contained in the union the line above

thofma: Oops, that was a typo. For `minpoly` I only want `AbsSNumFieldOrderElem`.

PR #1756, opened 2025-02-16
fix: add missing hash functions

thofma:


src/AlgAssAbsOrd/Ideal.jl
```diff
@@ -1503,7 +1503,11 @@ function Base.one(S::AlgAssAbsOrdIdlSet)
   return ideal(order(S), one(order(S)))
 end
 
- ==(x::AlgAssAbsOrdIdlSet, y::AlgAssAbsOrdIdlSet) = x.order === y.order
+==(x::AlgAssAbsOrdIdlSet, y::AlgAssAbsOrdIdlSet) = x.order === y.order
+
+function Base.hash(x::AlgAssAbsOrdIdlSet, h::UInt)
+  return xor(hash(order(x), h), 0x3d4177ad98877e36 % UInt)
```
lgoettgens, line 1509:
```suggestion
  return hash(objectid(x.order), h)
```
since the `==` delegates to `===` on `.order`, we can make the hash stronger here by using `objectid` (which is basically the hash wrt `===`)

src/NumField/CM.jl
```diff
@@ -60,6 +60,10 @@ end
 Base.:(==)(f::CMType, g::CMType) = (f.field === g.field) &&
   issetequal(f.embeddings, g.embeddings)
 
+# this is not optimal, we would need to sort the embeddings on creation using
+# _absolute_index, but do we want this?
+Base.hash(f::CMType, h::UInt) = h
```
lgoettgens, line 65:
```suggestion
Base.hash(f::CMType, h::UInt) = hash(f.field, h)
```
You can include the field here safely. Regarding the embeddings, you could just add `Set(f.embeddings)` to the hash, as that hashes the elements without any order (so the perfect analog to `issetequal`)

src/NumFieldOrd/NfOrd/Ideal/Ideal.jl
```diff
@@ -68,6 +68,11 @@ parent_type(::Type{AbsNumFieldOrderIdeal{AbsSimpleNumField, AbsSimpleNumFieldEle
 
 ==(x::AbsNumFieldOrderIdealSet, y::AbsNumFieldOrderIdealSet) = x.order === y.order
 
+function Base.hash(x::AbsNumFieldOrderIdealSet, h::UInt)
+  h = hash(x.order, h)
+  return xor(h, 0x5d49d4dc68dec3e2 % UInt)
```
lgoettgens, line 73: As for `AlgAssAbsOrdIdlSet`, this could use `objectid(x.order)` for a stronger hash

src/QuadForm/Quad/Spaces.jl
```diff
@@ -2357,6 +2366,12 @@ function Base.:(==)(G1::QuadSpaceCls, G2::QuadSpaceCls)
   return all(local_symbol(G1, p) == local_symbol(G2,p) for p in P)
 end
 
+function Base.hash(G::QuadSpaceCls, h::UInt)
+  h = hash(dim(G), h)
```
lgoettgens, line 2370:
```suggestion
  h = hash(base_ring(G), h)
  h = hash(dim(G), h)
```
the base_ring is used in the first line of `==`, so why not use it here as well

test/AlgAssAbsOrd/Ideal.jl
```diff
@@ -199,4 +199,12 @@
   I = 1 * O
   @test Hecke.is_full_lattice(I)
 
+  # parent
+  let
+    A = zero_algebra(QQ)
+    O = Order(A, elem_type(A)[])
+    I = 1 * O
+    @test parent(I) == parent(I)
+    @test hash(parent(I)) == hash(parent(I))
```
lgoettgens, line 208: what is this supposed to test? Maybe you meant `O` instead of `I` for some of them?

lgoettgens (approved):

src/AlgAssAbsOrd/Ideal.jl
thofma: Isn't this the default hash anyway?

src/NumField/CM.jl
thofma: Have you looked at the hashing of `Set`? I looks pretty wrong to me.

test/AlgAssAbsOrd/Ideal.jl
thofma: hashing for `AlgAssAbsOrdIdlSet`

src/AlgAssAbsOrd/Ideal.jl
lgoettgens: yes, the `hash` fallback delegated to `objectid`. But here, if at some point there is a specific hash function for the type of `x.order`, this would be used. This would both be slower than just `objectid` and it would produce a weaker hash

src/NumField/CM.jl
lgoettgens: Why? it hashes all elements on their own and then xors them. And xor is commutative, so order shouldn't matter

test/AlgAssAbsOrd/Ideal.jl
lgoettgens: But isn't `parent(I) === parent(I)`? (and thus these tests basically useless) If not, sorry for my confusion

src/NumField/CM.jl
thofma: ah, I missed the commutativity

test/AlgAssAbsOrd/Ideal.jl
thofma: Hm no, it is not:
```
julia> parent(I) == parent(I)
true

julia> parent(I) === parent(I)
false
```
I am not super happy with the design of some of these things, but I don't want to change it here.

PR #1762, opened 2025-02-19
feat: add maximal tori for algebras

thofma:
(at the moment only in the "small base field" case)


src/AlgAss/Torus.jl
```diff
@@ -0,0 +1,139 @@
...
+
+# Compute the subalgebra generated by B (note that subalgebras have the same
+# identitet as the overalgebra).
+function _subalgebra_basis(A, B::Vector, iscom = true)
+  @assert iscom
+  push!(B, one(A))
+  closed = false
+  while !closed
+    bmat = echelon_form(basis_matrix(B); trim = true)
+    B = [elem_from_mat_row(A, bmat, i) for i in 1:nrows(bmat)]
```
thofma, line 79: @wfsteiner this is similar to what I wrote on paper today

PR #1766, opened 2025-02-21
optimize the sparse HNF, some startegy, but mostly gc friendly

fieker:
- add temporary scalar variables in SMat
 - use (and define) more ! functions
 - careful memory management in ZZRow
 - improve sub, transpose, ..
 - add strategy in HNF to alternate between full HNF and only produce some upper triangular matrix
 - better support for getting the trafo by appending the identity


thofma:
there is some conflict

src/Sparse/HNF.jl
```diff
@@ -123,35 +150,48 @@ function reduce(A::SMat{T}, g::SRow{T}, m::T) where {T}
...
     end
-    st_g = 2
-    st_A = 2
-    p = g.values[1]
-    if divides(p, A.rows[j].values[1])[1]
-      g = add_scaled_row(A[j], g, - divexact(p, A.rows[j].values[1]))
+    
+    p_v = getindex!(p_v, g.values, 1)
+    A_v = getindex!(A_v, A.rows[j].values, 1)
+    fl, g_v = divdes!(g_v, p_v, A_v)
```
thofma, line 182:
```suggestion
    fl, g_v = divides!(g_v, p_v, A_v)
```

fieker:
On Tue, Feb 25, 2025 at 06:39:55AM -0800, Tommy Hofmann wrote:
> thofma left a comment (thofma/Hecke.jl#1766)
> 
> from the tests I would guess that some function call does not finish
Indeed - I'm looking at it now. For me it was s.w. in genera of quad
forms...

thofma:
they are all computing class groups I believe, so it might be related?

src/Sparse/ZZRow.jl
```diff
@@ -86,11 +86,17 @@ function Base.iterate(a::ZZRingElem_Array, state::Int = 1)
 end
 
 function Base.sizehint!(a::ZZRingElem_Array, i::Int)
+  return sizehint!(a.ar, i)
```
thofma, line 89: Remove the rest below?

fieker:
There is more unused code. I'm not sure I want to remove it in particular I
do not understand what size hint does. I suspect it guarantees nothing

PR #1770, opened 2025-02-24
feat: allow inseparable algebras when decomposing

wfsteiner:
cc @thofma


thofma (commented):
Looks good, thanks.

I added a comment on how to make the decomposition fully functional. You can test it with the example you added. (You can also cook up examples using `direct_product`.)

PR #1771, opened 2025-02-24
Remove most uses of gap_to_julia

lgoettgens:
Some progress towards https://github.com/oscar-system/Oscar.jl/issues/4625.

The remaining three calls are in the meataxe code, and are present in `Oscar.jl/experimental/GModule/src/GModule.jl` as well. Those need probably the same changes.

I am not quite sure if such a strict restrictions of GAP.jl versions is necessary, but I am pretty certain that 0.9.6 won't work any more with all of the changes regarding conversions last year. If there is a good reason to keep older versions in there, I can try to test a few combinations, but otherwise I won't spend the time.

cc @fingolfin


thofma:
Don't worry about the version.

PR #1773, opened 2025-02-24
Use Requires.jl for package extensions pre-1.9

lgoettgens:
This enables some code deduplication in Oscar, see https://github.com/oscar-system/Oscar.jl/pull/4635. https://github.com/thofma/Hecke.jl/pull/1772 is a duplicate to this PR that runs downstream CI against https://github.com/oscar-system/Oscar.jl/pull/4635. (Thus closes #1772.)

It is exactly the same setup that we use for TestExt in AbstractAlgebra. Once we raise the minimum required julia version, most of this can get reverted again.

ping @thofma


lgoettgens:
After talking to @thofma on slack, we decided to not do these changes. And they turned out to not be needed for https://github.com/oscar-system/Oscar.jl/pull/4635.

PR #1779, opened 2025-02-25
Add `is_cyclotomic_polynomial_with_data`

### LINKED ISSUE #1777: Extend `is_cyclotomic_polynomial`
Currently there is the function `is_cyclotomic_polynomial` which returns either true or false. But internally there is some more information computed. Namely in the case of returning true it is know which cyclotomic polynomial was entered. Sometimes it is interesting to know exactly this.
I'm not so sure how this method should be named and what it should return if the input is not a cyclotomic. Is there something similar already implemented which could be used as an inspiration?

SoongNoonien:
This fixes #1777.

### DIFF
--- a/src/Misc/Poly.jl
+++ b/src/Misc/Poly.jl
@@ -931,6 +931,35 @@ function cyclotomic_polynomial(n::Int, R::PolyRing{T} = Hecke.Globals.Zx) where 
 end
 
 @doc raw"""
+    is_cyclotomic_polynomial_with_data(p::PolyRingElem{T}) where T -> Bool, Int
+
+Return a tuple containing whether `p` is cyclotomic and which cyclotomic polynomial it is.
+
+# Examples
+
+```jldoctest
+julia> _, b = cyclotomic_field_as_cm_extension(12)
+(Relative number field of degree 2 over maximal real subfield of cyclotomic field of order 12, z_12)
+
+julia> is_cyclotomic_polynomial_with_data(minpoly(b))
+(false, -1)
+
+julia> is_cyclotomic_polynomial_with_data(absolute_minpoly(b))
+(true, 12)
+```
+"""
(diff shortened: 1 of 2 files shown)


src/Misc/Poly.jl
```diff
@@ -930,6 +930,36 @@ function cyclotomic_polynomial(n::Int, R::PolyRing{T} = Hecke.Globals.Zx) where
...
+(false, -1)
+
+julia> is_cyclotomic_polynomial_with_data(absolute_minpoly(b))
+(true, 12)
+```
+"""
+function is_cyclotomic_polynomial_with_data(p::PolyRingElem{T}) where T
+  n = degree(p)
+  R = parent(p)::PolyRing{T}
+  list_cyc = union(Int[k for k in euler_phi_inv(n)], [1])::Vector{Int}
```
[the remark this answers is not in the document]
thofma: if it works, sure, why not

src/Misc/Poly.jl
```diff
@@ -930,6 +930,36 @@ function cyclotomic_polynomial(n::Int, R::PolyRing{T} = Hecke.Globals.Zx) where
   return map_coefficients(base_ring(R), p, parent = R)::PolyRingElem{T}
 end
 
+@doc raw"""
+    is_cyclotomic_polynomial_with_data(p::PolyRingElem{T}) where T -> Bool
+
+Return a tuple containing whether `p` is cyclotomic and which cyclotomic polyinomial it is.
```
StevellM, line 936:
```suggestion
    is_cyclotomic_polynomial_with_data(p::PolyRingElem{T}) where T -> Bool, Int

Return a tuple containing whether `p` is cyclotomic and which cyclotomic polynomial it is.
```

SoongNoonien: Thank you!

fingolfin (approved):

PR #1784, opened 2025-02-26
Add cyclotomic factorization

SoongNoonien:
For GenericCharacterTables we would like to have a factorization of polynomials into cyclotomic polynomials. This is mostly relevant for more human readable printing. I'm not completely sure if Hecke is the right place for this but I couldn't find a more fitting place. I think GenericCharacterTables is not the right place since this factorization may also be of interest outside of GenericCharacterTables.
Currently the phis are always printed in unicode. I could use the `is_unicode_allowed` method to fix this but I think this should be handled by the expressify system which also converts the unicode characters to latex code.

Are there any methods missing I should implement for `CyclotomicFac`?


thofma:
The docstring says:
> Return a factorization of `p` into cyclotomic polynomials. The main difference
to `factor` is the way the result gets printed.

The example does not factor into cyclotomic polynomials. Can you clarify what the function does?

SoongNoonien:
I'm mean a factorization mostly into cyclotomic polynomials. For some inputs there will be a remainder factor which is not cyclotomic.

thofma:
I imagine that this is useful for nice output in GenericCharacterTables, which I think would be the best place for it. (I think a user would assume that `factor_cyclotomic` would factor a polynomial completely into cyclotomic polynomials.)

SoongNoonien:
I've talked to @fieker today and we've come to the conclusion that this should indeed be a part of GenericCharacterTables.jl for now.

PR #1790, opened 2025-03-04
Reactivate rank for FinGenAbGroup

fingolfin:


thofma:
I guess it is time

lgoettgens:
> I guess it is time

Could you then please merge both this and https://github.com/oscar-system/Oscar.jl/pull/4698?

PR #1798, opened 2025-03-11
feat: export `subfield`

thofma:
- add `sub` alias


src/NumField/Subfields.jl
```diff
@@ -29,7 +29,7 @@ function _subfield_basis(K::S, as::Vector{T}) where {
       (Fvs,phivs) = sub(Kvs, typeof(Fvs)[Fvs, abvs])
       if dF != length(Fvs.gens) # dim(Fvs)
         dF = length(Fvs.gens) # dim(Fvs)
-        append!(nbs, abs)
+        ppend!(nbs, abs)
```
fieker, line 32: ppend? for m it is truncated. The "a" is missing

thofma: oops

lgoettgens:
Using this PR together with Oscar (1.3.0 or master) results in `WARNING: using SubfieldLattice_Module.subfield in module Oscar conflicts with an existing identifier.` being printed during precompilation. I haven't checked if any functionality breaks.

lgoettgens:
If somehow possible, a fix for that should be in Oscar 1.3.1 (cc @aaruni96)

thofma:
I was looking for this over at https://github.com/thofma/Hecke.jl/actions/runs/13791087110/job/38570946079 before merging, but could not find it.

thofma:
seems to only appear in the 1.10 tests

thofma:
We could fix it in 1.3.1, but there would still be a combination of Oscar 1.3 + Hecke 0.35.?, where it prints a warning during precompilation. Is this acceptable?

lgoettgens:
> We could fix it in 1.3.1, but there would still be a combination of Oscar 1.3 + Hecke 0.35.?, where it prints a warning during precompilation. Is this acceptable?

Yes. To not have this appear in the "most recent" Oscar release, this should however only be released in Hecke *after* Oscar 1.3.1 with the fix has been released. That way normal usage of `] up` never resolves to the above combination (of course unless someone pins Oscar to 1.3.0, but we shouldn't care for that).

PR #1803, opened 2025-03-14
Fix some docs typos

lgoettgens:
I collect everything I find while reading the Number theory part of the oscar docs.


thofma:
Thanks

PR #1809, opened 2025-03-18
Add `quaternion_algebra` constructor

thofma:
- export it

### DIFF
--- a/src/AlgAss/AlgQuat.jl
+++ b/src/AlgAss/AlgQuat.jl
@@ -1,3 +1,18 @@
+@doc raw"""
+    quaternion_algebra(K::Field, a, b) -> QuaternionAlgebra
+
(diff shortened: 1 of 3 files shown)


src/AlgAss/AlgQuat.jl
```diff
@@ -1,3 +1,18 @@
+@doc raw"""
+    quaternion_algebra(K::Field, a, b) -> QuaternionAlgebra
+
+Return the quaternion algebra $(a, b | K)$ defined by $i^2 = -1$, $j^2 - 1$.
```
StevellM, line 4: Do you mean :
```suggestion
Return the quaternion algebra $(a, b | K)$ defined by $i^2 = a$, $j^2 = b$.
```
?

thofma: yeah sounds better (although always taking -1 would make everything easier)

PR #1812, opened 2025-03-20
Convert `@repl` blocks to `jldoctest`s; improve printing in doctests

lgoettgens:
Same reasoning as in https://github.com/oscar-system/Oscar.jl/pull/4747. In particular, all `@repl` blocks in Hecke doc pages are always run when building the Oscar docs, independent of if the particular page is visible in the Oscar docs or not.

As a bonus, this uncovered some errors that were not known before, as `@repl` happily continues when encountering an error. (see my comments below for details)

This PR reduces the time spent on the first call of a session to `Oscar.build_doc()` from 273sec to 135sec on my machine, and for follow-up runs in the same session from 26sec to 18sec.

### DIFF
--- a/docs/src/manual/quad_forms/basics.md
+++ b/docs/src/manual/quad_forms/basics.md
@@ -1,8 +1,6 @@
 ```@meta
 CurrentModule = Hecke
-DocTestSetup = quote
-    using Hecke
-end
+DocTestSetup = Hecke.doctestsetup()
 ```
 # [Spaces](@id Spaces2)
 
--- a/docs/src/manual/quad_forms/basics.md
+++ b/docs/src/manual/quad_forms/basics.md
@@ -21,13 +21,29 @@ hermitian_space(::NumField, ::MatElem)
 Here are easy examples to see how these constructors work. We will keep the two
 following spaces for the rest of this section:
 
-```@repl 2
-using Hecke # hide
-K, a = cyclotomic_real_subfield(7);
-Kt, t = K["t"];
-E, b = number_field(t^2-a*t+1, "b");
-Q = quadratic_space(K, K[0 1; 1 0])
-H = hermitian_space(E, 3)
+```jldoctest
+julia> K, a = cyclotomic_real_subfield(7);
+
+julia> Kt, t = K[:t];
+
+julia> E, b = number_field(t^2-a*t+1, :b);
+
+julia> Q = quadratic_space(K, K[0 1; 1 0])
+Quadratic space of dimension 2
+  over maximal real subfield of cyclotomic field of order 7
+with gram matrix
+[0   1]
+[1   0]
+
+julia> H = hermitian_space(E, 3)
+Hermitian space of dimension 3
+  over relative number field with defining polynomial t^2 - (z_7 + 1/z_7)*t + 1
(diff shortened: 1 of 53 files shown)


docs/src/manual/quad_forms/basics.md
```diff
@@ -229,14 +319,21 @@ biproduct(::Vector{AbstractSpace})
 
 ### Example
 
-```@repl 2
-using Hecke # hide
-E, b = cyclotomix_field_as_cm_extensions(7);
```
lgoettgens, line 234: This function does not exist. Please check if I fixed it in a sensible way.

docs/src/manual/quad_forms/genusherm.md
```diff
@@ -80,19 +80,46 @@ There are two ways of creating a local genus symbol for hermitian lattices:
 We will construct two examples for the rest of this section. Note that the prime
 chosen here is bad.
 
-```@repl 2
```
lgoettgens, line 83: In this file, each example block re-defines everything it needs. Thus, we don't need to add an identifier (here `2`) to force all of them to be executed in the same namespace

docs/src/manual/quad_forms/genusherm.md
```diff
@@ -307,23 +415,53 @@ lattices:
...
+julia> g1 = genus(HermLat, E, p, [(0, 1, 1, 0), (2, 2, -1, 1)], type = :det);
+
+julia> infp = infinite_places(E)
+3-element Vector{InfPlc{Hecke.RelSimpleNumField{AbsSimpleNumFieldElem}, RelSimpleNumFieldEmbedding{AbsSimpleNumFieldEmbedding, Hecke.RelSimpleNumField{AbsSimpleNumFieldElem}}}}:
+ Infinite place corresponding to (Complex embedding corresponding to root -1.19 of relative number field)
+ Infinite place corresponding to (Complex embedding corresponding to root 1.19 of relative number field)
+ Infinite place corresponding to (Complex embedding corresponding to root 0.00 + 1.19 * i of relative number field)
+
+julia> SEK = unique([r.base_field_place for r in infp if isreal(r.base_field_place) && !isreal(r)]);
+ERROR: type InfPlc has no field base_field_place
```
lgoettgens, line 440: This is broken, but I don't know how this was intended to be

thofma: I'll push a fix to your branch later today, ok?

lgoettgens: Works for me, thanks. I won't force-merge anymore

lgoettgens:
Ok, that should be it from my side. No more force-pushing. You can now take over to fix the example.

lgoettgens:
I have the feeling that I added a lot of randomized doctests. These should either get a fixed seed in a `setup = :(dosomething)` block, or an appropriate filter.
For the errors in `manual/quad_form/lattices.md:795-891`, it may be suitable to add `jldoctest; filter = r"\[\d, \d, \d, \d\]"` to suppress the randomness. If some output is intrinsically random, one needs to add `jldoctest; filter = r".*"` to allow every output.
But that is something someone somewhat familiar with the code needs to do (since I won't see myself qualified to distinguish an error from some randomized output).

thofma:
Thanks, it looks good

PR #1835, opened 2025-04-01
Cleanup hom functions for `FinGenAbGrp` and `TorQuadModule`

StevellM:
Needed some cleanup, and there were some problems for trivial groups/empty lists of images.
Deprecate some previous methods which relied on empty lists in input which were not checked; add preferable methods where we enforce an input parent.

@simonbrandhorst


simonbrandhorst:
Let me add that the `hom(ListOfElements,ListOfElements)` and 
`hom(Domain,ImagesOfGenerators)` methods exist only for these types of abelian groups and are not supported in general. I think we should get rid of these methods, because they lead to bugs for corner cases.
Thanks for looking into this Stevell.

src/GrpAb/Map.jl
```diff
@@ -192,28 +218,64 @@ function check_mat(A::FinGenAbGroup, B::FinGenAbGroup, M::ZZMatrix)
...
-function hom(A::FinGenAbGroup, B::FinGenAbGroup, M::ZZMatrix; check::Bool = true)
+@doc raw"""
+    hom(
+      G::FinGenAbGroup,
+      H::FinGenAbGroup,
+      M::ZZMatrix;
+      check::Bool=true,
+    ) -> FinGenAbGroupHom
+
+Return the group homomorphism $G\to H$ defined by the integer matrix `M`.
```
simonbrandhorst, line 229:
```suggestion
Return the group homomorphism $G\to H$ defined by the integer matrix `M=(m_{ij})`, 
by sending the $i$-th generator of $G$ to $\sum_{j=1}^n m_{ij}h_j$ where $h_1,\dots,h_m$ are the generators of $H$. 
```

simonbrandhorst (approved):

thofma:
I am happy to always have `hom(A, B, ...)`. I can talk to Claus tomorrow morning and see if he has any strong opinion.

thofma:
No strong opinion from Kaiserslautern.

src/GrpAb/Map.jl
```diff
@@ -103,80 +103,106 @@ end
...
-    m = vcat(ZZMatrix[x.coeff for x in A])
-    m = vcat(m, rels(parent(A[1])))
-    T = kernel(transpose(m), side = :right)
+function hom(
+    G::FinGenAbGroup,
+    H::FinGenAbGroup,
+    A::Vector{FinGenAbGroupElem},
+    B::Vector{FinGenAbGroupElem};
+    check::Bool=true
+  )
```
simonbrandhorst, line 142: Ideally, we want the types of `A` and `B` as stated. The method will complain if there is some type instable or sloppy code and we get a `Vector{Any}`. These errors can be confusing for beginners. 
Do we care? Maybe it would be better to just put `A::Vector` ? Any opinions?

thofma: I agree. Just put `Vector`

StevellM: Sure

simonbrandhorst (approved):

simonbrandhorst:
This is probably to be considered breaking since we added a deprecation. Curiously, it does not seem to break anything in Oscar according to CI :-).

thofma:
It breaks things in Oscar, but the tests we run do not throw for deprecations, but only print warnings.

PR #1840, opened 2025-04-02
feat: overhaul printing of orders and ideals

thofma:


thofma:
still missing fractional ideals

thofma:
Updated (went back to `<...>`, because it looks much better in expressions that use an ideal:
```
julia> 2*OK
Ideal of order of number field of degree 2 over QQ
  with generator 2
  of norm 4
  of minimum 2

julia> [2*OK, 3*OK]
2-element Vector{AbsSimpleNumFieldOrderIdeal}:
 <2, 2>
 <3, 3>

julia> K(2) * OK
Fractional ideal with numerator
  ideal of order of number field of degree 2 over QQ
    with generator 2
    of norm 4
  with denominator 1

julia> [K(1//2) * OK]
1-element Vector{AbsSimpleNumFieldOrderFractionalIdeal}:
 <1, 1>//2
```

docs/src/howto/defineorder.md
```diff
@@ -19,29 +19,37 @@ This is so-called *equation order* can be defined as follows:
 
 ```jldoctest 1
 julia> E = equation_order(K)
-Order of number field of degree 2 over QQ
-with Z-basis [1, a]
+Order of
+  number field with defining polynomial x^2 - 5
+    over rational field
+  with Z-basis [1, a]
```
StevellM, line 25: I am wondering if, according to OSCAR printing conventions, it would  not be better to have something printed as:
```
Order
  of number field with defining polynomial x^2 - 5
    over rational field
with Z-basis [1, a]
```
All the `of`, `over`, `from`... should be on a new line just before what they refer to.

docs/src/howto/reduction.md
```diff
@@ -31,10 +31,10 @@ julia> OK = maximal_order(K);
...
 
 julia> frakp = prime_decomposition(OK, p)[1][1]
-<29, z_7 + 22>
-Norm: 29
-Minimum: 29
-two normal wrt: 29
+Ideal of order of cyclotomic field of order 7
+  with 29-normal generators [29, z_7 + 22]
+  of norm 29
+  of minimum 29
```
StevellM, line 37: And this printed as
```
Ideal
  of order of cyclotomic field of order 7
with 29-normal generators [29, z_7 + 22]
  of norm 29
  of minimum 29
```

thofma:
Hm, I don't like
```
Order
  of ...
with ...
with ...
```
it kind of defeats the purpose of using indentation to mark where things start or end.

docs/src/howto/defineorder.md
thofma: Hm? You mean
```
Order
of number field with defining polynomial x^2 - 5
  over rational field
with Z-basis [1, a]
```
or else I don't understand the "rule".

docs/src/howto/reduction.md
thofma: This make it look like the second `of` are properties of the `with 29-normal ...` line.

docs/src/howto/defineorder.md
StevellM: This is the same as when we define maps I guess
```
julia> F = hom(A, B, [x^2+z, y^2-1, z^3])
Ring homomorphism
  from multivariate polynomial ring in 3 variables over QQ
  to quotient of multivariate polynomial ring by ideal (x*y)
defined by
  a -> x^2 + z
  b -> y^2 - 1
  c -> z^3
```
or what has been implemented with schemes also
```
julia> Y = affine_space(QQ,3)
Affine space of dimension 3
  over rational field
with coordinates [x1, x2, x3]
```
But yeah, anyway, within OSCAR there are still inconsistencies... I was just giving my opinion. I guess the most important was to move the `of` in the indented part. Where you put the `with Z-basis ...` is up to taste, maybe. I would still argue to de-indent it. My reason is that the basis belong to the object, the number field does not.

thofma: But the scheme example also has an `of` in the nonindented part. I really don't understand the rule (if there is actually one).

Compare with
```
julia> lie_algebra(QQ, :A, 5)
Abstract Lie algebra
  of type A5
  of dimension 35
over rational field
```
where als `of` things are indented, but where the indentation of `over` is the opposite of the scheme example.

StevellM: I guess in the case of scheme it is because the `of dimension 3` is small enough to be put on the first line (which happens often in other printings), and it is not a standalone `of`, there is something after it. 
I do not think there is a general rule, just a global guideline and suggestions. Some people might disagree with what I am telling you now, of course. 
And for the Lie Algebra printing, well, I do not agree with it.

docs/src/howto/reduction.md
StevellM: Yes that's my bad, I misunderstood it. Then why not keeping (non-indented)  `Norm: 29` and `Minimum: 29` as it was before ?

thofma: I find the rule of what is intended and what not quite arbitrary. So I intended to indent everything consistently.

Indentation should help parse things. But if indentation is not consistent, this does not help. For example printing many ideals would give something like
```
Ideal
  of order of cyclotomic field of order 7
with 29-normal generators [29, z_7 + 22]
  of norm 29
  of minimum 29
Ideal
  of order of cyclotomic field of order 7
with 29-normal generators [29, z_7 + 22]
  of norm 29
  of minimum 29
Ideal
  of order of cyclotomic field of order 7
with 29-normal generators [29, z_7 + 22]
  of norm 29
  of minimum 29
```
It is now difficult to see where one thing starts/ends.

thofma:
Added the relative stuff and also the pseudo-matrix. It went from
```
Pseudo-matrix over Relative maximal order of Relative number field of degree 2 over K
with pseudo-basis
(1, 1//1 * <1, 1>)
(b + 1, 1//2 * <1, 1>)
Fractional ideal with row [1 0 0 0]
Fractional ideal with row [5 1 0 0]
Fractional ideal with row [3 0 1 0]
Fractional ideal with row [0 0 0 1]
```
to
```
<7, 1//2*b + 7//2>//1 * [1 0 0 0]
<1>//1 * [5 1 0 0]
<1>//1 * [3 0 1 0]
<1>//1 * [0 0 0 1]
```
Probably the right thing to do would be to first print the row and then the ideal so that matrix is aligned.

thofma:
I am finished for now and I think it is a clear improvement. There might be some disagreement about minor details, but this would be the case for any change.

StevellM:
It looks nice, I particularly like the new printing for pseudo-matrices! You are right in the sense that it would be hard to please everybody. I think in general the new printing is good, and it is clearly an improvement, as you said.
The only comment I have left, which is actually a question, is about fractional ideals. So currently it is of the form

```
Fractional ideal
  with numerator ideal of maximal order of relative number field of degree 2 over K
  with pseudo-basis
    (1, <7>//1)
    (b + 1, <7>//2)
  with denominator 1
```
Would it make sense to indent more the numerator ideal, like
```
Fractional ideal
  with numerator
    ideal of maximal order of relative number field of degree 2 over K
    with pseudo-basis
      (1, <7>//1)
      (b + 1, <7>//2)
  with denominator 1
```
or to just compactify it

```
Fractional ideal
  with numerator [compact printing of numerator]
  with denominator 1
```
? I have no preferences; the current printing looks fine to me.

thofma:
OK, I see. I think I will go with
```
Fractional ideal of maximal order of relative number field of degree 2 over K
with numerator
  ideal of maximal order of relative number field of degree 2 over K
  with pseudo-basis
    (1, <1, 1>//1)
    (a + 3, <6, 17*a^2 + 22*a + 2>//6)
with denominator 1
```
if there is no objection.

thofma:
Tests look good. Breakage is expected. I might use the chance to update some other printing as well in this PR here.

PR #1848, opened 2025-04-12
fix: add verification in sparse row constructor

thofma:
Positions need to be unique

(I tried to make it as cheap as possible, but we could also add a `check` to eliminate the check completely.)

### DIFF
--- a/src/Sparse/ZZRow.jl
+++ b/src/Sparse/ZZRow.jl
@@ -409,7 +409,7 @@ function pop_first!(a::SRow{ZZRingElem, ZZRingElem_Array}, len::Int = length(a))
     len -= 1
     _perc_down!(a, 1, len)
   end
-    
+
   return xi=>xv, len
 end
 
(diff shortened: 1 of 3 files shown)


thofma:
@lgoettgens this actually catches errors in `BasisLieHighestWeight` in the `kron` method: https://github.com/thofma/Hecke.jl/actions/runs/14417762076/job/40436509316?pr=1848#step:7:1465

thofma:
Maybe in `MBOld` we can just do `kron(A, B) = kronecker_product(A, B)`?

lgoettgens:
> Maybe in `MBOld` we can just do `kron(A, B) = kronecker_product(A, B)`?

If that passes the tests, please go ahead and do that.
I actually don't know much about MBOld, since that was written before my time (and before much functionality in Oscar existed). This nowadays just gets used as a (slow) reference implementation in the tests to compare the results of that and the actual algorithm. Thus, if you just adapt MBOld and not the algorithm itself, the testsuite should catch any possible issues.

src/Sparse/ZZRow.jl
```diff
@@ -485,14 +485,24 @@ function ==(a::SRow{ZZRingElem, ZZRingElem_Array}, b::SRow{ZZRingElem, ZZRingEle
...
+function sparse_row(R::ZZRing, A::Vector{Tuple{Int, Int}}; sort::Bool = true)
   if sort && length(A) > 1
     A = Base.sort(A, lt=(a,b)->isless(a[1], b[1]))
   end
   a = ZZRingElem_Array()
   sizehint!(a, length(A))
   l = Int[]
-  for (p, v) = A
+  # assume A is sorted
+  local c
```
lgoettgens, line 496: Is there a reason that the check is inlined here instead of calling the helper function? Same question for the one below around line 520

thofma: Yes, to not traverse the array twice.

PR #1861, opened 2025-04-27
hotfix for buggy Oscar code

thofma:


src/NumFieldOrd/NfOrd/NfOrd.jl
```diff
@@ -725,6 +725,19 @@ order is cached for future use.
 """
 function order(K::S, a::Vector{T}; check::Bool = true, isbasis::Bool = false,
                cached::Bool = false) where {S <: NumField{QQFieldElem}, T <: NumFieldElem{QQFieldElem}}
+  # hot-fix for some Oscar bug
+  # Should be removed once Oscar 1.4(?) is out
```
lgoettgens, line 729: That's not possible, as older Oscar versions may still resolve with that new Hecke.
Instead:
```suggestion
  # Should be removed in a future breaking Hecke release
```

thofma: Yes, then one gets the broken version one has now, so no regression.

PR #1865, opened 2025-05-13
Remove dependency on InteractiveUtils

lgoettgens:
The one use of `subtypes` was fragile anyway due to not adding methods for subtypes only defined after loading Hecke (e.g. in OSCAR).


src/AlgAss/Elem.jl
```diff
@@ -735,22 +735,23 @@ function (A::MatAlgebra)(a::MatAlgebraElem)
...
-  @eval begin
-    function (A::$T)(a::Union{Integer, ZZRingElem, Rational{<: Integer}})
-      return A(base_ring(A)(a))
-    end
 
-    #function (A::$T{S})(a::S) where {S <: RingElem}
-    #  return a*one(A)
-    #end
-  end
+function (A::AbstractAssociativeAlgebra)(a::Union{Integer, ZZRingElem, Rational{<: Integer}})
```
lgoettgens, line 739: This now creates an ambiguity for CliffordAlgebras in Oscar, as the previous eval loop was brittle and required the type in downstream to define it themselves. So unfortunately, this PR will need to wait for a breaking release. @thofma could you add the label for that?

PR #1869, opened 2025-05-16
Move things from Oscar/MoveToHecke.jl

thofma:


src/Misc/Poly.jl
```diff
@@ -1060,3 +1060,15 @@ function gcd_right(f::NCPolyRingElem, g::NCPolyRingElem)
   end
   return f
 end
+
+##
+
+function numerator(f::QQPolyRingElem, parent::ZZPolyRing = Hecke.Globals.Zx)
```
fingolfin, line 1066: Should we maybe also introduce a `Nemo.Globals.Zx` etc.? Then we could move this and a couple other similar things to Nemo?

CC @lgoettgens

PR #1874, opened 2025-05-21
Add workaround for GLIBCXX issue with julia 1.6

lgoettgens:
Same as https://github.com/Nemocas/Nemo.jl/pull/2088.
This should fix the downstream tests. As this may happen with all of Polymake.jl (at least in theory), it is safest to do just in all CI jobs that may run Polymake.jl.


thofma:
Merci

PR #1875, opened 2025-05-21
fix: Make `norm_equation` work for q-adic fields

wfsteiner:
While there, fix some trailing whitespaces.


fieker:
what is wrong with norm equations - apart from the trailing spaces?

thofma:
```
julia> K = qadic_field(3, 5, precision = 10)[1]
Unramified extension of 3-adic numbers of degree 5

julia> k = padic_field(3, precision = 10)
Field of 3-adic numbers

julia> norm_equation(K, k(4))
ERROR: MethodError: no method matching tr(::QadicFieldElem, ::PadicField)
```

PR #1876, opened 2025-05-30
feat: enumeration of vectors in `ZZLat` with given square and divisibility

StevellM:


simonbrandhorst:
Each of the methods for `vectors_of_square_and_divisibility` is documented individually which is creating a lot of noise in the documentation. Would it make sense to just document them all in a single docstring?

src/QuadForm/Quad/ZLattices.jl
```diff
@@ -1670,6 +1670,267 @@ function divisibility(L::ZZLat, v::QQMatrix)
...
+fractional ideal $\Phi(v, L)$ of $\mathbb{Z}$.
+
+The entry `n` must be nonzero and `d` must be a positive rational number,
+set to `scale(L)` by default.
+
+!!! note
+    Alternatively, instead of single values `n` and `d` one can input:
+    * a list of pairs of rational numbers `(n, d)` where `n` is nonzero and
+      `d` is positive;
+    * a dictionary whose keys are positive rational number `d` and the
```
simonbrandhorst:
```suggestion
    * a dictionary whose keys are positive rational numbers `d` and the
```

src/QuadForm/Quad/ZLattices.jl
```diff
@@ -1670,6 +1670,267 @@ function divisibility(L::ZZLat, v::QQMatrix)
...
+
+!!! note
+    In the case where one wants to choose ``S`` to be ``L`` itself, one can
+    call instead `vectors_of_square_and_divisibility(L, n, d)`.
+
+One can choose in which coordinates system each vector `v` in output is
+represented by changing the symbol `coordinates_representation`.
+There are three possibilities:
+  - `coordinates_representation = :L`: the vector `v` is given in terms of its
+    coordinates in the fixed basis of the top lattice ``L``;
```
simonbrandhorst: Do you require $S \subseteq L \otimes \mathbb{Q}$? Otherwise this may not be well defined.

StevellM: No in fact, I do not require it. I should change the docstring to mention the "coordinates in the standard basis of the rational span of ``L`` ".

simonbrandhorst: I cannot follow. If $x$ is not contained in $L \otimes \mathbb{Q}$ it does not have "coordinates in the standard basis of the rational span of `L`" . (Besides these are just the coordinates in the basis of $L$)

StevellM: The beginning of the docstring mentions that we look for vectors in $S \cap (L\otimes \mathbb{Q})$, so the vectors are contained both in $S$ and the rational span of $L$.
And yes, these are coordinates in the basis of $L$, which is what I wrote before, but since the vector could be not in $L$, then it is not clear if coordinates mean integral or rational coordinates... I guess talking about rational span could remove this ambiguity.

simonbrandhorst (approved):

src/QuadForm/Quad/ZLattices.jl
simonbrandhorst: Right, sorry. It must have been too hot in my office to think clearly :-).

PR #1882, opened 2025-06-10
fix `norm(f::PolyRingElem{AbsSimpleNumFieldElem})`

### LINKED ISSUE #1878: problem with `norm(f::PolyRingElem{AbsSimpleNumFieldElem})`
```
julia> using Hecke
 _    _           _        
| |  | |         | |         |  Software package for
| |__| | ___  ___| | _____   |  algorithmic algebraic number theory
|  __  |/ _ \/ __| |/ / _ \  |  
| |  | |  __/ (__|   <  __/  |  Manual: https://thofma.github.io/Hecke.jl
|_|  |_|\___|\___|_|\_\___|  |  Version 0.36.0

julia> _, x = polynomial_ring(QQ, "x");

julia> F, _ = number_field(x^2-2);

julia> R, x = polynomial_ring(F, "x");

julia> f = x^10 + x^9;

julia> norm( f )
x^20 + 2*x^19 + x^18

julia> norm( f*x)
ERROR: Requires nonzero constant coefficient
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] polynomial_to_power_sums(f::AbstractAlgebra.Generic.Poly{AbsSimpleNumFieldElem}, n::Int64)
[...]
```
The code for `polynomial_to_power_sums` gets called if the argument `f` of `norm` has degree larger than 10.
Should we switch to the code for smaller degrees (base on resultants) if the constant coefficient of `f` is zero?
Or perhaps shift `f` such that it has nonzero constant coefficient?

ThomasBreuer:
resolves #1878 

The new test looks complicated.
Apparently each `norm` result may or may not get a newly created parent, and we do not know whether such a result is compatible with other polynomials.
```
julia> Qx, x = polynomial_ring(QQ, "x");

julia> F, _ = number_field(x^2-2);

julia> Fx, x = polynomial_ring(F, "x");

julia> parent(norm(x^2+1)) == parent(norm(x^2+1))
true

julia> parent(norm(x^2+x)) == parent(norm(x^2+x))
false
```
Is this behaviour intended?
Wouldn't it be better to set the parent of `norm(f)` to `base_ring(defining_polynomial(base_ring(parent(f))))`?

Another point:
In order to write a test, getting a norm with a given parent seems to be nontrivial.
For example, the following does not work.
```
julia> f = x^2 + x
x^2 + x

julia> n = norm(x^2 + x);

julia> parent(Qx(n)) == Qx
false
```
This is because of the method `(a::QQPolyRing)(b::QQPolyRingElem) = b` (from Nemo.jl), which ignores the parent.
Is this behaviour intended?

### DIFF
--- a/src/NumField/NfAbs/Elem.jl
+++ b/src/NumField/NfAbs/Elem.jl
@@ -265,6 +265,7 @@ function norm(f::PolyRingElem{AbsSimpleNumFieldElem})
     h = nf_poly_to_xy(f, Qxy, Qx)
     N = resultant(T, h)
   end
+  N.parent = parent(defining_polynomial(base_ring(parent(f))))
   return inflate(N, i)
 end
 
--- a/test/NfAbs/Elem.jl
+++ b/test/NfAbs/Elem.jl
@@ -169,11 +169,8 @@ end
   F, _ = number_field(xQ^2-2)
   Fx, xF = F["x"]
   d = degree(F)
-  for i in 1:11
-    f = xF^i + xF^(i+1)
-    nf = norm(f)
-    cand = xQ^(i*d) * norm(xF+1)
-    @test change_base_ring(QQ, nf, parent = Qx) == change_base_ring(QQ, cand, parent = Qx)
+  for i in 0:11
+    @test norm(xF^i + xF^(i+1)) == xQ^(i*d) * norm(xF+1)
   end
 end
 


fieker:
On Tue, Jun 10, 2025 at 06:58:18AM -0700, Thomas Breuer wrote:
> resolves #1878 
> 
> The new test looks complicated.
> Apparently each `norm` result may or may not get a newly created parent, and we do not know whether such a result is compatible with other polynomials.
> ```
> julia&gt; Qx, x = polynomial_ring(QQ, &quot;x&quot;);
> 
> julia&gt; F, _ = number_field(x^2-2);
> 
> julia&gt; Fx, x = polynomial_ring(F, &quot;x&quot;);
> 
> julia&gt; parent(norm(x^2+1)) == parent(norm(x^2+1))
> true
> 
> julia&gt; parent(norm(x^2+x)) == parent(norm(x^2+x))
> false
> ```
> Is this behaviour intended?
It is neither planned nor intended. So far norm was used as a once-off:
either factor it or
use it to create a field
in no application did we ever mix two norms...
> Wouldn&#39;t it be better to set the parent of `norm(f)` to `base_ring(defining_polynomial(base_ring(parent(f))))`?
Yep.

fingolfin:
Just looked at this with @fieker and he thought this looks OK. So I hope this can be merged soon by someone with commit rights

PR #1886, opened 2025-06-13
fix _is_isometric_indef

simonbrandhorst:


simonbrandhorst:
Added the example as a test.

PR #1891, opened 2025-06-23
fix a corner case

simonbrandhorst:


thofma:
obligatory question: do you want to add your example as a test?

test/QuadForm/Torsion.jl
```diff
@@ -245,6 +245,10 @@
   k = sub(t,gens(t)[3:4])
   @test order(radical_quadratic(k[1])[1])==2
 
+  discriminant_group(integer_lattice(ZZ[4*30]))
+  D = discriminant_group(integer_lattice(gram=ZZ[60*5;]))
+  s = sub(D,[10*D[1],30*D[1],60*D[1]])[1]
```
StevellM, line 249:
```suggestion
  s = sub(D, [10*D[1]])[1]
```
You just need one generator.

simonbrandhorst: I would imagine. But it doesn't hurt to test that it also works for redundant generators.

simonbrandhorst:
O.K. finally this should be the correct fix.

PR #1893, opened 2025-07-02
Support GAP 0.14

fingolfin:
CC @lgoettgens


lgoettgens (approved):

thofma:
What is up with the "matching: [Oscar#master] - ubuntu-latest, julia ~1.10.0-0" job? I removed the 1.6 jobs, but this one should work? Or did some name change?

Any idea @benlorenz?

thofma:
Ah right, thanks. Forgot about that one.

Issue #1899, opened 2025-07-03
simplify(..., canonical = true) sometimes much slower than pari's polredabs


fagu:
An example where pari's polredabs is much faster than Hecke's `simplify(..., canonical = true)`:
```julia
using Oscar
R, x = polynomial_ring(ZZ, :x)
f = x^4 + 121550*x^2 + 4136832700
K, _ = number_field(f, :w, cached = false)
@time L, _ = simplify(K, canonical = true)
@time L, _ = simplify(K, canonical = true)
```
```
  6.377594 seconds (19.94 M allocations: 1.241 GiB, 10.03% gc time, 73.39% compilation time)
  1.353708 seconds (19.28 M allocations: 1.213 GiB, 24.02% gc time, 0.97% compilation time)
```
In Pari/GP, it takes about 0.0002 seconds:
```gp
t = getwalltime();
for (i=1,1000,polredabs(x^4 + 121550*x^2 + 4136832700));
print("Time: ", strtime(getwalltime() - t), " / ", 1000);
quit;
```
```
Time: 241 ms / 1000
```

fingolfin:
Thank you for your report, much appreciated. @fieker is working on this in PR #1900 (I am sure he'll chime in here once he's happy with the outcome)

thofma:
It is still a bit slowish, so I will keep it open.

Issue #1897, opened 2025-07-03
simplify(..., canonical=True) disagrees with pari's polredabs


fagu:
Here's an example from LMFDB in which `simplify(..., canonical=true)` returns a different polynomial than specified at https://beta.lmfdb.org/knowledge/show/nf.polredabs:
```julia
using Oscar

function polredabs(f::QQPolyRingElem)::QQPolyRingElem
	K, _ = number_field(f, :w, cached=false)
	L, _ = simplify(K, canonical=true)
	return defining_polynomial(L)
end

R, x = polynomial_ring(QQ, :x)

f = x^3 - 65637*x - 5630196
g = polredabs(f)

println("f = ", f)
println("g = ", g)

println("disc(f) = ", discriminant(f))
println("disc(g) = ", discriminant(g))
```
outputs
```
f = x^3 - 65637*x - 5630196
g = x^3 - 65637*x - 87516
disc(f) = 275237544762180
disc(g) = 1130906639362500
```
The polynomials f and g have the same T2-norm 131274, but |disc(f)| < |disc(g)|, so f should win against g.

According to Pari/GP, polredabs(f) = f.

fingolfin:
Thank you for your report, much appreciated. @fieker is working on a fix in PR #1900

fagu:
Thanks for making me happy! :)

PR #1900, opened 2025-07-03
make fagu happy: #1897, #1899

fieker:


fieker:
ahh well, back to the old, other corner cases...
nothing is ever easy!

PR #1903, opened 2025-07-05
fix: avoid calling a nondefined `getindex` function

StevellM:
I have written a `getindex` function locally for `LineEnumCtx` but I am not sure we want it, so instead I keep track of whether or not we have used the automorphisms (sometimes all isometries are given by scalar multiplication so we have to fallback to a regular `enumerate_lines`... We should not `collect` it since it could consist of many lines).


StevellM:
Makes sense, thank you for the suggestion!

simonbrandhorst (approved):

PR #1907, opened 2025-07-09
Added quaternions in characteristic 2

eoinmackall:
Added a definition for quaternion algebras in characteristic 2.

(Tried some of the functionality with characteristic 2 algebras. The standard involution works as is --
but I was wondering if there was a reason it wasn't hard coded?)

### DIFF
--- a/docs/src/manual/algebras/basics.md
+++ b/docs/src/manual/algebras/basics.md
@@ -38,7 +38,7 @@ Elements of algebras can be constructed by arithmetic with basis elements, gener
 julia> Q = quaternion_algebra(QQ, -1, -1)
 Quaternion algebra
   over rational field
-  defined by i^2 = -1, j^2 = -1
+  defined by i^2 = -1, j^2 = -1, ij = -ji
 
 julia> B = basis(Q);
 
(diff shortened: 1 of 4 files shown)


eoinmackall:
I think it's somewhat common to call k := ij but it's technically not in the algebra <i,j | i^2=a, j^2=b, ij=-ji> or sim. for char 2.

I can switch it back if that's preferred. Probably I need to edit the doctests also, one way or another.

thofma:
OK. I would prefer to stick with `k` for now.

Regarding doctests:
```
julia> using Hecke; Hecke.build_doc(;doctest = :fix)
```
should adjust them automatically.

thofma:
I think there are some tests that were marked as broken before and now work.

thofma:
We would really like to have this code. I would be happy to help with pushing this over the finish line in case you want some help.

eoinmackall:
Heya,

Sorry it's taking me a bit. I was just going through and checking whether other functions worked when called in char. 2 (I think there is an implementation of the reduced trace somewhere that divides by the degree, and returns an error now).

I'll get it finished soon but I have a visitor this week, so probably it'll be done sometime next week.

Best,
Eoin

PR #1910, opened 2025-07-11
Slightly improve printing of `MapClassGrp`

lgoettgens:
This resolves the first bullet point of https://github.com/oscar-system/Oscar.jl/issues/4935.

### DIFF
--- a/docs/src/manual/number_fields/class_fields.md
+++ b/docs/src/manual/number_fields/class_fields.md
@@ -65,8 +65,7 @@ julia> Qx, x = polynomial_ring(QQ, :x);
 julia> K, a = number_field(x^2 - 10, :a);
 
 julia> c, mc = class_group(K)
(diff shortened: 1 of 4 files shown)


thofma:
I already started working on this and was waiting for the other issue to be resolved.

thofma:
By the way, the `@show_name` should be behind a `is_terse(io)` check, if I understand it correctly?

lgoettgens:
> I already started working on this and was waiting for the other issue to be resolved.

Sorry for that. I thought you wanted to wait for the other thing to be resolved since I didn't see a draft.



> By the way, the `@show_name` should be behind a `is_terse(io)` check, if I understand it correctly?

I don't think so (and basically no other method does that). `@show_name` does the terseness check itself

thofma:
OK, thanks for checking.

PR #1913, opened 2025-07-13
feat: `primitive_root` and `coprime_residues`

thofma:


src/Misc/Integer.jl
```diff
@@ -757,3 +757,48 @@ function support(a::QQFieldElem)
...
+end
+
+function _coprime_residues(n::Int)
+  if n == 0
+    return Int[]
+  end
+  if n == 1
+    return Int[0]
+  end
+  n = abs(n)
```
ThomasBreuer, line 795: If negative `n` are supported, why are `_coprime_residues(1)` and `_coprime_residues(-1)` different?

src/Misc/UnitsModM.jl
```diff
@@ -667,3 +667,36 @@ end
...
+  @req n >= 1 "Argument ($n) must be positive"
+  if n == 1
+    return zero(n)
+  end
+  if n == 2
+    return one(n)
+  end
+  e, _n = remove(n, 2)
+  @req n == 1 || n == 2 || n == 4 ||
+    e in [0, 1] && is_prime_power(_n) "Argument ($n) must be 2, 4, p^k, or 2p^k with p an odd prime"
```
ThomasBreuer: The code and the tests support also `n = 1`.
(Note that `Hecke.is_prime_power(1)` is `false`.)

ThomasBreuer (approved):
Looks good, thanks; just two minor comments.

src/Misc/UnitsModM.jl
thofma: `1 = p^0`, so this should be included?

PR #1915, opened 2025-07-14
Bump julia compat to 1.10

lgoettgens:
This should not get any conflicts, if https://github.com/thofma/Hecke.jl/pull/1914 gets merged before.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,7 @@ docs/site/*
 docs/build/*
 docs/node_modules/*
 docs/package-lock.json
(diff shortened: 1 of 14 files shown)


thofma:
I'll do a patch release first later today

lgoettgens:
> I'll do a patch release first later today

Sounds good, thanks! You could even add https://github.com/thofma/Hecke.jl/pull/1884 into this breaking release, if the book test changes are mathematically correct.

PR #1923, opened 2025-07-16
Make `is_integral` and `is_algebraic_integer` more consistent

### LINKED ISSUE #1908: Make `is_integral` and `is_algebraic_integer` more consistent
Hi!
I just noted that for `QQBarFieldElem` only `is_algebraic_integer` is defined whereas for `AbsSimpleNumFieldElem` only `is_integral` is defined. That's a bit unexpected. I think in both cases both functions should be defined and return the same results. Perhaps `is_integral` should be the base function in both cases and `is_algebraic_integer` should be an alias of some kind. But I think there are some `is_integral` methods which wouldn't make sense to be available under `is_algebraic_integer` (e.g. for lattices).

SoongNoonien:
I noticed that `is_algebraic_integer` for `QQBarFieldElem` is actually defined in Nemo and directly calls a Flint method named `qqbar_is_algebraic_integer`. So, I decided to make `is_algebraic_integer` the "base function".

fixes #1908

### DIFF
--- a/src/exports.jl
+++ b/src/exports.jl
@@ -449,6 +449,7 @@ export is_GLZ_conjugate
 export is_abelian
 export is_absolute
 export is_absolutely_irreducible
+export is_algebraic_integer
 export is_anti_isometric_with_anti_isometry
 export is_anti_isometry
 export is_automorphous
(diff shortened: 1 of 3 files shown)


src/exports.jl
```diff
@@ -449,6 +449,7 @@ export is_GLZ_conjugate
 export is_abelian
 export is_absolute
 export is_absolutely_irreducible
+export is_algebraic_integer
```
SoongNoonien: I think this could be superfluous.

SoongNoonien: Yes, it was...

PR #1926, opened 2025-07-25
Add `vector_space_dim` for group algebras

emikelsons:
Add `vector_space_dim` for group algebras over fields. In the case the underlying group is infinite an `InfiniteDimensionError` is raised(to be moved to AA). This is rounding out oscar-system/Oscar.jl#5056.

cc @fingolfin

### DIFF
--- a/src/AlgAss/AlgGrp.jl
+++ b/src/AlgAss/AlgGrp.jl
@@ -31,6 +31,20 @@ group(A::GroupAlgebra) = A.group
 
 has_one(A::GroupAlgebra) = true
 
+##To be removed
+struct InfiniteDimensionError <: Exception
+end
+
+##To be removed
+function Base.showerror(io::IO, err::InfiniteDimensionError)
+  println(io, "Infinite-dimensional vector space")
+end
+
+function vector_space_dim(A::GroupAlgebra{T, S, R}) where {T <: FieldElem, S, R}
+  isfinite(group(A)) || throw(InfiniteDimensionError())
+  return order(group(A))
+end
+
 function (A::GroupAlgebra{T, S, R})(c::Union{Vector, SRow}; copy::Bool = false) where {T, S, R}
   c isa Vector && length(c) != dim(A) && error("Dimensions don't match.")
   return GroupAlgebraElem{T, typeof(A)}(A, copy ? deepcopy(c) : c)
--- a/src/Hecke.jl
+++ b/src/Hecke.jl
@@ -53,7 +53,7 @@ import AbstractAlgebra:
   set_assertion_level,
   set_verbosity_level
 
-import AbstractAlgebra: Solve, coprime_base_steel
+import AbstractAlgebra: Solve, coprime_base_steel, vector_space_dim
 
 import LinearAlgebra: dot, nullspace, rank, ishermitian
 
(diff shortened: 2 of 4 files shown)


src/AlgAss/AlgGrp.jl
```diff
@@ -31,6 +31,20 @@ group(A::GroupAlgebra) = A.group
...
+end
+
+##To be removed
+function Base.showerror(io::IO, err::InfiniteDimensionError)
+  println(io, "Infinite-dimensional vector space")
+end
+
+function vector_space_dim(A::GroupAlgebra{T, S, R}) where {T <: FieldElem, S, R}
+  isfinite(group(A)) || throw(InfiniteDimensionError())
+  return order(group(A))
```
thofma:
```suggestion
  return dim(A)
```

thofma:
The plan is to move the error first to AA, right?

src/Hecke.jl
```diff
@@ -53,7 +53,7 @@ import AbstractAlgebra:
   set_assertion_level,
   set_verbosity_level
 
-import AbstractAlgebra: Solve, coprime_base_steel
+import AbstractAlgebra: Solve, coprime_base_steel, vector_space_dim
```
lgoettgens:
```suggestion
import AbstractAlgebra: Solve, coprime_base_steel
```
this is not needed as Hecke imports everything that gets exported from AA

src/AlgAss/AlgGrp.jl
lgoettgens: I think the idea behind all of this `dim` business is to do it the other way around. Aka with this change, `dim` should delegate to `vector_space_dim`. (But IMO this should be in this PR @emikelsons)

Another thing: `order(group(A))` will return a ZZRingElem. I haven't followed what other `vector_space_dim` methods return, if Int or ZZRingElem. In any case, this should be consistent.

For Ints, you would do
```suggestion
  return order(Int, group(A))
```

src/AlgAss/AlgGrp.jl
```diff
@@ -12,7 +12,12 @@ base_ring(A::GroupAlgebra{T}) where {T} = A.base_ring::parent_type(T)
 
 base_ring_type(::Type{GroupAlgebra{T, S, R}}) where {T, S, R} = parent_type(T)
 
-Generic.dim(A::GroupAlgebra) = order(Int, group(A))
+Generic.dim(A::GroupAlgebra) = vector_space_dim(A)
+
+function vector_space_dim(A::GroupAlgebra{T, S, R}) where {T <: FieldElem, S, R}
+  isfinite(group(A)) || throw(InfiniteDimensionError())
+  return order(group(A))
```
thofma:
```suggestion
  return order(Int, group(A))
```
as suggested by @lgoettgens

test/AlgAss/AlgGrp.jl
```diff
@@ -37,6 +37,15 @@
     end
   end
 
+  @testset "Vector space dimension" begin
+    G1 = abelian_group(2,3)
+    G2 = abelian_group(0) 
+    R1 = group_algebra(QQ, G1)
+    R2 = group_algebra(QQ, G2)
+    @test vector_space_dim(R1) == 6
+    @test_throws InfiniteDimensionError vector_space_dim(R2)
```
thofma:
```suggestion
    @test_throws Hecke.InfiniteDimensionError vector_space_dim(R2)
```

PR #1932, opened 2025-07-29
Drop broken `Hecke.varinfo`

lgoettgens:
This function does fail with an `UndefVarError` when called as `varinfo()` (so just using defaults). Furthermore, IMO, this does have nothing to do with Computeralgebra and/or Hecke, and better suited to some personal `startup.jl` file instead of in this package.
~Dropping this allows tus o remove the dependency on `Markdown` (thus potentially slightly improving time-to-load).~ (Edit: not completely yet, but hopefully in the near future, see https://github.com/thofma/Hecke.jl/issues/1934)


thofma:
I guess we don't need it anymore.

PR #1937, opened 2025-08-05
feat: allow more index types in submatrix of sparse matrix

thofma:


src/Sparse/Matrix.jl
```diff
@@ -917,26 +917,61 @@ end
...
+        push!(rw.pos, ra.pos[j]-first(c)+1)
+      end
+    end
+    push!(B, rw)
+  end
+  return B
+end
+
 @doc raw"""
     sub(A::SMat, r::AbstractUnitRange, c::AbstractUnitRange) -> SMat
```
lgoettgens:
```suggestion
    sub(A::SMat, r::AbstractVector, c::AbstractVector) -> SMat
```
the newly added sentence only makes sense if you adapt the signature as well

PR #1938, opened 2025-08-05
Add resultant for univariate polynomials

SirToby25:
This pull requests adds the resultant methods for universal polynomial rings as well as some tests for it. (Probably this should be a part of AbstractAlgebra at some point...)

### DIFF
--- a/src/NumField/NfRel/NfRelNS.jl
+++ b/src/NumField/NfRel/NfRelNS.jl
@@ -579,6 +579,7 @@ function tr_via_minpoly(a::RelNonSimpleNumFieldElem)
   return -coeff(f, degree(f)-1)*div(degree(parent(a)), degree(f))
 end
 
+# TODO: move to AA together with resultant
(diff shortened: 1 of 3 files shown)


thofma:
Hm, I am a bit confused about the location. Why did you put it there?

SirToby25:
I tried to put it in AbstractAlgebra at first, but there is no `resultant` for multivariate polynomials available. I think that the original resultant method in the modified file should be a part of AbstractAlgebra.

src/NumField/NfRel/NfRelNS.jl
```diff
@@ -579,6 +579,14 @@ function tr_via_minpoly(a::RelNonSimpleNumFieldElem)
   return -coeff(f, degree(f)-1)*div(degree(parent(a)), degree(f))
 end
```
thofma, line 581:
```suggestion

# TODO: move to AA together with resultant
```

thofma:
Ah ok, makes sense.

PR #1950, opened 2025-08-18
feat: enable caching of automorphism group

thofma:
Not sure why we never enabled this.


fieker:
The PermGroup case ni Oscar caches for several years now...(to add embarassment)

PR #1961, opened 2025-08-25
chore: Declare compatibility with GAP.jl 0.15

lgoettgens:
I'll undraft this and restart CI once the release is actually done (i.e. the registration and everything is through), but I wanted to open this already now to make @thofma aware of this.
I don't expect any breakages in the GAP extension here.
It would be great if we could have a patch release once this is in. To ease this, I already bumped the version number.


lgoettgens:
The "Run long tests" job now runs with 0.15.0 (see https://github.com/thofma/Hecke.jl/actions/runs/17233301753/job/48892203429?pr=1961#step:7:73). So once that job is done, I think this is good to go.

PR #1962, opened 2025-08-26
CI: don't specify architecture

lgoettgens:
Jobs complain that the requested arch does not work and taking the default would be better, see e.g. https://github.com/thofma/Hecke.jl/actions/runs/17233301753/job/48892203435?pr=1961#step:3:8


lgoettgens:
The two failures here look unrelated.

One important thing though: The name of two CI jobs changes with this, so you would need to go into the branch protection settings and change the names of required CI jobs to match the new ones (i.e. without the `x64`)

PR #1981, opened 2025-08-31
fix: cornercase in ray class field

thofma:


src/RCF/rcf.jl
```diff
@@ -893,8 +893,8 @@ function _extend_auto(K::Hecke.RelSimpleNumField{AbsSimpleNumFieldElem}, h::NumF
   end
 end
 
-
 function _rcf_descent(CF::ClassField_pp)
+  Main.CF = CF
```
fieker: Some debugging?

thofma: Good catch, thanks.

PR #1982, opened 2025-08-31
Tweaks for  Hermitian genera

simonbrandhorst:
- Allow enumeration of hermitian genera with just giving min and max scale (no determinant)
- Add second keyword argument to discriminant_group as a shortcut for primary_part
 - Cache a representative of a hermitian genus
 - Allow evaluation of non-integral polynomials at TorQuadModuleHom

### DIFF
--- a/src/QuadForm/Lattices.jl
+++ b/src/QuadForm/Lattices.jl
@@ -1081,7 +1081,7 @@ function trace_lattice_with_isometry_and_transfer_data(H::AbstractLat{T}; alpha:
   # We only consider full rank lattices for simplicity
   @req degree(H) == rank(H) "Lattice must be of full rank"
   @req parent(beta) === E "beta must be an element of the base algebra of H"
-  @req (beta == QQ(1) || norm(beta) == 1) "beta must be of norm 1"
+  #@req (beta == QQ(1) || norm(beta) == 1) "beta must be of norm 1"
   @req !is_zero(alpha) "alpha must be non zero"
 
   n = degree(H)
@@ -1098,7 +1098,7 @@ function trace_lattice_with_isometry_and_transfer_data(H::AbstractLat{T}; alpha:
   end
 
   @req H isa HermLat "H must be hermitian or defined over the integers"
-  @req maximal_order(E) == equation_order(E) "Equation order and maximal order must coincide"
+  #@req maximal_order(E) == equation_order(E) "Equation order and maximal order must coincide"
 
   # This function perform the trace construction on the level of the
   # ambient spaces - we just need to transport the lattice
--- a/src/QuadForm/Torsion.jl
+++ b/src/QuadForm/Torsion.jl
@@ -101,7 +101,7 @@ function torsion_quadratic_module(M::ZZLat, N::ZZLat; gens::Union{Nothing, Vecto
 end
 
 @doc raw"""
-    discriminant_group(L::ZZLat) -> TorQuadModule
+    discriminant_group(L::ZZLat, [n::Int]) -> TorQuadModule
 
 Return the discriminant group of `L`.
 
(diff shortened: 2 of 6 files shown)


thofma:
The test fails because it does `@inferred`, which means that it checks whether it is type stable.

simonbrandhorst:
Yea, I failed to make `representative(::HermGenus)` type stable because of caching. 
What I would need to do is to infer the type of a representative from the type parameters in `HermGenus` 
I think caching outweighs type stability here.

simonbrandhorst:
Should be type stable now.

thofma:
Great, thanks you.

src/QuadForm/Lattices.jl
```diff
@@ -1081,7 +1081,7 @@ function trace_lattice_with_isometry_and_transfer_data(H::AbstractLat{T}; alpha:
   # We only consider full rank lattices for simplicity
   @req degree(H) == rank(H) "Lattice must be of full rank"
   @req parent(beta) === E "beta must be an element of the base algebra of H"
-  @req (beta == QQ(1) || norm(beta) == 1) "beta must be of norm 1"
+  #@req (beta == QQ(1) || norm(beta) == 1) "beta must be of norm 1"
```
StevellM: This does not seem related to the pull request, but rather a local change ?

src/QuadForm/Lattices.jl
```diff
@@ -1098,7 +1098,7 @@ function trace_lattice_with_isometry_and_transfer_data(H::AbstractLat{T}; alpha:
   end
 
   @req H isa HermLat "H must be hermitian or defined over the integers"
-  @req maximal_order(E) == equation_order(E) "Equation order and maximal order must coincide"
+  #@req maximal_order(E) == equation_order(E) "Equation order and maximal order must coincide"
```
StevellM, line 1103: Same

src/QuadForm/Torsion.jl
```diff
@@ -101,7 +101,7 @@ function torsion_quadratic_module(M::ZZLat, N::ZZLat; gens::Union{Nothing, Vecto
 end
 
 @doc raw"""
-    discriminant_group(L::ZZLat) -> TorQuadModule
+    discriminant_group(L::ZZLat, [n::Int]) -> TorQuadModule
```
StevellM, line 104: Does this work for any integer `n` or only prime numbers ? Maybe you could add also a piece of documentation to what the function does when one specifies a value for `n`.

src/QuadForm/Torsion.jl
```diff
@@ -1014,7 +1016,14 @@ end
 
 function evaluate(p::QQPolyRingElem, f::TorQuadModuleMap)
   @req domain(f) === codomain(f) "f must be a self-map"
-  @req all(a -> is_integral(a), coefficients(p)) "p must have integral coefficients"
+  if !all(a -> is_integral(a), coefficients(p)) "p must have integral coefficients"
+    l = lcm(ZZRingElem[denominator(i) for i in coefficients(p)])
+    e = elementary_divisors(domain(f))[end]
+    R,iR = residue_ring(ZZ,e;cached=false)
+    s = preimage(iR,inv(iR(l)))
```
StevellM:
```suggestion
    R, iR = residue_ring(ZZ, e; cached=false)
    s = preimage(iR, inv(iR(l)))
```

src/QuadForm/Torsion.jl
```diff
@@ -1014,7 +1016,14 @@ end
...
+    p = s*l*p
+    @req all(a -> is_integral(a), coefficients(p)) "the denominator of p must be coprime to the exponent of the domain of f"
```
StevellM:
```suggestion
    @req all(is_integral, coefficients(p)) "The denominator of p must be coprime to the exponent of the domain of f"
```

src/QuadForm/Torsion.jl
```diff
@@ -1014,7 +1016,14 @@ end
 
 function evaluate(p::QQPolyRingElem, f::TorQuadModuleMap)
   @req domain(f) === codomain(f) "f must be a self-map"
-  @req all(a -> is_integral(a), coefficients(p)) "p must have integral coefficients"
+  if !all(a -> is_integral(a), coefficients(p)) "p must have integral coefficients"
```
StevellM:
```suggestion
  if !all(is_integral, coefficients(p))
```

src/QuadForm/Lattices.jl
simonbrandhorst: This change should stay. You do not need the order to be maximal in order for the trace construction to work. 
I have some follow up code for non-maximal orders.

PR #1985, opened 2025-09-05
make subfield PE integral - to avoid problems

fieker:


thofma:
maybe put it in the `_subfield_from_primitive_element` method below

PR #1986, opened 2025-09-05
fix: don't rely on `Fac` internals

thofma:


src/EllCrv/Misc.jl
```diff
@@ -129,13 +129,13 @@ function nrootscubic(b, c, d, p)
...
   if length(fac) == 1
-    if fac[first(keys(fac.fac))] == 3
+    if fac[first(fac)[1]] == 3
       return ZZ(3)
     else
       return ZZ(0)
     end
   elseif length(fac) == 2
-    if fac[first(keys(fac.fac))]== 1 && fac[first(keys(fac.fac))] == 1
+    if fac[first(fac)[1]]== 1 && fac[first(fac)[1]] == 1
```
lgoettgens: this line contains twice the same condition, doesn't it? Is this code maybe assuming that `first` modifies the state of the iterator?

src/EllCrv/Misc.jl
```diff
@@ -157,13 +157,13 @@ function nrootscubic(b::AbsSimpleNumFieldElem, c::AbsSimpleNumFieldElem, d::AbsS
...
```
lgoettgens: same here

src/FieldFactory/ab_exts.jl
```diff
@@ -959,7 +959,7 @@ function discriminant_conductor(C::ClassField, bound::ZZRingElem; lwp::Dict{Tupl
   n = degree(C)
   e = Int(exponent(C))
   lp = mr.fact_mod
-  abs_disc = factor(discriminant(O)^n).fac
+  abs_disc = Dict(p => e for (p, e) in factor(discriminant(O)^n))
```
fieker: I think we had a bad day when we wrote this - or I am blind now. What about
```
abs_disc = Dict(p => n*e for (p, e) in factor(discriminant(O)))
```
to factor a smaller number?

src/EllCrv/Misc.jl
thofma: Who knows. These are questions for another day.

src/FieldFactory/ab_exts.jl
thofma: I just did the 1:1 translation. But you are right, although the factor code will probably try this first anyway.

src/EllCrv/Misc.jl
fingolfin: I would say this code is buggy: this function is meant to count roots of cubic polynomial modulo $p$. In this case there are exactly two factors... If there are two entries in `fac` then either both have multiplicity 1 (then one has degree 1 and the other degree 2). Or else one will have multiplicity 1 the other 2.

The comment in the next line says "one linear and one irreducible quadratic factor".
So I am sure what was really the intention here is to verify that the first and second/last entry both are. one. So it really should be
```suggestion
    if fac[first(fac)[1]]== 1 && fac[last(fac)[1]] == 1
```

src/EllCrv/Misc.jl
fingolfin:
```suggestion
    if fac[first(fac)[1]]== 1 && fac[last(fac)[1]] == 1
```

src/EllCrv/Misc.jl
thofma: There is no method for `last(fac)`.

src/Map/NumField.jl
```diff
@@ -1063,3 +1063,23 @@ function AbstractAlgebra.show_map_data(io::IO, f::NumFieldHom)
...
   end
 end
+
+################################################################################
+#
+#  QQ
+#
+################################################################################
+
+struct QQHom{S} <: Map{QQField, S, HeckeMap, Any}
```
fingolfin, line 1073: Huh, should this be in here? Or maybe rename the PR?

src/NumField/NfAbs/MPolyAbsFact.jl
```diff
@@ -94,7 +94,7 @@ mutable struct HenselCtxFqRelSeries{T}
     is_squarefree(g) || return nothing
     @assert is_squarefree(g)
 
-    lf = collect(keys(factor(g).fac))
+    lf = first.(collect(factor(g)))
```
fingolfin, line 97: Isn't this the same as prime divisors?
```suggestion
    lf = prime_divisors(g)
```

thofma: No, `prime_divisors` does not exist with this signature.

src/Map/NumField.jl
thofma: I'll adjust the commit message

PR #1991, opened 2025-09-09
Remove dependency on Pkg

fingolfin:
... to speed up 'using Hecke' in Julia >= 1.11

See https://github.com/JuliaLang/julia/issues/57970

Clearly this is not acceptable in its current form. But I think we can move the building of the documentation out of the code without loosing too much (or any) convenience if we do it right (also for OSCAR). I will make a proposal for this soon.

Without this PR (and with Julia 1.10.10 / 1.11.6 / 1.12.0-rc2:
```
tutulla:~$ hyperfine 'julia +1.10 --startup-file=no -e "using Hecke"' 'julia +1.11 --startup-file=no -e "using Hecke"' 'julia +1.12 --startup-file=no -e "using Hecke"'
Benchmark 1: julia +1.10 --startup-file=no -e "using Hecke"
  Time (mean ± σ):      1.685 s ±  0.026 s    [User: 2.028 s, System: 5.061 s]
  Range (min … max):    1.654 s …  1.750 s    10 runs

Benchmark 2: julia +1.11 --startup-file=no -e "using Hecke"
  Time (mean ± σ):      1.988 s ±  0.081 s    [User: 7.078 s, System: 0.372 s]
  Range (min … max):    1.894 s …  2.117 s    10 runs

Benchmark 3: julia +1.12 --startup-file=no -e "using Hecke"
  Time (mean ± σ):      2.322 s ±  0.091 s    [User: 7.345 s, System: 0.435 s]
  Range (min … max):    2.115 s …  2.411 s    10 runs

Summary
  julia +1.10 --startup-file=no -e "using Hecke" ran
    1.18 ± 0.05 times faster than julia +1.11 --startup-file=no -e "using Hecke"
    1.38 ± 0.06 times faster than julia +1.12 --startup-file=no -e "using Hecke"
```

With this PR:
```
tutulla:~/Projekte/OSCAR/Hecke.jl (git:mh/rm-Pkg)$ hyperfine 'julia +1.10 --project=p/1.10 --startup-file=no -e "using Hecke"' 'julia +1.11 --project=p/1.11 --startup-file=no -e "using Hecke"' 'julia +1.12 --project=p/1.12 --startup-file=no -e "using Hecke"'
Benchmark 1: julia +1.10 --project=p/1.10 --startup-file=no -e "using Hecke"
  Time (mean ± σ):      1.666 s ±  0.022 s    [User: 2.044 s, System: 5.027 s]
  Range (min … max):    1.641 s …  1.720 s    10 runs

Benchmark 2: julia +1.11 --project=p/1.11 --startup-file=no -e "using Hecke"
  Time (mean ± σ):      1.755 s ±  0.097 s    [User: 6.856 s, System: 0.338 s]
  Range (min … max):    1.687 s …  2.023 s    10 runs

Benchmark 3: julia +1.12 --project=p/1.12 --startup-file=no -e "using Hecke"
  Time (mean ± σ):      1.857 s ±  0.060 s    [User: 6.962 s, System: 0.341 s]
  Range (min … max):    1.756 s …  1.935 s    10 runs

Summary
  julia +1.10 --project=p/1.10 --startup-file=no -e "using Hecke" ran
    1.05 ± 0.06 times faster than julia +1.11 --project=p/1.11 --startup-file=no -e "using Hecke"
    1.11 ± 0.04 times faster than julia +1.12 --project=p/1.12 --startup-file=no -e "using Hecke"
```

So while it is not quite back to normal, it is quite a bit better.


thofma:
Yes, we can also ditch importing `nnz`. It does not matter which symbol it is. Very unfortunate that AbstractAlgebra depends on SparseArrays.

fingolfin:
The use of `SparseArrays` in AA is for the sake of permutations and Young tableaus. In fact as far as I can tell it is only used to implement various `matrix_repr` methods for those. As such I think it could be put into a package extension there as well.


Regarding `nnz` here: I would prefer `number_nonzero_entries` over `nnz` anyway ;-) but of course that'd be a breaking change.

PR #1996, opened 2025-09-10
fix: creation of number field morphisms into noncommutative rings

thofma:


thofma:
@wfsteiner should fix your problem

PR #2010, opened 2025-09-19
Add some size_hint for sparse rows

HechtiDerLachs:
This is a small chunk of things I played around with today to speed up some computations. I was wondering whether this is interesting in general. If you like it, feel free to take it in. 

I will add an example showcasing the benefit.

### DIFF
--- a/src/HeckeTypes.jl
+++ b/src/HeckeTypes.jl
@@ -346,6 +346,8 @@ mutable struct SRow{T, S} # S <: AbstractVector{T}
 
   function SRow(R::NCRing, A::Vector{Tuple{Int, T}}) where T
     r = SRow(R)
+    sizehint!(r.pos, length(A))
+    sizehint!(r.values, length(A))
     for (i, v) = A
       if !iszero(v)
         @assert parent(v) === R
@@ -358,6 +360,8 @@ mutable struct SRow{T, S} # S <: AbstractVector{T}
 
   function SRow(R::NCRing, A::Vector{Tuple{Int, Int}})
     r = SRow(R)
+    sizehint!(r.pos, length(A))
+    sizehint!(r.values, length(A))
     for (i, v) = A
       if !iszero(v)
         push!(r.pos, i)
--- a/src/Sparse/Row.jl
+++ b/src/Sparse/Row.jl
@@ -96,9 +96,9 @@ The elements $x_i$ must belong to the ring $R$.
 """
 function sparse_row(R::NCRing, A::Vector{Tuple{Int, T}}; sort::Bool = true) where T
   if sort && length(A) > 1
-    A = Base.sort(A, lt=(a,b) -> isless(a[1], b[1]))
+    A = Base.sort(A, by=first)
   end
-  _assert_is_unique_sorted(a[1] for a in A)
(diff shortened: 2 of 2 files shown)


HechtiDerLachs:
It seems more tricky than expected to come up with a reasonable example. Maybe the effect is not so big after all. When using `ProfileView` on my examples (not yet compatible with Oscar master, unfortunately), it showed that the numerous `push!` operations around my changes made here lead to many iterative allocations. I wanted to avoid that and `ProfileView` output suggested that these changes do the job. But when I use `@time` now, it does not seem to much of a difference. 

Either way, I would like to leave this here for discussion. If you think that this might hurt more than be useful, we can close it again. 

```julia
n=4
S, x = graded_polynomial_ring(QQ, ["x$i" for i in 0:n])

# f = [rand(S, 1:10, 1:10, 1:10) for _ in 1:n]
x0, x1, x2, x3, x4 = x
f = MPolyDecRingElem{QQFieldElem, QQMPolyRingElem}[9//5*x0^5 + 9//2*x0^3*x1^2 + 9//7*x0^3*x1*x2 + 5//2*x1*x2*x3^2*x4, 17//6*x0^3*x1 + 8//5*x0^3*x2 + 10//3*x0^2*x1^2 + 2*x0*x1^3 + 7//10*x0*x1*x2*x3 + x1^3*x4 + 5//6*x1*x3^2*x4, 5*x0^3*x1 + 10//7*x0^2*x2*x4 + 7//5*x0*x1^2*x3 + 1//2*x0*x1*x3*x4, 2//3*x0^6*x1 + 49//10*x0^6*x2 + 7*x0^3*x1^2*x2^2 + 2*x0^2*x1^3*x2*x3 + 8//3*x1^7 + 5//6*x1^2*x4^5]

K = Oscar.HomogKoszulComplex(S, f)

G = grading_group(S);
s, _ = Oscar.strand(K, 3*n*G[1]);
[ngens(s[i]) for i in 0:n]
density(sparse_matrix(map(s, 1)))

simp = simplify(s);
@time [ngens(simp[i]) for i in 0:n]
# without: 16.467451 seconds (17.35 M allocations: 1.081 GiB, 4.13% gc time, 0.17% compilation time)
# with these changes:  15.975815 seconds (18.59 M allocations: 1.099 GiB, 6.37% gc time, 0.16% compilation time)


P, t = QQ[:t];
A = sparse_matrix(map(s, 1));
length(A[1])
v = sum(v for v in A) # note that this is maybe a bad example as `v` is not really dense?
@time map_entries(P, v);
# without:   0.000309 seconds (3.03 k allocations: 149.547 KiB)
# with these changes:   0.000222 seconds (3.02 k allocations: 129.672 KiB)

c, _ = change_base_ring(P, s);
@time [map(c, i) for i in 1:n];
# without: 0.042210 seconds (494.22 k allocations: 22.887 MiB, 65.59% compilation time)
# with these changes:   0.042203 seconds (494.22 k allocations: 22.887 MiB, 65.79% compilation time)
```

thofma:
I remember trying this in the past and did not see great improvements. Since on the other hand it does not make things slower, I am happy merge this. Any objections @fieker?

PR #2014, opened 2025-09-24
define `modord(a, 1) = 1`

ThomasBreuer:
(Currently `modord(a, 1)` runs into an infinite loop.)

The definition makes sense if one interprets the mult. order of a mod m as the smallest positive integer d such that m divides a^d-1.


thofma:
Could you add a test to test/Misc/Integer.jl?

PR #2021, opened 2025-09-26
Add singleton sparse row constructor

HechtiDerLachs:
This is an attempt at #5371 on Oscar.jl.

### DIFF
--- a/src/Sparse/Row.jl
+++ b/src/Sparse/Row.jl
@@ -116,6 +116,17 @@ function sparse_row(R::NCRing, A::Vector{Tuple{Int, Int}}; sort::Bool = true)
   return SRow(R, A)
 end
 
+@doc raw"""
+    sparse_row(idx::Int, coeff::T; check::Bool=true) where {T}
+
+Constructs a sparse row with at most one non-zero entry `coeff` in 
+position `idx`.
+"""
+function sparse_row(idx::Int, coeff::T; check::Bool=true) where {T}
+  check && is_zero(coeff) && return SRow(parent(coeff))
+  return SRow(parent(coeff), Int64[idx], T[coeff]; check=false)
+end
+
 function Base.empty!(A::SRow)
   empty!(A.pos)
   empty!(A.values)
--- a/test/Sparse/Row.jl
+++ b/test/Sparse/Row.jl
@@ -32,6 +32,15 @@
   @test_throws ErrorException sparse_row(ZZ, [2, 2], [ZZRingElem(1), ZZRingElem(2)])
   @test_throws ErrorException sparse_row(ZZ, [2, 2], [ZZRingElem(1), ZZRingElem(2)])
 
+  G = @inferred sparse_row(1, one(R))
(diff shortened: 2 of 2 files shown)


src/Sparse/Row.jl
```diff
@@ -116,6 +116,22 @@ function sparse_row(R::NCRing, A::Vector{Tuple{Int, Int}}; sort::Bool = true)
...
 
+@doc raw"""
+    sparse_row(idx::Int, coeff::T; check::Bool=true) where {T}
+
+Constructs a sparse row with at most one non-zero entry `coeff` in 
+position `idx`.
+"""
+function sparse_row(idx::Int, coeff::T; check::Bool=true) where {T}
+  check && is_zero(coeff) && return SRow(parent(coeff))
+  return SRow(parent(coeff), Int64[idx], T[coeff]; check=false)
```
HechtiDerLachs: Note that `check=false` is necessary here in order to not throw away the already allocated vectors. 

I do not know what to do in the case of integers. The constructor does not seem to accept a `Vector` of `ZZRingElem`s in that case, so I just defaulted this to a dummy constructor for the moment. Maybe someone with more insight can adjust the constructor below to also be performant.

thofma:
```suggestion
  return SRow(parent(coeff), Int[idx], T[coeff]; check=false)
```

fingolfin: Well, see `src/Sparse/ZZRow.jl`, line 318 and following:
```julia
sparse_inner_type(::Type{ZZRingElem}) = ZZRingElem_Array

function sparse_row(R::ZZRing)
  return SRow(R, Int[], ZZRingElem_Array())
end
```
and then again line 489 and following:
```julia
for T in (Int, ZZRingElem)
  @eval begin
    function sparse_row(R::ZZRing, A::Vector{Tuple{Int, $T}}; sort::Bool = true)
      if sort && length(A) > 1
        A = Base.sort(A, lt=(a,b)->isless(a[1], b[1]))
      end
      a = ZZRingElem_Array()
...
```

Based on this I think something like this might work, in a separate `sparse_row` method for `ZZRingElem`:
```julia
  arr = ZZRingElem_Array(1)
  arr[1] = coeff
  return SRow(ZZ, Int[idx], arr)
```

src/Sparse/Row.jl
```diff
@@ -116,6 +116,22 @@ function sparse_row(R::NCRing, A::Vector{Tuple{Int, Int}}; sort::Bool = true)
...
+end
+
+# For ZZRingElems this has to be overwritten, because the values are stored differently. 
+function sparse_row(idx::Int, coeff::ZZRingElem; check::Bool=true)
+  return sparse_row(parent(coeff), [(idx, coeff)])
```
fingolfin:
```suggestion
  arr = ZZRingElem_Array(1)
  arr[1] = coeff
  return SRow(parent(coeff), Int[idx], arr)
```

thofma:
Thanks!

PR #2025, opened 2025-10-03
feat: make stable_subgroups faster

thofma:
- use minimal_submodules iteratively
- use a dictionary


fieker:
I suspect the files are missing

PR #2026, opened 2025-10-04
sb/overlattices and genera

simonbrandhorst:
- **primitive_extension**
- **add to docs**

### DIFF
--- a/src/QuadForm/Quad/ZLattices.jl
+++ b/src/QuadForm/Quad/ZLattices.jl
@@ -2864,6 +2864,67 @@ function glue_map(L::ZZLat, S::ZZLat, R::ZZLat; check=true)
 end
 
 @doc raw"""
+    primitive_extension(glue_map::TorQuadModuleMap)
+
+Given the glue map of a primitive extension of $\mathbb Z$-lattices
+$S \oplus R \subseteq L$, return `L` and the inclusions of
+$S\otimes \QQ $ and $R \otimes \QQ$ into $L \otimes \QQ$.
+
+# Example
+
+We construct the $E_8$ root lattice as a primitive extension of
+the $E_6$ and $A_2$ root lattice.
+```jldoctest
+julia> R = root_lattice(:A,2);
+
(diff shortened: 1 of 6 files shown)


src/QuadForm/Quad/ZLattices.jl
```diff
@@ -2863,6 +2863,71 @@ function glue_map(L::ZZLat, S::ZZLat, R::ZZLat; check=true)
   return glue_map, iS, iR
 end
 
+@doc raw"""
+    primitive_extension(glue_map::TorQuadModuleMap)
```
StevellM, line 2867:
```suggestion
    primitive_extension(glue_map::TorQuadModuleMap) -> ZZLat, QQMatrix, QQMatrix
```

src/QuadForm/Quad/ZLattices.jl
```diff
@@ -2908,21 +2973,21 @@ true
 function overlattice(glue_map::TorQuadModuleMap)
   S = relations(domain(glue_map))
   R = relations(codomain(glue_map))
+  @req ambient_space(S)==ambient_space(R) "lattices lie in different ambient spaces, try `primitive_extension` instead"
```
StevellM:
```suggestion
  @req ambient_space(S) === ambient_space(R) "Lattices lie in different ambient spaces, try `primitive_extension` instead"
```
Here we want a strong equality I guess.. It should by default, but just to be sure

StevellM (approved):
Looks good, thank you very much for these functions!

src/QuadForm/Quad/ZLattices.jl
simonbrandhorst: We are actually returning an `AbsSpaceMor`.   But I think it can be omitted.

StevellM: Ah right, I forgot this type existed. Well if it is clear from the documentation what the output is, maybe we can omit it.

PR #2033, opened 2025-10-08
Use map_coefficients instead of soon-to-be-forbidden coercion

JohnAAbbott:
Soon Nemo will be updated so that "coercion" of polynomials has to be effected via `map_coefficients`; this update is intended to avoid consequent problems in Hecke.


src/NumField/NfAbs/Elem.jl
```diff
@@ -390,10 +391,10 @@ function factor_trager(f::PolyRingElem{AbsSimpleNumFieldElem})
 #    d = mapreduce(x->denominator(x, A), lcm, coefficients(g), init = ZZRingElem(1))
 #    @vtime :PolyFactor 2 N = norm_mod(g*d, Zx)
 #    N = Hecke.Globals.Qx(N) * QQFieldElem(1, d)^degree(K)
-    @vtime :PolyFactor 2 N = Hecke.Globals.Qx(norm(g))
+    @vtime :PolyFactor 2 N = map_coefficients(identity,norm(g); parent=Qx)
```
thofma:
```suggestion
    @vtime :PolyFactor 2 N = map_coefficients(identity, norm(g); parent=Qx)
```

PR #2038, opened 2025-10-13
Fix scale(::HermGenus) and speed up representative(::HermGenus) for rank 1

simonbrandhorst:
- **...**
- **speedup, bugfix**

### DIFF
--- a/src/QuadForm/Herm/Genus.jl
+++ b/src/QuadForm/Herm/Genus.jl
@@ -2024,24 +2024,27 @@ function rescale(g::T, a::Union{FieldElem, RationalUnion}) where {T<:HermLocalGe
   a = K(a)
 
   val = valuation(a, p)
-  b = is_local_norm(E, a, p)
+  # Copy the relevant
+  # attributes/properties of g
   G = T()
   G.E = g.E
   G.p = g.p
   G.is_ramified = g.is_ramified
   G.is_split = g.is_split
   G.is_dyadic = g.is_dyadic
-
   if isdefined(g, :non_norm_rep)
     G.non_norm_rep = g.non_norm_rep
   end
   if isdefined(g, :P)
     G.P = g.P
   end
-  if isdefined(g, :norm_val)
+
+  # rescaled copies of the data
+  if is_dyadic(g) && is_ramified(g)
     G.norm_val = g.norm_val .+ val
   end
   G.data = typeof(g.data)()
+  b = is_local_norm(E, a, p)
   for v in g.data
     if !b && isone(mod(v[2], 2))
       det = -v[3]
@@ -2050,8 +2053,6 @@ function rescale(g::T, a::Union{FieldElem, RationalUnion}) where {T<:HermLocalGe
     end
     if is_split(g)
       push!(G.data, (v[1]+ val, v[2], v[3]))
-    elseif is_dyadic(g) && is_ramified(g)
-      push!(G.data, (v[1]+ 2*val, v[2], det, v[4]))
     elseif is_ramified(g)
       push!(G.data, (v[1]+ 2*val, v[2], det))
     else


src/QuadForm/Herm/Genus.jl
```diff
@@ -1951,10 +2014,55 @@ Given a local genus symbol `G` of hermitian lattices and an element `a` lying
...
+  val = valuation(a, p)
+  b = is_local_norm(E, a, p)
+  G = T()
+  G.E = g.E
+  G.p = g.p
+  G.is_ramified = g.is_ramified
+  G.is_split = g.is_split
+
+  if isdefined(g, :non_norm_rep)
+    G.non_norm_rep=G.non_norm_rep
```
StevellM:
```suggestion
    G.non_norm_rep = g.non_norm_rep
```

src/QuadForm/Herm/Lattices.jl
```diff
@@ -748,15 +749,23 @@ function maximal_integral_lattice(V::HermSpace)
     if i !== nothing
       i = i + 1 # findfirst gives the index and not the element
       @assert fac[i][2] == fac[1][2]
-      s = s * fractional_ideal(S, fac[1][1])^fac[1][2]
+      s = s * inv(fac[1][1])^fac[1][2]
```
StevellM, line 752: Why is this better ?

src/QuadForm/Herm/LocallyIsometricSublattice.jl
```diff
@@ -293,17 +293,18 @@ Given rationally isometric hermitian lattices `M` and `L` over $E/K$ and a prime
 of $\mathcal O_K$, return a sublattice `N` of `M` with $N_q = M_q$ for $q \neq p$ and
 $N_p$ isometric to $L_p$.
 """
```
StevellM, line 295: The changes in this file, which are part of your first commit, seems to be part of another PR of yours. It creates conflicts

src/QuadForm/Herm/Genus.jl
```diff
@@ -1308,6 +1335,7 @@ Return the rank of any hermitian lattice with global genus symbol `G`.
 """
 rank(G::HermGenus) = G.rank
 
+# not quite the scale at the split primes (their conjugate is missing)
```
StevellM: The function below looks at what happens on the lower field `K` of the extension `E/K` so there should not be any conjugate "missing". I guess this function returns $s(L) \cap O_K$ for any $L$ in the genus `G` (or something of that sort)

src/QuadForm/Herm/Genus.jl
```diff
@@ -1951,10 +2014,55 @@ Given a local genus symbol `G` of hermitian lattices and an element `a` lying
...
+      det = v[3]
+    end
+    if is_split(g)
+      push!(G.data, (v[1]+ val, v[2], v[3]))
+    elseif is_dyadic(g) && is_ramified(g)
+      push!(G.data, (v[1]+ 2*val, v[2], det, v[4]))
+    elseif is_ramified(g)
+      push!(G.data, (v[1]+ 2*val, v[2], det))
+    else
+      push!(G.data, (v[1]+ val, v[2], det))
```
StevellM, line 2059: Is it clear why the determinant class does not change in the inert case ?

src/QuadForm/Herm/Genus.jl
simonbrandhorst: Mhh, right at the split primes and at the inert primes it is $s(L) \cap O_K$. 
But at the ramified primes it is not the scale but `s(L_p)^2 \cap O_K`. 

```suggestion
# not quite the scale at the ramified primes 
```

src/QuadForm/Herm/Genus.jl
simonbrandhorst: It should change in the inert case. But in the split case everything is a norm so there it is always `1`.

src/QuadForm/Herm/Lattices.jl
simonbrandhorst: Because previously it was wrong. There was no `inv`. 
We are replacing `L` by `s*L` in such a way that it has maximum possible integral scale.

StevellM: Ah yes sorry, I completely missed the point of inverting, I thought the problem was the `fractional_ideal` constructor

src/QuadForm/Herm/Genus.jl
```diff
@@ -100,8 +100,41 @@ lying over $\mathfrak p$.
...
+  end
+  g.P = P
+  return P
+end
+
+function prime_over(g::HermLocalGenus{S,T,U}) where {S,T,U}
+  if is_split(g)
+    return first(factor(maximal_invariant_ideal(g)))[1]
+  end
+  return maximal_invariant_ideal(g)
```
StevellM:
```suggestion
  P = maximal_invariant_ideal(g)
  P.is_prime = 1
  return P
```
or something like this... This new way of getting `P` does not remember `P` being prime and this breaks `valuation` later

StevellM: maybe `set_attribute!` would be a more stable solution

simonbrandhorst: I think is rather a bug in valuation. Because it does not use the interface.

simonbrandhorst:
```suggestion
  P =  maximal_invariant_ideal(g)
  @assert is_prime(P)
  return P
```

StevellM: Probably. We already had the discussion at some point, if I remember correctly, but I do not remember why it was decided to keep it as is.

src/QuadForm/Herm/Genus.jl
```diff
@@ -1951,10 +2016,58 @@ Given a local genus symbol `G` of hermitian lattices and an element `a` lying
...
+  a = K(a)
+
+  val = valuation(a, p)
+  b = is_local_norm(E, a, p)
+  G = T()
+  G.E = g.E
+  G.p = g.p
+  G.is_ramified = g.is_ramified
+  G.is_split = g.is_split
+
```
StevellM:
```suggestion
G.is_dyadic = g.is_dyadic

```

StevellM: CI is breaking because of a dyadic ramified symbol not having 4 invariants par scale, which seems odd. I do not know if that solves the issue, probably not, but at least it was missing in the list of known data

src/QuadForm/Herm/Genus.jl
```diff
@@ -1951,10 +2016,59 @@ Given a local genus symbol `G` of hermitian lattices and an element `a` lying
...
+  a = K(a)
+
+  val = valuation(a, p)
+  b = is_local_norm(E, a, p)
+  G = T()
+  G.E = g.E
+  G.p = g.p
+  G.is_ramified = g.is_ramified
+  G.is_split = g.is_split
+G.is_dyadic = g.is_dyadic
```
simonbrandhorst:
```suggestion
  G.is_dyadic = g.is_dyadic
```

PR #2040, opened 2025-10-14
fix: group algebras over non-fields

thofma:


thofma:
Forgot that I forgot to disable automerge. I am also not too happy with this. The only other option is adding `_dim` or `rank`. Using `_dim` is bad, because no one will remember it when writing algebra code. On the other hand, no one will remember `rank`, since it's no terminology people would use.

lgoettgens:
I would say that `rank` would be the right terminology when talking about group rings over a ring (and not over a field).

PR #2041, opened 2025-10-14
Added new combinatorics functionalities

Tobias271828:
Together with Simon, we discussed the need of an iterator 'partition_with_condition(n, k, l)'. This iterator enumerates all possible k-tuples [ a_0, ..., a_{k-1} ] of non-negative integers satisfying a_0 + a_1 + ... + a_{k-1} = n and, furthermore, 0*a_0 + 1*a_1 + 2*a_2 + ... + (k-1)*a_{k-1} = l.

### DIFF
--- a/test/Misc/PartitionsWithCondition.jl
+++ b/test/Misc/PartitionsWithCondition.jl
@@ -9,11 +9,10 @@
 
   @test iterate(Hecke.partitions_with_condition(2,1,1)) == nothing
 
-  n = 0
-  for i in Hecke.partitions_with_condition(7, 4, 12)
-    n += 1
-  end
-  @test n == 5
+  L = [copy(i) for i in Hecke.partitions_with_condition(7, 4, 12)]
+  @test length(L) == 10
+  @test all(sum(i)==7 for i in L)
+  @test all(sum(i[j]*(j-1) for j in 1:4)==12 for i in L)
 
   n = 0
   for i in Hecke.partitions_with_condition(2, 2, 1)


src/Combinatorics/PartitionWithCondition.jl
```diff
@@ -0,0 +1,79 @@
...
+# This file provides an iterator 'partition_with_condition(n, k, l)'. This
+# iterator enumerates all possible k-tuples [ a_0, ..., a_{k-1} ] of non-
+# negative integers satisfying a_0 + a_1 + ... + a_{k-1} = n and, further-
+# more, 0*a_0 + 1*a_1 + 2*a_2 + ... + (k-1)*a_{k-1} = l.
+#
+# Example
+# julia> for i in partition_with_condition(5, 3, 7) println(i) end
+# [5, 0, 0]
+# [0, 3, 2]
+# [1, 1, 3]
```
simonbrandhorst: Please put this kind of description as a docstring of the struct (or its constructor). Then it is easier to find.

src/Combinatorics/PartitionWithCondition.jl
```diff
@@ -0,0 +1,79 @@
...
+    a1 = parti.l - weighted_sum
+    if a1 >= 0
+      a0 = parti.n - sum - a1
+      if a0 >= 0
+        iter[1][3:end] = view(iter[2], 3:parti.k)
+        @inbounds iter[1][2] = a1
+        @inbounds iter[1][1] = a0
+        @inbounds iter[2][1] = weighted_sum
+        @inbounds iter[2][2] = sum
+        return iter[1], iter
```
simonbrandhorst:
```
julia> @btime b[1:3] = view(a,1:3)
  58.579 ns (1 allocation: 48 bytes)
3-element view(::Vector{Int64}, 1:3) with eltype Int64:
 1
 2
 3

julia> @btime b[1:3] = a[1:3]
  79.533 ns (2 allocations: 80 bytes)
3-element Vector{Int64}:
 1
 2
 3

julia> @btime for i in 1:3 b[i] = a[i] end
  139.125 ns (0 allocations: 0 bytes)

```
Creating a view seems like the fastest but at the cost of one allocation.  My suggestion here is to avoid the slicing of the arrays completely:

Give `partition_with_condition` 4 extra fields:
`sum::Int`
`weighted_sum::Int`
`current_position::Int`
`a::Vector{Int}`
Then use these fields for everything.

src/Combinatorics/PartitionWithCondition.jl
```diff
@@ -0,0 +1,79 @@
...
+#
+# Example
+# julia> for i in partition_with_condition(5, 3, 7) println(i) end
+# [5, 0, 0]
+# [0, 3, 2]
+# [1, 1, 3]
+#
+#########################################################################
+
+struct partition_with_condition
```
simonbrandhorst: By convention types are CamelCase
```suggestion
struct PartitionWithCondition
```

Tobias271828: Simply changing the type name to CamelCase results in UndefVarErrors. Thus, we would have to change the name of the constructor function to CamelCase as well. This, however, implies that
`for i in partition_with_condition(5, 3, 7) println(i) end`
throws an error, while only
`for i in PartitionWithCondition(5, 3, 7) println(i) end`
works.

If I understood it correctly, iterators are usally denoted in snake_case. Is that correct? If so, we would break conventions anyways. Which way out would you recommend?

src/Combinatorics/PartitionWithCondition.jl
Tobias271828: This is done now.

src/Misc/PartitionWithCondition.jl
```diff
@@ -0,0 +1,99 @@
...
+  
+  This code provides an iterator ``partition_with_condition(n, k, l)``. This
+  iterator enumerates all possible ``k``-tuples ``[ a_0, ..., a_{k-1} ]`` of
+  nonnegative integers satisfying ``a_0 + a_1 + ... + a_{k-1} = n`` and,
+  furthermore, ``0*a_0 + 1*a_1 + 2*a_2 + ... + (k-1)*a_{k-1} = l``.
+
+  # Examples
+  ```jldoctest
+  julia> for i in partition_with_condition(5, 3, 7) println(i) end
+  [5, 0, 0]
```
simonbrandhorst: Looks to me like the first return value does not match the specification?

src/Combinatorics/PartitionWithCondition.jl
Tobias271828: This is corrected now.

src/Misc/PartitionWithCondition.jl
Tobias271828: Thank you very much for reporting this but! It is due to a small, but urgent but. It should be fixed now.
The test set and the doc are adjusted and should also be correct now.

src/Combinatorics/PartitionWithCondition.jl
Tobias271828: Thank you! This really improved the performance.

thofma:
We need to add `partition_with_condition` to the list in src/exports.jl.

src/Misc/PartitionWithCondition.jl
```diff
@@ -0,0 +1,114 @@
...
+  # weighted_sum = iter[2][1]
+  # sum = iter[2][2]
+  # For every choice of integers a_2, a_3, ... a_{k-1} there exists a unique pair of integers
+  # (a_0, a_1) fulfilling the two equations. Hence, we can iterate over all such choices and
+  # calculate the corresponding pair (a_0, a_1) and check, whether a_0 and a_1 are both >= 0.
+  while current_position < parti.k+1
+    parti.sum += 1
+    parti.weighted_sum += current_position - 1
+    # @inbounds iter[2][current_position] += 1
+    parti.vect[current_position] += 1
```
simonbrandhorst:
```suggestion
    @inbounds parti.vect[current_position] += 1
```

src/Misc/PartitionWithCondition.jl
```diff
@@ -0,0 +1,114 @@
...
+        # @inbounds iter[1][1] = a0
+        # @inbounds iter[2][1] = parti.weighted_sum
+        # @inbounds iter[2][2] = parti.sum
+        return parti.vect, nothing
+      end
+    else
+      @inbounds parti.sum -= parti.vect[current_position]
+      @inbounds parti.weighted_sum -= (current_position - 1) * parti.vect[current_position]
+      # @inbounds iter[2][current_position] = 0
+      parti.vect[current_position] = 0
```
simonbrandhorst:
```suggestion
      @inbounds parti.vect[current_position] = 0
```

src/Misc/PartitionWithCondition.jl
```diff
@@ -0,0 +1,114 @@
...
+    if a1 >= 0
+      a0 = parti.n - parti.sum - a1
+      if a0 >= 0
+        @inbounds parti.vect[1] = a0
+        @inbounds parti.vect[2] = a1
+        # iter[1][3:end] = view(iter[2], 3:parti.k)
+        # @inbounds iter[1][2] = a1
+        # @inbounds iter[1][1] = a0
+        # @inbounds iter[2][1] = parti.weighted_sum
+        # @inbounds iter[2][2] = parti.sum
```
simonbrandhorst: please remove stale comments

simonbrandhorst:
We reached a speedup of a factor of about 1000 from the first draft of this function. I gave all the advice I could to get this function as fast as possible. Is there anything else we could do to improve it? 
Maybe @thofma  @fingolfin could take a short look and tell if we overlooked anything. 
At this point the function is fast enough for my purposes, but it would be nice to learn new tricks how to code better. (In case we need a really efficient implementation the next time.)

src/Misc/PartitionWithCondition.jl
Tobias271828: Stale Comments are removed now

src/Misc/PartitionWithCondition.jl
Tobias271828: Thanks! This is resolved now.

src/Misc/PartitionWithCondition.jl
Tobias271828: Thanks! This is resolved now.

test/Misc/PartitionsWithCondition.jl
```diff
@@ -0,0 +1,17 @@
...
+  a = partition_with_condition(2, 1, 0)
+  @test iterate(a,iterate(a)[2]) == nothing
+
+  @test iterate(partition_with_condition(2,1,1)) == nothing
+
+  n = 0
+  for i in partition_with_condition(7, 4, 12)
+    n += 1
+  end
+  @test n == 5
```
simonbrandhorst: Can this be shortened as follows?

```suggestion
  @test length(partition_with_condition(7, 4, 12)) == 5
```

src/Misc/PartitionWithCondition.jl
```diff
@@ -0,0 +1,104 @@
...
+    if parti.l == 0
+      return Int[parti.n], nothing
+    else
+      return nothing
+    end
+  end
+  # The iterators hidden output 'iter' is a tuple of two integer vectors of the same size.
+  # The first vector is the last non-hidden output, i.e. the vector [a_0, ..., a_{k-1}].
+  # The second vector is of the form [W, S, a_2, a_3, ..., a_{k-1}], where S = a_2 + ... + a_{k-1}
+  # and W = 2*a_2 + 3*a_3 + ... + (k-1)*a_{k-1}.
```
simonbrandhorst: This comment seems outdated to me. 
Perhaps move parts of it to the `struc`?

src/Misc/PartitionWithCondition.jl
```diff
@@ -0,0 +1,104 @@
+@doc raw"""
+  partition_with_condition(
```
simonbrandhorst: Would you please move this docstring to the actual function `partition_with_condition` in line 50?
Then if someone types `?partition_with_condition` they will see the docstring.

src/Misc/PartitionWithCondition.jl
Tobias271828: Yes. It is resolved now.

src/Misc/PartitionWithCondition.jl
Tobias271828: I didn't know that before. Thank you!

src/Misc/PartitionWithCondition.jl
```diff
@@ -0,0 +1,103 @@
...
+  end
+
+end
+
+@doc raw"""
+  ``partition_with_condition(
+    n::Int,
+    k::Int,
+    l::Int
+  ) -> iterator``
```
simonbrandhorst:
```suggestion
    partition_with_condition(
    n::Int,
    k::Int,
    l::Int
  ) -> iterator
```

src/Misc/PartitionWithCondition.jl
```diff
@@ -0,0 +1,103 @@
...
+  
+  This code provides an iterator ``partition_with_condition(n, k, l)``. This
+  iterator enumerates all possible ``k``-tuples ``[ a_0, ..., a_{k-1} ]`` of
+  nonnegative integers satisfying ``a_0 + a_1 + ... + a_{k-1} = n`` and,
+  furthermore, ``0*a_0 + 1*a_1 + 2*a_2 + ... + (k-1)*a_{k-1} = l``.
```
simonbrandhorst: The following is more concise.

```suggestion
  Return an iterator over all possible ``k``-tuples ``[ a_0, ..., a_{k-1} ]`` of
  nonnegative integers satisfying ``a_0 + a_1 + ... + a_{k-1} = n`` and,
  furthermore, ``0*a_0 + 1*a_1 + 2*a_2 + ... + (k-1)*a_{k-1} = l``.
```

src/Misc/PartitionWithCondition.jl
```diff
@@ -0,0 +1,103 @@
...
+
+  # Examples
+  ```jldoctest\
+  julia> for i in partition_with_condition(5, 3, 7) println(i) end\
+  [0, 3, 2]\
```
simonbrandhorst: What is the `\` needed for? I do not remember having seen it in other docstrings.

src/Misc/PartitionWithCondition.jl
simonbrandhorst: As far as I recall these backticks are not needed.

src/Misc/PartitionWithCondition.jl
```diff
@@ -0,0 +1,105 @@
...
+    partition_with_condition(
+      n::Int,
+      k::Int,
+      l::Int
+    ) -> iterator
+
+This code provides an iterator `partition_with_condition(n, k, l)`. This
+iterator enumerates all possible `k`-tuples `[ a_0, ..., a_{k-1} ]` of
+nonnegative integers satisfying `a_0 + a_1 + ... + a_{k-1} = n` and,
+furthermore, `0*a_0 + 1*a_1 + 2*a_2 + ... + (k-1)*a_{k-1} = l`.
```
simonbrandhorst:
```suggestion
Return an iterator which enumerates all possible `k`-tuples `[ a_0, ..., a_{k-1} ]` of
nonnegative integers satisfying `a_0 + a_1 + ... + a_{k-1} = n` and,
furthermore, `0*a_0 + 1*a_1 + 2*a_2 + ... + (k-1)*a_{k-1} = l`.
```

simonbrandhorst (approved):

thofma:
I pushed a minor improvement (10% when benchmarking) and a fix to the docstring. In general, just look at how the other docstrings look like and copy the style (including the indentation).

I don't see any other obvious improvements. According to the profiler all the time is spent in `getindex`, `setindex` and arithmetic, which is how it should be.

thofma:
The function should be called `partitions_with_condition`. I will fix it myself.

src/Misc/PartitionsWithCondition.jl
```diff
@@ -0,0 +1,104 @@
...
+    parti.sum += 1
+    parti.weighted_sum += current_position - 1
+    @inbounds parti.vect[current_position] += 1
+    a1 = parti.l - parti.weighted_sum
+    if a1 >= 0
+      a0 = parti.n - parti.sum - a1
+      if a0 >= 0
+        @inbounds parti.vect[1] = a0
+        @inbounds parti.vect[2] = a1
+        return parti.vect, nothing
```
fingolfin, line 109: It seems you implemented a stateful iterator, i.e. iterating modifies the iterator object itself. That's a potential source of bugs in code. But it seems not so hard to change this? Just move `sum`, `weighted_sum` and `vect` out of the structure (which then can become immutable), and use them for the state (e.g. in a named tuple)

src/Misc/PartitionWithCondition.jl
Tobias271828: I thought, these were necessary for the line breaks. But it was due to indentation and other stuff. Thommy fixed it.

src/Misc/PartitionsWithCondition.jl
fingolfin: Or just a plain tuple, then the first line can become
```
  sum, weighted_sum, vec = state
``` 
or so, and here you
```suggestion
        @inbounds parti.vect[1] = a0
        @inbounds parti.vect[2] = a1
        return parti.vect, (sum, weighted_sum, vec)
```

src/Misc/PartitionsWithCondition.jl
```diff
@@ -0,0 +1,104 @@
...
+  else
+    return iterate(parti, nothing)
+  end
+end
+
+function Base.iterate(parti::PartitionsWithCondition, xxx::Nothing)
+  if parti.k == Int(1)
+    return nothing
+  end
+  current_position = Int(3)
```
fingolfin: Small integer literals are of type `Int` anyway, why the explicit `Int` ?
```suggestion
  current_position = 3
```

Tobias271828: Because I didn't knew this so far. Thank you!

PR #2045, opened 2025-10-20
Fix bug in #2044 and add more comments

JHanselman:
@simonbrandhorst This should take care of the broken test.


simonbrandhorst:
Thanks :-).

PR #2047, opened 2025-10-21
Preparation for OSCAR `isometry_group_via_heuristics`

StevellM:
- Lattices with a fixed basis matrix which is LLL-reduced remembers it (so we can avoid repeating the computation and pass it to relevant functions... we allow to redo an LLL-reduction though,  if one wants to change reduction parameters for instance)
- computations of automorphism group generators and order remember that we have already computed an LLL-reduced basis (currently, it recomputes an LLL-reduction which does not make sense)
- allow to input a list of all short vectors (up to sign) of bounded square if they have been previously computed
- allow to specify a lower bound in `_short_vectors_gram_integral`

Related to the heuristics that we are trying to fine-tune in https://github.com/oscar-system/Oscar.jl/pull/5434.

### DIFF
--- a/src/QuadForm/Morphism.jl
+++ b/src/QuadForm/Morphism.jl
@@ -149,7 +149,7 @@ function init(
   depth::Int=-1,
   bacher_depth::Int=0,
   is_lll_reduced_known::Bool=false,
-  known_short_vectors=(QQFieldElem(0), Tuple{Vector{ZZRingElem}, QQFieldElem}[]),
+  known_short_vectors=(0, []),
 )
   # Compute the necessary short vectors
 
@@ -164,10 +164,13 @@ function init(
 
   @assert bound > 0
 
-  @vprintln :Lattice 1 "Computing short vectors of length $bound"
-  # If one already knows all the short vectors of square at most equal to alpha
+  @vprintln :Lattice 1 "Computing short vectors of length <= $bound"
+  # If one already knows all the short vectors of length at most equal to alpha
   alpha, V = known_short_vectors
 
+  # If _V is not empty, then it should contain (up to sign) all the short vectors
+  # of length at most equal to alpha. So if alpha is lower than bound, we add the
+  # missing vectors
   if alpha < bound
     @vtime :Lattice 1 append!(V, _short_vectors_gram_integral(Vector, C.G[1], alpha+1, bound; is_lll_reduced_known))
   end
@@ -308,7 +311,7 @@ function try_init_small(
   depth::Int=-1,
   bacher_depth::Int=0,
   is_lll_reduced_known::Bool=false,
-  known_short_vectors=(QQFieldElem(0), Tuple{Vector{ZZRingElem}, QQFieldElem}[]),
+  known_short_vectors=(0, []),
  )
   automorphism_mode = bound == ZZRingElem(-1)
 
@@ -327,15 +330,17 @@ function try_init_small(
   @assert bound > 0
 
   # Compute the necessary short vectors
-  @vprintln :Lattice 1 "Computing short vectors of length $bound"
-  # If one already knows all the short vectors of square at most equal to alpha
-  alpha, _V = known_short_vectors
-  V = Tuple{Vector{Int}, QQFieldElem}[(Int.(v[1]), v[2]) for v in _V]
+  @vprintln :Lattice 1 "Computing short vectors of length <= $bound"
+  # If one already knows all the short vectors of length at most equal to alpha
+  alpha, V = known_short_vectors
+  @assert all(Base.Fix2(isa, Vector{Int})∘first, V)
 
-  if alpha < bound
+  # If _V is not empty, then it should contain (up to sign) all the short vectors
+  # of length at most equal to alpha. So if alpha is lower than bound, we add the
+  # missing vectors
+  if alpha <= bound
     @vtime :Lattice 1 append!(V, _short_vectors_gram_integral(Vector, C.G[1], alpha+1, bound, Int; is_lll_reduced_known))
   end
-
   vectors = Vector{Vector{Int}}(undef, length(V))
 
   lengths = Vector{Vector{Int}}(undef, length(V))
@@ -403,7 +408,6 @@ function try_init_small(
   end
 
   V = VectorList(vectors, lengths, use_dict)
-
 
   Csmall.V = V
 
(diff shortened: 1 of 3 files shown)


src/QuadForm/Quad/ZLattices.jl
```diff
@@ -2059,12 +2083,79 @@ function lll(L::ZZLat; same_ambient::Bool = true)
...
+    Llll = lattice(ambient_space(L), B2; check=false)
   else
-    return integer_lattice(; gram = (1//d)*change_base_ring(QQ, G2))
+    Llll = integer_lattice(; gram=(1//d)*change_base_ring(QQ, G2))
+  end
+  set_attribute!(Llll, :is_lll_reduced, true)
+  return Llll
+end
+
+function lll!(
```
simonbrandhorst: This kind of function is dangerous. Mutating an object can easily lead to bugs. (e.g. when something else depends on the basis matrix which is then altered )
I would vote for removing unless it is really needed.

src/QuadForm/Types.jl
```diff
@@ -534,6 +534,7 @@ mutable struct ZLatAutoCtx{S, T, V}
   prime::S
 
   is_symmetric::BitArray{1} # whether G[i] is symmetric
+  is_lll_reduced::BitArray{1} # whether G[i] is lll-reduced
```
fieker: what is the point of having 2 BitArray{1} in the structure? If you want to save space, I'd do a single BitArray{2} and use the 2 entries. Each BitArray{1} takes 64 bits, so this is currently equivalent to using Bool

src/QuadForm/Quad/ZLattices.jl
StevellM: You are right, I will remove it.

src/QuadForm/Types.jl
StevellM: I wanted to mimic the `is_symmetric` field, but eventually that's unnecessary: I can pass the info without storing it on the object. I will modify that.

thofma:
@fieker is there some misunderstanding? `BitArray{n}` is an n-dimensional array. `n` is not the length. `BitArray{1}` is the same as `BitVector`.

fieker:
Sorry, indeed.Forget most of my comments, please

src/QuadForm/Morphism.jl
```diff
@@ -306,7 +328,13 @@ function try_init_small(C::ZLatAutoCtx, auto::Bool = true, bound::ZZRingElem = Z
 
   # Compute the necessary short vectors
   @vprintln :Lattice 1 "Computing short vectors of length $bound"
-  @vtime :Lattice 1 V = _short_vectors_gram_integral(Vector, C.G[1], bound, Int)
+  # If one already knows all the short vectors of square at most equal to alpha
+  alpha, _V = known_short_vectors
+  V = Tuple{Vector{Int}, QQFieldElem}[(Int.(v[1]), v[2]) for v in _V]
```
simonbrandhorst: Previously this conversion was not necessary, now it is. Why? Is it justified? Will it slow things down?

simonbrandhorst: maybe require the input type of known_short_vectors to be Int already?

src/QuadForm/Morphism.jl
```diff
@@ -306,7 +328,13 @@ function try_init_small(C::ZLatAutoCtx, auto::Bool = true, bound::ZZRingElem = Z
...
+
+  if alpha < bound
```
simonbrandhorst: could you please document what alpha is? It is not obvious to me.

src/QuadForm/Quad/ZLattices.jl
```diff
@@ -391,30 +396,53 @@ function assert_has_automorphisms(L::ZZLat; redo::Bool = false,
...
     res[1] = -res[1]
   end
-  # Make the Gram matrix small
-  Glll, T = lll_gram_with_transform(res[1])
-  res[1] = Glll
+  if !get_attribute(L, :is_lll_reduced, false)
+    # Make the Gram matrix small
+    Glll, T = lll_gram_with_transform(res[1])
+    res[1] = Glll
+    sv = eltype(sv)[(solve(T, v[1]; side=left), v[2]) for v in sv]
```
simonbrandhorst: This need a solve_init because you solve with the same left hand side many times. 
But in any case: is this faster than recomputing the short vectors?

src/QuadForm/Morphism.jl
StevellM: Yes that would be better, for my application it will be the case anyway

src/QuadForm/Quad/ZLattices.jl
StevellM: I was not aware about this `solve_init`, I discovered it in your recent PRs. Anyway: in my application, this loop will never be reached, it was more a safeguard in case we modify the basis inbetween. So in the application I have in mind, if we pre-compute some short vectors, the lattice will have a reduced basis and so we won't call that part of the code. Otherwise I do not know whether it is slower than computing again short vectors.

simonbrandhorst: Please use the solve init anyways :-). Otherwise, it will put T in hermite normal form a thousand times.

src/QuadForm/Quad/ZLattices.jl
```diff
@@ -391,30 +396,51 @@ function assert_has_automorphisms(L::ZZLat; redo::Bool = false,
...
+  _alpha, sv = known_short_vectors
   V = ambient_space(L)
   GL = gram_matrix(L)
   d = denominator(GL)
-  res = ZZMatrix[change_base_ring(ZZ, d * GL)]
+  if !isone(d)
+    res = ZZMatrix[change_base_ring(ZZ, d * GL)]
+    if !iszero(_alpha)
+      alpha = abs(numerator(d*_alpha))
+      sv = eltype(sv)[(v[1], d*v[2]) for v in sv]
```
simonbrandhorst, line 407: is it guaranteed that `typeof(d*v[2]) == typeof(v[2])`?

simonbrandhorst (approved):

simonbrandhorst:
Thanks!

src/QuadForm/Quad/ZLattices.jl
StevellM: For the applications I have in mind, `_alpha` and `v[2]` should be of type `QQFieldElem` (worst case scenario, some are `ZZRingElem`), and so `d` is a `ZZRingElem` . So it should be fine in those cases. Because I take them from the output of `short_vectors` so everything should be a `QQFieldElem` technically.

Issue #2051, opened 2025-10-22
Norm equations sometimes large


thofma:
I know, I know, not clear if we can do something about this, but sometimes `is_norm` returns "large" elements:
```
julia> k, _a = Hecke.number_field(x^3 + 11*x^2 + 8*x - 1)
julia> ky, y = k[:y]
julia> l, _b = Hecke.number_field(y^3 + (3*_a - 2)*y^2 + (3*_a^2 - 4*_a - 8)*y - 13*_a^2 - 16*_a + 9)
# _, n1 = Hecke.is_norm(l, k(-2))
julia> n1 = (-94298461707737030787129//16384*_a^2 - 2080417641007139625138849//32768*_a - 1586977050802009659961479//32768)*_b^2 + (393422482194948133205409//16384*_a^2 + 2174529619712306886603447//8192*_a + 1702234347536301306440155//8192)*_b - 23232913862876524680321//2048*_a^2 - 2117168264649881482131805//16384*_a - 4493697201581404452050543//32768
julia> n2 = (5//8*_a^2 + 29//4*_a + 71//8)*_b^2 + (15//4*_a^2 + 79//2*_a + 47//4)*_b - 6*_a^2 - 511//8*_a - 133//8
julia> Hecke.norm(n1) == Hecke.norm(n2) == -2
```

Just posting it so that we don't forget.

Issue #2050, opened 2025-10-22
Norm equation and non-nice defining polynomial


thofma:
```
julia> QQx, x = Hecke.QQ[:x]
julia> f = x^2 - 1//2
julia> k, _ = Hecke.number_field(f)
julia> Hecke.is_norm(k, 2)
ERROR: Denominator has to be 1 ...
```

PR #2052, opened 2025-10-23
Cf/reduce neq

fieker:
should address #2051, #2050


src/NumFieldOrd/NfOrd/Clgp/Saturate.jl
```diff
@@ -519,10 +522,14 @@ function simplify(c::Hecke.ClassGrpCtx, U::Hecke.UnitGrpCtx, cp::Int = 0; use_LL
   new_rels = Vector{FacElem{AbsSimpleNumFieldElem, AbsSimpleNumField}}()
   vals_new_rels = Vector{SRow{ZZRingElem}}()
   @vprintln :Saturate 1 "Computing rels..."
+  scan = true
+  #basis is in HNF, the saturation only needs the "essential" bit,
+  #that is all starting at the first none-one entry
```
fieker: no longer convinced this is neccessary, this is not wrong, but too much work. Reasoning: if the matrix is, locally, (p 0 *; 0 1 *) then th eonly possibility of getting a p in 2nd position is to add p x the 2nd, but if this is then a p-th power, the 1st row alone would have already been...

PR #2065, opened 2025-10-27
Bug in small_period_matrix (RiemannSurface)

Tobias271828:
By testing and comparing with Magma, I found a bug in this code and added this example as a test (@test_broken).

### DIFF
--- a/test/RieSrf/RiemannSurface.jl
+++ b/test/RieSrf/RiemannSurface.jl
@@ -22,4 +22,10 @@
   RS = RiemannSurface(f, v)
   small_period_matrix(RS)
 
+  # ERROR: A larger example -- this throws an ArgumentError although it works perfectly fine in Magma
+  # Check example later again -- currently the crucial error bound handling is being processed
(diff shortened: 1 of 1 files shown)


test/RieSrf/RiemannSurface.jl
```diff
@@ -0,0 +1,35 @@
...
+  # an elliptic curve
+  f = x^3+1 - y^2
+  RS = RiemannSurface(f, v)
+  tau = small_period_matrix(RS)
+  R = base_ring(tau)
+  x = R("0.5 +/- 1e-10") + R("0.86602540378443864676372317 +/- 1.91e-10")*im
+  @test contains(x, tau[1,1])
+
+  # another elliptic curve
+  @test_broken f = x^3 + 2*x^2*y + 3*x - 4 + 5*y^2 # Redefining Kxy helps somehow? I don't understand what is going on here ...
```
lgoettgens: you are re-defining the variable `x` from line 5 (where it is a variable in a polynomial ring) in line 17 (where it is an `acb`). I would suggest renaming the variable in lines 17 and 18 (and later in 26 and 27) to something else

Tobias271828: Thank you for finding this! This is now fixed.

PR #2070, opened 2025-10-31
Renamed dense_poly_type to poly_type

JohnAAbbott:
This probably should require a new version of `AbstractAlgebra` ?  0.47.4, I believe

### DIFF
--- a/src/Deprecations.jl
+++ b/src/Deprecations.jl
@@ -7,7 +7,7 @@
 # ALL aliases here are only a temporary measure to allow for a smooth transition downstream.
 # they will be replaced by deprecations eventually
 
-#= End of list =#
+#= currently none =#
 
 ###############################################################################
 #


src/Deprecations.jl
```diff
@@ -7,7 +7,7 @@
 # ALL aliases here are only a temporary measure to allow for a smooth transition downstream.
 # they will be replaced by deprecations eventually
 
-#= currently none =#
+#= End of list =#
```
lgoettgens:
```suggestion
#= currently none =#
```
unnecessary change, better leave it as is

lgoettgens (approved):
The changes look good. CI currently fails due to Cloudflare issues

fingolfin:
Restart CI?

@JohnAAbbott this is still marked as draft but I think it is actually ready, isn't it?

fingolfin:
Another failure in a require test, looks like a crash (possibly caused by a timeout)?

Perhaps @thofma or @joschmitt or whoever has access rights here can restart just that job?

PR #2074, opened 2025-11-03
Fix a bug in gcd_sircana

fingolfin:
So far, we had:

    julia> using Hecke

    julia> R = residue_ring(ZZ,9)[1]
    Integers modulo 9

    julia> S, x = R[:x]
    (Univariate polynomial ring in x over R, x)

    julia> f = x^2 ; g = x^2-5*x+6 ;

    julia> Hecke.gcd_sircana(f,g)
    (x + 6, x + 3, x + 7)

    julia> Hecke.gcd_sircana(f,g*x)
    ERROR: Not an element of the domain
    ...

The last error was due to "lifting" from Z/p to Z/p^2 "directly" does not work anymore. We now lift to ZZ and project down.

Alas, this is only gets us to the next error. But at least it is some minor progress.
```julia-repl
julia> Hecke.gcd_sircana(f,g*x)
ERROR: DomainError with -1:
Index must be non-negative
Stacktrace:
 [1] coeff(x::ZZModPolyRingElem, n::Int64)
   @ Nemo ~/.julia/packages/Nemo/kdloy/src/flint/fmpz_mod_poly.jl:49
 [2] fun_factor(f::ZZModPolyRingElem)
   @ Hecke ~/Projekte/OSCAR/Hecke.jl/src/Misc/nmod_poly.jl:1047
 [3] gcd_sircana(f::zzModPolyRingElem, g::zzModPolyRingElem)
   @ Hecke ~/Projekte/OSCAR/Hecke.jl/src/Misc/nmod_poly.jl:829
 [4] top-level scope
   @ REPL[30]:1
```
What happens here is that `fun_factor` is invoked on `6*x` -- but that's a nilpotent, and my understanding is that `fun_factor` is for arguments which are not nilpotent but have a nilpotent lead coefficient.

In fact at that point `(fp, gp) = (x^2, 6*x)`.

CC @fieker 

Motivation: <https://mathoverflow.net/q/500571/8338> and <https://math.stackexchange.com/q/5096453/12415>.

### DIFF
--- a/test/Misc/nmod_poly.jl
+++ b/test/Misc/nmod_poly.jl
@@ -14,5 +14,5 @@
     d = u*f + v*g
     @test is_zero(rem(d, a))
 =#
-@end
+end
 


test/Misc/nmod_poly.jl
```diff
@@ -0,0 +1,18 @@
...
+    d = u*f + v*g
+    @test is_zero(rem(d, a))
+
+#=
+    g *= x
+    a, u, v = Hecke.gcd_sircana(f,g)
+    d = u*f + v*g
+    @test is_zero(rem(d, a))
+=#
+@end
```
thofma:
```suggestion
end
```

PR #2076, opened 2025-11-06
Prepare Oscar dispatch of is_isometric_with_isometry and assert_has_automorphisms

simonbrandhorst:


test/QuadForm/Quad/ZLattices.jl
```diff
@@ -327,10 +327,6 @@ end
     b, T = is_isometric_with_isometry(L, L2, ambient_representation = false)
     @test b
     @test T * gram_matrix(L2) * transpose(T) == gram_matrix(L)
-    b, T = is_isometric_with_isometry(L, L2, ambient_representation = true)
-    @test b
-    @test T * gram_matrix(ambient_space(L2)) * transpose(T) ==
-    gram_matrix(ambient_space(L))
```
thofma, line 333: What is up with the test?

simonbrandhorst: The feature of ambient representation does not make sense in general:
Let `L_1 \cong L_2`  but `M` not isometric to `N` (not even rationally). Form `V_1 = QQ(L_1 + M) and V_2 = QQ(L_2+N)`.  Then `V_1` is not isometric to `V_2`. 
Then `is_isometric(L_1,L_2)` returns `true`. The docstring of 
`is_isometric_with_isometry(L_1,L_2; ambient=true)` claims to return an isometry of `V_1` to `V_2`.... which does not exist. Our implementation just does some nonsense and returns some linear map which is not an isometry. 
The test however goes through because in the example `V_1 === V_2` (and `L_1 == L_2`, just with different bases I think.)

PR #2078, opened 2025-11-12
Less allocations, coordinates, and a corner case

simonbrandhorst:


src/QuadForm/Morphism.jl
```diff
@@ -2371,15 +2371,17 @@ end
...
   while i < c
     i += 1
-    if x[i] == y[i]
+    xi = mat_entry_ptr(x, 1, i)
+    yi = mat_entry_ptr(y, 1, i)
+    if cmp(xi,yi) == 0
       continue
     else
-      return x[i] < y[i]
+      return cmp(xi,yi)<0
```
fieker: wanna have fun?  `c = cmp(xi, yi); c == 0 && continue; return c;` then the comparison is only done once...

simonbrandhorst: Right, thanks!

src/QuadForm/Morphism.jl
```diff
@@ -2371,16 +2371,16 @@ end
...
   while i < c
     i += 1
-    if x[i] == y[i]
-      continue
-    else
-      return x[i] < y[i]
-    end
+    xi = mat_entry_ptr(x, 1, i)
+    yi = mat_entry_ptr(y, 1, i)
+    c = cmp(xi,yi)
```
fieker, line 2381: too fast: c is also used for the ncols. This is definitely a speed up!

PR #2081, opened 2025-11-13
Adds Igusa invariants and reduction of binary forms

JHanselman:


src/HypellCrv/Auxiliary.jl
```diff
@@ -0,0 +1,160 @@
...
+#Only consider invertible elements
+  non_zero = findall(map(is_unit, w1))
+  non_zero_w2 = findall(map(is_unit, w2))
+
+  if non_zero != non_zero_w2
+    return false 
+  end
+
+#Compute gcd of the weights
+  A = gcdx(ws[non_zero]...)
```
thofma: The version with > 2 arguments only exists in julia 1.12. @fieker do you know if we have a gcdx with arbitrary many arguments somwhere in Hecke?

test/HypellCrv/G2Invs.jl
```diff
@@ -0,0 +1,88 @@
...
+    C = hyperelliptic_curve(f1)
+
+    ig_cl_invs = @inferred igusa_clebsch_invariants(C)
+    ig_invs = @inferred igusa_invariants(C)
+    g2_invs = @inferred g2_invariants(C)
+
+    @test weighted_equality(ig_cl_invs,[ F(2), F(3), F(0), F(3) ] ,[2,4,6,10])
+    @test weighted_equality(ig_invs,[ F(4), F(1), F(2), F(3), F(3)] ,[2,4,6,8,10])
+    @test g2_invs == [ F(3), F(3), F(4) ]
+    @test weighted_equality(igusa_clebsch_from_igusa(ig_invs), ig_invs(C), [2,4,6,10])
```
thofma:
```suggestion
    @test weighted_equality(igusa_clebsch_from_igusa(ig_invs), ig_invs, [2,4,6,10])
```
?

src/HypellCrv/Auxiliary.jl
fieker: yes/no/maybe. I've answered this several times, but on machines that were not trustworthy. The easiest is to use
`hnf_with_transform`: `hnf_with_transform(matrix(ZZ, 3, 1, [2,3,4]))[2][1, :]` yields a vector of the cofactors, hnf...[1][1,1] will be the gcd

PR #2084, opened 2025-11-17
Type Stability in RieSrf

Tobias271828:
@JHanselman @simonbrandhorst 
I've started improving type stability. There's more to come in the next weeks.


src/RieSrf/PeriodMatrix.jl
```diff
@@ -88,15 +88,15 @@ function big_period_matrix(RS::RiemannSurface)
...
+  bound_temp = Vector{ArbFieldElem}()
   for path in paths
     for subpath in get_subpaths(path)
       compute_ellipse_bound(subpath, embedded_differentials, int_group_rs, RS)
       append!(bound_temp, subpath.bounds)
     end
   end
-  bound_temp = maximum(bound_temp)
-  push!(RS.bounds, bound_temp)
+  bound_temp_max = maximum(bound_temp)
```
thofma, line 98: (By the way, this is wrong if one wants something rigorous. One way to do this is do `reduce(max, bound_temp)`.)

src/RieSrf/PeriodMatrix.jl
```diff
@@ -114,7 +114,7 @@ function big_period_matrix(RS::RiemannSurface)
 
   Kxy = parent(f)
   Ky, y = polynomial_ring(base_ring(Kxy), "y")
-  m = degree(f, 2)
+  m = degree(f, 2)::Int64
```
thofma: Why is this necessary? Why is `f` not known to be a polynomial?

src/RieSrf/RiemannSurface.jl
```diff
@@ -116,7 +116,7 @@ mutable struct RiemannSurface
   #  min_pows = [1,1,1]
   #  range_pows = [4, 2, 9]
 
-  differential_form_data::Any
+  differential_form_data::Tuple{Vector{AbstractAlgebra.Generic.MPoly{AbsSimpleNumFieldElem}}, Matrix{Int64}, Vector{Int64}, Vector{Int64}}
```
thofma: There is `mpoly_type(AbsSimple...)`. And all the `Int64` must be `Int`.

src/RieSrf/PeriodMatrix.jl
simonbrandhorst: Could you please elaborate why `maximum` differs from `reduce(max, bound_temp)` ? Naively, I would expect these to be equivalent.

thofma: The call `maximum(it)` is some julia fallback and assumes that `!(x < y)` is equivalent to `x >= y`, but this is wrong for intervals.
```
julia> R = real_field();

julia> a = R("0 +/- 0.5");

julia> b = R("0.1 +/- 0.6");

julia> maximum([a, b])
[+/- 0.701]

julia> maximum([b, a])
[+/- 0.501]

julia> reduce(max, [a, b])
[+/- 0.701]
```

It is the same underlying issue as https://github.com/thofma/Hecke.jl/pull/2019#issuecomment-3349448374 and the reason most of the comparisons in the code can lead to wrong results.

src/RieSrf/PeriodMatrix.jl
Tobias271828: I don't understand. However, code_warntype tells me, m is Any if I remove the Int here.

Tobias271828: Bildschirmfoto vom 2025-11-19 14-34-24

Tobias271828: On the right I used the code with "m = degree(f, 2)::Int" and on the right just "m = degree(f, 2)". As you can see, m is either marked as Int64 or as Any. All other lines of the code where unchanged between the two runs of code_warntype.

src/RieSrf/PeriodMatrix.jl
simonbrandhorst: I am wondering, should we maybe overwrite `maximum(::Vector{ArbFieldElem})`?

Tobias271828: Thank you very much for the information!
@JHanselman Is this easy to fix or does this happens all over the place in RieSrf?

src/RieSrf/RiemannSurface.jl
Tobias271828: Maybe I am just to stupid ...
What do you mean by "mpoly_type"?

Tobias271828: @simonbrandhorst has just explained this to me.

src/RieSrf/PeriodMatrix.jl
thofma: Yes, but fixing the type of `m` in this way is just a bandaid. The real problem is that  the type of`f` cannot be inferred. And indeed, there is `f = ...` and later `f = change_base_ring(...)`. Both things should not be called `f`.

(For the types, there is no "flow" of the variables. The compiler can see them only all at once in general. So using for example `x = 1` and `x = 1.0` in the same function body makes it type instable.)

src/RieSrf/PeriodMatrix.jl
thofma: Yes, we could probably overload `maximum(::Vector{ArbFieldElement})` to fix this specific problem here.

JHanselman: There will probably be more places in the code involving inequalities, maxima and minima where we need to carefully check whether the computations are still rigorous, but overloading maximum will already be helpful.

src/RieSrf/PeriodMatrix.jl
```diff
@@ -124,8 +124,9 @@ function big_period_matrix(RS::RiemannSurface)
   # we compute the integrals during analytic continuation here
   # and that we use the Ns determined by the integration scheme.
   # If we are only interested in the monodromy we need far less.
-  s_m = SymmetricGroup(m)
+  s_m = SymmetricGroup(m)::AbstractAlgebra.Generic.SymmetricGroup{Int}
```
simonbrandhorst: This one should be type stable.  The problem was probably the `m` before.

thofma: Yes, and the `m` comes from a bad `f`, and this is what should be fixed. See https://github.com/thofma/Hecke.jl/pull/2084#discussion_r2542341186 which for some reason is "outdated".

src/RieSrf/PeriodMatrix.jl
Tobias271828: I tried to track down this problem a bit further. The problem seems to be that the function defining_polynomial(RS) seems to be type unstable. More precisely, the Riemann surface given by f = x^5 + x^4 + x^3 - x + 4 - y^2 (= the first example in the test file) is an AbstractAlgebra.Generic.MPoly{AbsSimpleNumFieldElem} whereas f in big_period_matrix, line 110 in PeriodMatrix.jl is an AbstractAlgebra.Generic.MPoly{AcbFieldElem}. The declaration of a Riemann Surface handles this by asking f to be an arbitrary MPolyRingElem. This, however, makes big_period_matrix type unstable.

@JHanselman Of which type do you want f to be? I didn't wanted to change the direct declaration of a Riemann Surface without asking you. Hence, I stopped here. Could you look over this, please?

thofma: Does it help if you change
```
  f = embed_mpoly(defining_polynomial(RS), v, max_prec)
  Cc = base_ring(f)
  I = onei(Cc)
  f = change_base_ring(Cc, f, parent = parent(f))
```
to
```
  _f = embed_mpoly(defining_polynomial(RS), v, max_prec)
  Cc = base_ring(f)
  I = onei(Cc)
  f = change_base_ring(Cc, _f, parent = parent(f))
```
?

simonbrandhorst: I would suggest in addition to give RiemannSurface a type parameter T such that 
`f::T`.

Tobias271828: > Does it help if you change
> 
> ```
>   f = embed_mpoly(defining_polynomial(RS), v, max_prec)
>   Cc = base_ring(f)
>   I = onei(Cc)
>   f = change_base_ring(Cc, f, parent = parent(f))
> ```
> 
> to
> 
> ```
>   _f = embed_mpoly(defining_polynomial(RS), v, max_prec)
>   Cc = base_ring(f)
>   I = onei(Cc)
>   f = change_base_ring(Cc, _f, parent = parent(f))
> ```
> 
> ?

I tried this yesterday without success. Moreover, when I run `@code_warntype defining_polynomial(RS)` there are red markings appearing.

thofma: Ah ok, I see. Then I agree with Simon.

JHanselman: I don't really see the point of adding a parameter T. The Riemann surfaces always expect a polynomial over a number field as everything is set up in a way that assumes the polynomial needs to be embedded. (Over Q the polynomial will automatically be converted after the changes in https://github.com/thofma/Hecke.jl/pull/2086 ) 

In that same PR I had already changed it to 
`defining_polynomial::AbstractAlgebra.Generic.MPoly{AbsSimpleNumFieldElem}`
a while ago.

A case could be made to allow for polynomials over ArbField or AcbField, but then the code would need to be rewritten to deal with that.

thofma: Yes, in view of #2086 there is nothing to be changed in the type definition. This PR here should wait for #2086. Otherwise there will be a lot of unnecessary work.

Tobias271828: Yes, this solved the issue. f, m and so on are now typestable.

src/RieSrf/PeriodMatrix.jl
Tobias271828: Indeed, m is now type stable. Hence, I've just removed the `::Ab...`.

Tobias271828:
@simonbrandhorst and me just discussed that we do not include the "maximum" for Arbs here, but open a new pull request for it.

simonbrandhorst (approved):

src/RieSrf/PeriodMatrix.jl
Tobias271828: This has been solved, see https://github.com/Nemocas/Nemo.jl/pull/2200.

PR #2086, opened 2025-11-19
Speed up and bug fixes

JHanselman:
- Sped up Riemann surface computations a bit
- Sped up differential form computation
- Several bug fixes
- It is no longer needed to choose an embedding when defining a Riemann surface
- Reverted add_row! to add_row (resp. add_column! to add_column) because it produces the wrong results for some reason.

### DIFF
--- a/src/RieSrf/RiemannSurface.jl
+++ b/src/RieSrf/RiemannSurface.jl
@@ -29,7 +29,7 @@ tau = small_period_matrix(RS)
 mutable struct RiemannSurface
   #A polynomial f(x,y) in K[x,y] for some number field K defining the Riemann
   #surface (or equivalently) a not necessarily smooth plane curve in P_2
-  defining_polynomial::MPolyRingElem
+  defining_polynomial::AbstractAlgebra.Generic.MPoly{AbsSimpleNumFieldElem}
   genus::Int
   function_field::AbstractAlgebra.Generic.FunctionField
 
@@ -154,8 +154,8 @@ mutable struct RiemannSurface
   #The constructor for a Riemann surface object
 
   function RiemannSurface(f::MPolyRingElem, v::T, prec = 100) where T<:Union{PosInf, InfPlc}
-    RS = new()
-    RS.v = v
+    RS = RiemannSurface(f, prec)
+    RS.embedding = v
     return RS
   end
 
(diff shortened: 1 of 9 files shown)


src/RieSrf/RiemannSurface.jl
```diff
@@ -1119,17 +1139,17 @@ function symplectic_reduction(K::ZZMatrix)
       v = -A[pivot, j]
       if v != 0
         #The version with ! gave different results for some reason.
-        add_row!(A, v, pivot_plus, j)
-        add_column!(A, v, pivot_plus, j)
-        add_row!(B, v, pivot_plus, j)
+        A = add_row(A, v, pivot_plus, j)
+        A = add_column(A, v, pivot_plus, j)
+        B = add_row(B, v, pivot_plus, j)
```
simonbrandhorst: Could you please investigate? If there is something fishy we should know what is wrong.

JHanselman: I think I spend about an hour trying to figure out what was going on, but I'm not quite sure. It could be something with the add_row! function, but it could also be that I used some bad programming practice somewhere. 
You can see that you get different output when you run the following code:
```
K = rationals_as_number_field()[1]
Kxy, (x,y) = polynomial_ring(K, ["x","y"])
f = x^5 + x^4 + x^3 - x + 4 - y^2
RS = RiemannSurface(f, v)
A, B, C = homology_basis(RS)
```

The version with `A = add_row(A, v, pivot_plus, j)` gives
`([[1, 6, 2, 1, 1], [1, 5, 2, 1, 1], [1, 4, 2, 1, 1], [1, 3, 2, 1, 1], [1, 2, 2, 1, 1]], [0 -1 -1 -1 -1; 1 0 -1 -1 -1; 1 1 0 -1 -1; 1 1 1 0 -1; 1 1 1 1 0], [1 -1 0 1 0; 0 1 0 0 0; 1 -1 1 0 0; 1 0 0 0 0; 1 -1 1 -1 1])`

The version with `add_row!(A, v, pivot_plus, j)` gives
`([[1, 6, 2, 1, 1], [1, 5, 2, 1, 1], [1, 4, 2, 1, 1], [1, 3, 2, 1, 1], [1, 2, 2, 1, 1]], [0 -1 -1 -1 -1; 1 0 -1 -1 -1; 1 1 0 -1 -1; 1 1 1 0 -1; 1 1 1 1 0], [3 -47 50 95 13; -4 -41 38 82 191; 1 0 1 0 -2; 0 1 -1 -2 -2; 0 0 0 0 1])`
which is incorrect.

simonbrandhorst: I cannot reproduce this behavior. Maybe you had an outdated Nemo version?
In some recent version the indices were inconsistent with the doc and abstract algebra 
We had that fixed.

JHanselman: That might have been the case. It's working for me now. (I also changed it back.) Oddly enough `A = add_row(a,v, etc.)`, stopped working for me instead.

simonbrandhorst (approved):
Looks good to me.

PR #2087, opened 2025-11-20
Fix some uses of `@inferred`

joschmitt:
Hecke equivalent of Nemocas/Nemo.jl#2179

I wasn't always certain whether somebody actually wanted to test the type stability of `==`. So please check everything.

### DIFF
--- a/src/FunField/Divisor.jl
+++ b/src/FunField/Divisor.jl
@@ -488,7 +488,7 @@ Return the degree of D.
 """
 function degree(D::Divisor)
   L = support(D)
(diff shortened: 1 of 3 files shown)


src/EllCrv/LocalData.jl
```diff
@@ -1060,7 +1060,7 @@ Return the conductor of $E$ over QQ.
 function conductor(E::EllipticCurve{QQFieldElem})
   badp = bad_primes(E)
 
-  result = 1
+  result = one(QQ)
```
thofma:
```suggestion
  result = one(ZZ)
```

PR #2094, opened 2025-11-28
sb/fix is totally isotropic

simonbrandhorst:
- **Fix is_totally_isotropic in the odd case**
- **add test**


StevellM:
It makes me wonder: is not it the same as asking whether `gram_matrix_quadratic(T)` is trivial ? Or am I missing something?

simonbrandhorst:
Yes, but with early abort.

PR #2101, opened 2025-12-09
allow frac ideals in RCG

fieker:


fieker:
On Tue, Dec 09, 2025 at 11:47:21AM -0800, Tommy Hofmann wrote:
> thofma left a comment (thofma/Hecke.jl#2101)
> 
> Is there a problem with things not being coprime to the modulus? I am thinking of $I = Q/(2)$, where $(2) = PQ$ and the modulus is $Q$. Then the integral split yields denominator a multiple of $2$?
> 
Integral split would return 1, P
But we aparently don't check the coprimeness at all

PR #2104, opened 2025-12-11
Coercion fixes for `minpoly`

SoongNoonien:
This should fix the tests for https://github.com/Nemocas/Nemo.jl/pull/2158 and https://github.com/Nemocas/Nemo.jl/pull/2150.

### DIFF
--- a/src/NumField/NfAbs/NonSimple.jl
+++ b/src/NumField/NfAbs/NonSimple.jl
@@ -525,11 +525,11 @@ function minpoly_sparse(a::AbsNonSimpleNumFieldElem)
 end
 
 function minpoly(a::AbsNonSimpleNumFieldElem)
-  return minpoly_via_trace(a)::QQPolyRingElem
+  return minpoly(Globals.Qx, a)
 end
 
 function minpoly(Qx::QQPolyRing, a::AbsNonSimpleNumFieldElem)
-  return Qx(minpoly(a))
+  return minpoly_via_trace(Qx, a)::QQPolyRingElem
 end
 
 function minpoly(Rx::ZZPolyRing, a::AbsNonSimpleNumFieldElem)
@@ -537,7 +537,7 @@ function minpoly(Rx::ZZPolyRing, a::AbsNonSimpleNumFieldElem)
   if !isone(denominator(f))
     error("element is not integral")
   end
-  return Rx(denominator(f)*f)
+  return Rx(f)
 end
 
 function minpoly(a::AbsNonSimpleNumFieldElem, R::ZZRing)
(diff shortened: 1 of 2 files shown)


src/NumField/NfAbs/NonSimple.jl
```diff
@@ -525,19 +525,19 @@ function minpoly_sparse(a::AbsNonSimpleNumFieldElem)
 end
 
 function minpoly(a::AbsNonSimpleNumFieldElem)
-  return minpoly_via_trace(a)::QQPolyRingElem
+  return minpoly(Globals.Qx, a)
 end
 
 function minpoly(Qx::QQPolyRing, a::AbsNonSimpleNumFieldElem)
-  return Qx(minpoly(a))
+  return minpoly_via_trace(Qx, a)::QQPolyRingElem
```
fingolfin, line 532: I don't mind the type assertion, but is it really needed? I.e., is `minpoly_via_trace` type unstable? Then someone should also look into that eventually.

src/NumField/NfAbs/NonSimple.jl
```diff
@@ -1058,7 +1058,7 @@ function minpoly_via_trace(a::AbsNonSimpleNumFieldElem)
       i += 1
     end
     q = QQFieldElem(1, div(d, i))
-    f = power_sums_to_polynomial([x*q for x = l])
+    f = power_sums_to_polynomial([x*q for x = l]; parent = Qx)
     if iszero(subst(f, a))  #TODO: to checks first...
       return f::QQPolyRingElem
```
fingolfin, line 1063: Ah so there is a type assertion here, too -- so the outer one should not be necessary?

But if we need that, it means `power_sums_to_polynomial` is type unstable, and we really should do
```suggestion
    f = power_sums_to_polynomial([x*q for x = l]; parent = Qx)::QQPolyRingElem
    if iszero(subst(f, a))  #TODO: to checks first...
      return f
```
so that we also benefit in the `subst` call.

src/NumFieldOrd/NfOrd/Elem.jl
```diff
@@ -193,7 +193,7 @@ The minimal polynomial of $a$. The parent of the polynomial can be supplied
 as a first argument.
 """
 function minpoly(R::ZZPolyRing, a::AbsNumFieldOrderElem)
-  return R(minpoly(Globals.Qx, elem_in_nf(a)))
+  return minpoly(R, elem_in_nf(a))
```
fingolfin, line 196: Is this working right, though? Before it invoked `minpoly` recursively with a `QQPolyRing`, now with a `ZZPolyRing` -- is the latter really supported?

Perhaps a bunch of methods need to be extended to allow for `Union{QQPolyRing,ZZPolyRing}`

src/NumField/NfAbs/NonSimple.jl
SoongNoonien: I don't think that it is needed but it was there before so I kept it.

src/NumFieldOrd/NfOrd/Elem.jl
SoongNoonien: > is the latter really supported?

Yes, it should be in line 535.

PR #2105, opened 2025-12-12
Upper Bounds for Arbs

Tobias271828:
In RieSrf, a maximum of Arbs is often taken which is not exact. Here, we are providing a function to solve this issue. A simple maximum function is implemented which returns a maximum together with a bool, stating whether the maximum is exact.

@simonbrandhorst and me also thought of a fast upper bound function which returns an exact upper bound. This should be a number which is close to the supremum, but fast to determine exactly so that it should be better for most computations in RieSrf.

### DIFF
--- a/src/RieSrf/UpperBound.jl
+++ b/src/RieSrf/UpperBound.jl
@@ -1,0 +1,26 @@
+function arb_maximum(vectofarbs::Vector{RealFieldElem})::Tuple{RealFieldElem,Bool}
+  if length(vectofarbs) <= 1
+    if length(vectofarbs) == 0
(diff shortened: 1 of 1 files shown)


Tobias271828:
Thank you @thofma ! Yes, indeed, this sounds much better.

My initial idea was to catch cases like comparing 0 +- 1, 0.4 +- 0.3 and 3 +- 0.5. Clearly, the third one is the maximum. However, the first two aren't comparable. However, usally we just need an upper bound which should ideally be the maximum. Hence, I think your idea is much more sensefully than mine ...

PR #2107, opened 2025-12-13
Less cache of QuadSpace and HermSpace in library code

simonbrandhorst:
- **Cache less QuadSpace**
- **less cache**

### DIFF
--- a/src/QuadForm/Herm/GenusRep.jl
+++ b/src/QuadForm/Herm/GenusRep.jl
@@ -676,9 +676,9 @@ function spinor_genera_in_genus(
   )
   @req rank(L) >= 2 "Lattice must have rank >= 2"
   s = denominator(scale(L))
-  L = rescale(L, s; cached=false)
   R = base_ring(L)
-  gens, def, P0 = genus_generators(L)
+  _L = rescale(L, s; cached=false)
+  gens, def, P0 = genus_generators(_L)
   a = involution(L)
   res = typeof(L)[ L ]
   if max == 1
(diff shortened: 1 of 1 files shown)


src/QuadForm/Spaces.jl
```diff
@@ -684,12 +684,12 @@ end
 #
 ################################################################################
 
-function _biproduct(x::Vector{T}) where T <: AbstractSpace
+function _biproduct(x::Vector{T}; cached) where T <: AbstractSpace
```
StevellM:
```suggestion
function _biproduct(x::Vector{T}; cached::Bool=true) where T <: AbstractSpace
```

src/QuadForm/Herm/Genus.jl
```diff
@@ -1706,14 +1710,14 @@ function representative(G::HermGenus)
...
-    L = rescale(L, 1//s)
+    L = rescale(L, 1//s; cached=false)
     G.representative = L
     L.scale = scale(G)
     return L
   end
   P = _non_norm_primes(G.LGS)
   E = base_field(G)
-  V = hermitian_space(E, _hermitian_form_with_invariants(base_field(G), rank(G), P, G.signatures))
+  V = hermitian_space(E, _hermitian_form_with_invariants(base_field(G), rank(G), P, G.signatures;all_non_split=true))
```
StevellM: I do not know if the problem comes from here, but are you sure about that ? By default `_non_norm_primes` does not ignore the split primes so `P` might contain split primes.. or is there a reason it would not ?

simonbrandhorst: `P` constists of non-norm primes but over a split extension everything is a norm.

simonbrandhorst:
@thofma  The remaining failure seems unrelated. Please merge.

PR #2108, opened 2025-12-14
Make ideal method stricter

fingolfin:
They redirect to _ideal_pid so the argument lists should match.

Also remove some unreachable code.

Motivated by https://github.com/Nemocas/AbstractAlgebra.jl/pull/2108 which runs into an ambiguity error for `ideal(::QQPolyRing, ::Vector{Any})`.

### DIFF
--- a/src/Misc/PIDIdeal.jl
+++ b/src/Misc/PIDIdeal.jl
@@ -54,7 +54,7 @@ function ideal(R::PolyRing{<:FieldElem}, x::RingElement, y::RingElement...)
   return _ideal_pid(R, x, y...)
 end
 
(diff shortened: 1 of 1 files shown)


fingolfin:
Turns out it relied on the changed code in an unexpected way:
```
  Test threw exception
  Expression: I == ideal(Kx, [x, x ^ 2, 0])
  MethodError: no method matching (::QQPolyRing)(::Vector{Any})
```
Ah well. I could fix it, but it might be easier to wait for that breaking AA release with the ideal changes in (0.48.0 I assume), and then I can update this PR to depend on AA 0.48.0

PR #2123, opened 2026-01-03
feat: add some functionality for multiplicative groups

thofma:


src/EllCrv/Misc.jl
fieker, line 133: why? In other placs we allow e.g. `valuation(8, 4)` as the largest k s.th. `p^k|a`
Also: testing irreducibility is too expensive

src/FunField/Divisor.jl
fieker, line 598: we really need to change/ fix divisors to store the inverse of the ideals!!! I do not know where the ideals themselves are useful
maybe in a seperate PR?

thofma:
sorry, I messed up my commits here

PR #2128, opened 2026-01-07
Address several JET reports

fingolfin:


src/NumField/NfAbs/PolyFact.jl
```diff
@@ -703,13 +703,12 @@ function van_hoeij(f::PolyRingElem{AbsSimpleNumFieldElem}, P::AbsNumFieldOrderId
 
     while length(have) > length(used)
       local m
-      m_empty = true
```
thofma, line 706: Doesn't this make the code type instabil? What is the problem with the code at the moment? Looks fine to me.

thofma (commented):
I guess in most cases JET complains about variables not being defined, since they are defined only conditionally? We can fix this now, but unintentionally such things will probably be added again.

src/NumField/NfAbs/PolyFact.jl
fingolfin: This leads to two-way type split, which Julia is guaranteed to optimize (after all, it is necessary for efficient `for` loops).

The code as it is is correct, but (a) may lead to boxing of `m_empty`, and (b) prevents static analysis of the code, as it can't track the relation between the value of `m_empty` and whether `m` is define or not. Ah, and it allows us to get rid of the `local m` as well

src/NumField/NfAbs/PolyFact.jl
```diff
@@ -703,13 +703,12 @@ function van_hoeij(f::PolyRingElem{AbsSimpleNumFieldElem}, P::AbsNumFieldOrderId
 
     while length(have) > length(used)
       local m
```
fingolfin, line 705:
```suggestion
```

src/NumFieldOrd/NfOrd/NfOrd.jl
```diff
@@ -668,8 +668,11 @@ function _norm_change_const(v::Vector{AbsSimpleNumFieldElem})
         break
       end
     end
+    if !fl1
+      fl1 = !(r[1] > 0)
+    end
   end
-  if fl1 || !(r[1] > 0)
```
fingolfin, line 672: Here the problem is that `r` is only sometimes defined in the code before, and a static analyzer can't prove that if `fl1` is `false` (i.e. `r` is accessed) then `r` is defined.

PR #2129, opened 2026-01-09
Remove dead, broken code

fingolfin:
JET flagged use of undefined variables in this code, and I then manually verified these functions are not called from anywhere.

Of course you may still wish to preserve (and fix) some of these. Then we can adjust PR.


src/HypellCrv/Auxiliary.jl
```diff
@@ -148,26 +148,3 @@ function weighted_reduction(w1::Vector{QQFieldElem}, ws::Vector{Int})
...
     return w1_min
 end
-
-# Given a linear equation of the form f(x, y) = a * x + b * y = 0,
-# find the minimal x_0, y_0 such that f(x_0, y_0) = 0.
-function minimize_linear_equation(f::MPolyRingElem{QQFieldElem})
-  R = parent(f)
-  x, y = gens(R)
-
-  a_num = numerator(coeff(f, u))
```
fingolfin, line 158: there is no `u`

src/NumField/NfAbs/MPolyAbsFact.jl
```diff
@@ -392,12 +392,6 @@ mutable struct RootCtxSingle{T}
   R::T  # the root
   o::T  # inv(f'(R)) for the double lifting.
 
-  function RootCtxSingle(f::PolyRingElem{S}, K::fqPolyRepField) where {S <: SeriesElem}
-    #not used I think
```
fingolfin, line 396: correct :-)

src/NumField/NfAbs/MPolyAbsFact.jl
```diff
@@ -392,12 +392,6 @@ mutable struct RootCtxSingle{T}
...
-    RR,  = power_series_ring(K, max_precision(R), var(R), cached = false) #can't get the model
```
fingolfin, line 397: there is no `R`

src/NumFieldOrd/NfOrd/Clgp/Main_enum.jl
```diff
@@ -1,25 +1,3 @@
...
-
-function class_group_random_ideal_relation(clg::ClassGrpCtx, r::Int,
-                                           I::AbsNumFieldOrderIdeal{AbsSimpleNumField, AbsSimpleNumFieldElem} = rand(clg.FB.ideals))
-  s = 1
-  if r < 2
-    r = 2
-  end
-  for i = 1:r
-    I = I*rand(clg.FB.ideals)
-    I, g = reduce_ideal_class(I)
```
fingolfin, line 15: there is no `reduce_ideal_class`

src/NumFieldOrd/NfOrd/Unit/UnitGrpCtx.jl
```diff
@@ -190,35 +190,6 @@ function _conj_log_mat_cutoff_inv(x::UnitGrpCtx, p::Int)
   end
 end
 
-function _isindependent(u::UnitGrpCtx{T}, y::FacElem{T}) where T
-  K = _base_ring(x[1])
```
fingolfin, line 194: there is no `x`

src/QuadForm/Morphism.jl
```diff
@@ -1139,63 +1111,6 @@ function _operate(point, A, V, tmp)
...
-function _find_point(w::Vector{Int}, V)
-  positive = false
-  for k in 1:length(w)
-    if !iszero(w[k])
-      positive = w[k] > 0
-      break
-    end
-  end
-  if positive
-    if sorted
```
fingolfin, line 1151: `sorted` is not defined

src/QuadForm/Morphism.jl
```diff
@@ -1139,63 +1111,6 @@ function _operate(point, A, V, tmp)
...
-function _find_point(w::ZZMatrix, V)
-  positive = false
-  for k in 1:length(w)
-    if !iszero(w[1, k])
-      positive = w[1, k] > 0
-      break
-    end
-  end
-  if positive
-    if sorted
```
fingolfin, line 1180: `sorted` is not defined

src/helper.jl
```diff
@@ -1,105 +0,0 @@
...
-  end
-end
-
-function contains_any(T)
-  if typeof(T) == TypeVar
-    #return contains_any(T.ub)
-    return false
-  elseif typeof(T) == Union
-    return false
-  elseif isleaftype(T)
```
fingolfin, line 93: There is no `isleaftype` (I guess this comes form before Julia 1.0?)

src/QuadForm/Morphism.jl
```diff
@@ -107,34 +107,6 @@ function has_point(w, V::VectorList)
   end
 end
 
-function _find_point(w::ZZMatrix, V::VectorList{ZZMatrix, T}) where T
```
fingolfin, line 110: There is no code calling `_find_point`, and the other two methods reference things that don't exist. It seemed logical to also remove this method.

src/HypellCrv/Auxiliary.jl
JHanselman: I don't think this necessarily needs to go. It needs to be fixed sure, (My guess is that I decided to rename variables at some point, but missed some) but the function will be used in an upcoming PR for reconstructing genus 2 curves from their Igusa invariants. If it does get removed I'll readd a fixed version of the function later in any case.

fingolfin: I am happy to remove removals in this PR, such as this one -- it simply was the easiest way to draw attention to definitely broken code. If you plan to take care of it by other means, that's great, and I'll shortly adjust this PR.

fingolfin: done

PR #2130, opened 2026-01-09
Fix a bunch of "foo is not defined" JET warnings

fingolfin:
In most cases this also means that the code I am editing here is not being tested. It might be a good idea to add tests, but that's not within my abilities.

Also, someone please carefully check all of these! I only made edits when it seemed fairly clear from context what was intended, but of course I may have messed up. And this code is not being tested!

### DIFF
--- a/src/Map/NumField.jl
+++ b/src/Map/NumField.jl
@@ -648,8 +648,6 @@ _convert_map_data(g::NumFieldHom, L) = __convert_map_data(g.image_data, L)
 
 __convert_map_data(d::MapDataFromAnticNumberField, L) = MapDataFromAnticNumberField{elem_type(L)}(d.isid ? true : L(d.prim_image))
 
-__convert_map_data(d::MapDataFromNfRel, L) = MapDataFromNfRel{elem_type(L)}(L(d.prim_image), d.isid ? true : __convert_map_data(d.base_field_map_data), d.isid)
-
 ################################################################################
 #
 #  Helper functions to pass through inverse data
--- a/src/ModAlgAss/Lattices/Basics.jl
+++ b/src/ModAlgAss/Lattices/Basics.jl
@@ -260,7 +260,7 @@ end
 
 function index(L::T, M::T) where {T <: ModAlgAssLat}
   t = basis_matrix(L) * basis_matrix_inverse(M)
-  @req !isone(denominator(t)) "First lattice not contained in second lattice"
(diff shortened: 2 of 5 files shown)


src/AlgAss/Conversions.jl
```diff
@@ -733,7 +733,7 @@ function StructureConstantAlgebra(I::Union{ RelNumFieldOrderIdeal{T, S}, AlgAssR
 
       for k in reducers
         d = -coeffs[k]//new_bmatJinI[k, k]
-        coeffs_c = coeffs_c + d .* rows_new_bmatJinI[k]
```
fingolfin, line 736: `coeffs_c` is not defined before this line

src/Map/NumField.jl
```diff
@@ -648,7 +648,7 @@ _convert_map_data(g::NumFieldHom, L) = __convert_map_data(g.image_data, L)
 
 __convert_map_data(d::MapDataFromAnticNumberField, L) = MapDataFromAnticNumberField{elem_type(L)}(d.isid ? true : L(d.prim_image))
 
-__convert_map_data(d::MapDataFromNfRel, L) = MapDataFromNfReld{elem_type(L)}(L(d.prim_image), d.isid ? true : __convert_map_data(d.base_field_map_data), d.isid)
+__convert_map_data(d::MapDataFromNfRel, L) = MapDataFromNfRel{elem_type(L)}(L(d.prim_image), d.isid ? true : __convert_map_data(d.base_field_map_data), d.isid)
```
fingolfin: Look like this might a typo, but my proposed change is a guess in the end

src/Misc/nmod_poly.jl
```diff
@@ -804,7 +804,7 @@ function gcd_sircana(f::PolyRingElem{T}, g::PolyRingElem{T}) where T <: ResElem{
   # cannot ignore the contents as it can contribute...
   @assert isone(content(f))
   @assert isone(content(g))
-
+  local qf, qg
```
fingolfin, line 807: These are defined in the loop but then used after it ends

src/AlgAss/Conversions.jl
thofma: @joschmitt this is correct right? (a bit strange that this is not hit in the tests)

joschmitt: Yes, this is correct.

PR #2133, opened 2026-01-12
impossible subsets, like (3, 4), would be infinite recursive

fieker:
came from
using Oscar
Qx, x = QQ["x"]

  a=[314721,x^4-x^3-46*x^2-47*x+2209]
  bnd=ZZ(10)^11
  k, _ = number_field(a[2]; cached = false, check = false)
  L = abelian_extensions(k, [2], bnd, signatures = [(2,3)])

Warning: detected a stack overflow; program state may be corrupted, so further execution might be unreliable.
ERROR: LoadError: StackOverflowError:
Stacktrace:
 [1] subsets(n::Int64, k::Int64) (repeats 19389 times)
   @ Hecke ~/.julia/packages/Hecke/Xk3rO/src/NumFieldOrd/NfOrd/LLL.jl:449
 [2] subsets(v::Vector{Int64}, k::Int64)
   @ Hecke ~/.julia/packages/Hecke/Xk3rO/src/NumFieldOrd/NfOrd/LLL.jl:464
 [3] abelian_extensions(K::AbsSimpleNumField, gtype::Vector{Int64},

(Malle)

### DIFF
--- a/test/NumFieldOrd/NumFieldOrd.jl
+++ b/test/NumFieldOrd/NumFieldOrd.jl
@@ -194,7 +194,9 @@ end
   # L = abelian_extensions(k, [2], bnd, signatures = [(2,3)])
   #will encounter an infinite recursion in subsets (Malle)
 
-  @test_throws AssertionError subsets(3, 4)
+  @test_throws ArgumentError subsets(3, -1)
+  @test_throws ArgumentError subsets(-3, 1)
+  @test_throws ArgumentError subsets(-3, -1)
   @test length(subsets(3, 3)) == 1
   @test length(subsets(3, 2)) == 3
   @test length(subsets(3, 4)) == 0


src/NumFieldOrd/NfOrd/LLL.jl
```diff
@@ -434,6 +434,7 @@ end
 Return a vector of all ordered `k`-element subsets of `1..n`.
 """
 function subsets(n::Int, k::Int)
+  @assert n >= k >= 0
```
lgoettgens: The following line already handles the case `n < k` by returning something sensible

src/NumFieldOrd/NfOrd/LLL.jl
```diff
@@ -461,6 +462,7 @@ end
 Return a vector of all ordered `k`-element sub-vectors of `v`.
 """
 function subsets(v::Vector{T}, k::Int) where T
+  @assert k <= length(v)
```
lgoettgens: since this calls the `subset(::Int, ::Int)` method in the following line, this assertion is not needed.

src/NumFieldOrd/NfOrd/LLL.jl
fieker: (3, -1) will produce an infinite recursion without some check.
The more interesting question would be if illegal input should get an empty result or an error

lgoettgens: From the current docstring, I would say an empty result would fit better. But for negative inputs, I have no strong preference.
However, for `n < k` (both `>= 0`), the current behaviour should be preserved

fingolfin: So at the very least:
```suggestion
  @assert n >= 0
  @assert k >= 0
```
But actually an "assert" is not really what we want, this should produce a user visible error, always (while `@assert` could be elided). Moreover, if we are willing to accept $k$ negative as long as $n$ is non-negative, the second check is redundant with `if n < k` ... so
```suggestion
  @req n >= 0 "n must be non-negative"
```
or
```suggestion
  @req (n >= 0 && k >= 0) "inputs must be non-negative"
```

src/NumFieldOrd/NfOrd/LLL.jl
fingolfin:
```suggestion
```

test/NumFieldOrd/NumFieldOrd.jl
```diff
@@ -185,3 +185,16 @@ let # extend_easy for bad polynomials
...
+  #without 
+  # a=[314721,x^4-x^3-46*x^2-47*x+2209]
+  # bnd=ZZ(10)^11
+  # k, _ = number_field(a[2]; cached = false, check = false)
+  # L = abelian_extensions(k, [2], bnd, signatures = [(2,3)])
+  #will encounter an infinite recursion in subsets (Malle)
+
+  @test_throws AssertionError subsets(3, 4)
+  @test length(subsets(3, 3)) == 1
+  @test length(subsets(3, 2)) == 3
```
thofma, line 201:
```suggestion
  @test length(subsets(3, 2)) == 3
  @test length(subsets(3, 4)) == 0
```

test/NumFieldOrd/NumFieldOrd.jl
```diff
@@ -185,3 +185,17 @@ let # extend_easy for bad polynomials
...
+
+@testset "Misc-subsets" begin
+  #without 
+  # a=[314721,x^4-x^3-46*x^2-47*x+2209]
+  # bnd=ZZ(10)^11
+  # k, _ = number_field(a[2]; cached = false, check = false)
+  # L = abelian_extensions(k, [2], bnd, signatures = [(2,3)])
+  #will encounter an infinite recursion in subsets (Malle)
+
+  @test_throws AssertionError subsets(3, 4)
```
fingolfin:
```suggestion
  @test_throws ArgumentError subsets(3, -1)
  @test_throws ArgumentError subsets(-3, 1)
  @test_throws ArgumentError subsets(-3, -1)
```

PR #2134, opened 2026-01-12
Remove trailing spaces in various files

fingolfin:
The `.editorconfig` forces remove of trailing whitespace. Which makes it annoying when editing a file where these show up, if one uses an editor following `.editorconfig`.

Alternative:
```patch
diff --git a/.editorconfig b/.editorconfig
index 4ce738897..305a4fc63 100644
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,7 +9,7 @@ charset = utf-8
 end_of_line = lf
 indent_style = space
 insert_final_newline = true
-trim_trailing_whitespace = true
+trim_trailing_whitespace = false
 
 [*.jl]
 indent_size = 2
```


thofma:
I vaguely remember some git magic to hide this in git blame etc?

PR #2144, opened 2026-01-16
Add CI action to use oscar-system/changelog-script

aaruni96:
This PR adds the use of oscar-system/changelog-script to Hecke.

~The original action needs some patching (for example, https://github.com/oscar-system/update-release-notes/issues/4 needs to be fixed), after which it will be tagged in a release. Until then, the action can still be used manually, but this PR should not be merged.~

The result from running the [action](https://github.com/thofma/Hecke.jl/actions/runs/21069779558) is available at https://github.com/thofma/Hecke.jl/pull/2145

[skip ci]

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -27,7 +27,7 @@ jobs:
 
     # Run the update script
     - name: Run python script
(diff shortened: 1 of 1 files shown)


thofma:
Great, thanks for your help!

PR #2147, opened 2026-01-18
Added in-place operations for elliptic curve points

eoinmackall:
Added the following in-place operations for elliptic curve points:
- `add!(R::EllipticCurvePoint{T}, P::EllipticCurvePoint{T}, Q::EllipticCurvePoint{T}) where T`
- `inv!(R::EllipticCurvePoint{T}, P::EllipticCurvePoint{T}) where T`
- `copy_coords!(R::EllipticCurvePoint{T}, P::EllipticCurvePoint{T}) where T`
- `sub!(R::EllipticCurvePoint{T}, P::EllipticCurvePoint{T}, Q::EllipticCurvePoint{T}) where T`
- `mul!(R::EllipticCurvePoint, n::S, P::EllipticCurvePoint, C::EllipticCurvePoint = infinity(parent(P))) where S<:Union{Integer, ZZRingElem}`

I'm not sure why the documentation for some of the original operations is commented out (e.g. `+`, `*`, `-`, ...). I imagine possibly because it can clutter the Julia documentation? If so, maybe some of these functions should have their documentation commented out as well.

One comment: all of the functions are implemented using direct field access for the `EllipticCurvePoint` struct. I'm not sure how frowned upon this is. Also, this introduces the behavior that the group identity can have nontrivial `coordx` and `coordy` fields.

I also deleted an unreachable `else` block in the original `+(P::EllipticCurvePoint{T}, Q::EllipticCurvePoint{T}) where T`.

TODO:
- add tests for the new functions

Comments are very welcome.

### DIFF
--- a/src/EllCrv/EllCrv.jl
+++ b/src/EllCrv/EllCrv.jl
@@ -659,7 +659,7 @@ end
 @doc raw"""
     infinity(E::EllipticCurve) -> EllipticCurvePoint
 
-Return the point at infinity with project coordinates $[0 : 1 : 0]$.
+Return the point at infinity with projective coordinates $(0 : 1 : 0)$.
 """
 function infinity(E::EllipticCurve{T}) where T
   infi = EllipticCurvePoint{T}(E)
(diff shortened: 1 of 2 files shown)


src/EllCrv/EllCrv.jl
```diff
@@ -1031,6 +1027,267 @@ function *(n::S, P::EllipticCurvePoint) where S<:Union{Integer, ZZRingElem}
...
+julia> P = E([3, 8]);
+
+julia> inv!(P,P)
+(3 : -8 : 1)
+
+julia> P
+(3 : -8 : 1)
+```
+"""
+function inv!(R::EllipticCurvePoint{T}, P::EllipticCurvePoint{T}) where T
```
thofma: Maybe this should be `neg!`?

src/EllCrv/EllCrv.jl
```diff
@@ -1031,6 +1027,267 @@ function *(n::S, P::EllipticCurvePoint) where S<:Union{Integer, ZZRingElem}
...
+
+  return R
+end
+
+@doc raw"""
+    copy_coords!(R::EllipticCurvePoint, P::EllipticCurvePoint)
+
+Copies the point $P$ onto the point $R$, mutating $R$. Points must have the same parent curve.
+"""
+function copy_coords!(R::EllipticCurvePoint{T}, P::EllipticCurvePoint{T}) where T
```
thofma:
```suggestion
function set!(R::EllipticCurvePoint{T}, P::EllipticCurvePoint{T}) where T
```
This is how we usually call this internally.

src/EllCrv/EllCrv.jl
```diff
@@ -1031,6 +1027,267 @@ function *(n::S, P::EllipticCurvePoint) where S<:Union{Integer, ZZRingElem}
...
+  parent(R) != parent(P) && error("Can only write to a point on the same curve")
+
+  # Is P = infinity or Q = infinity?
+  if P.is_infinite
+    if R !== Q
+      R.is_infinite = Q.is_infinite
+      if !Q.is_infinite
+        R.coordx = Q[1]
+        R.coordy = Q[2]
+      end
```
thofma:
```suggestion
      set!(R, Q)
```
I think this could be used here? Probably also other places

src/EllCrv/EllCrv.jl
```diff
@@ -1031,6 +1027,267 @@ function *(n::S, P::EllipticCurvePoint) where S<:Union{Integer, ZZRingElem}
...
+    R.is_infinite = true
+    return R
+  elseif R===P
+    inv!(P,P)
+    return inv!(P,add!(P,P,Q))
+  else
+    copy_coords!(R,Q)
+    inv!(R,R)
+    return add!(R,P,R)
+  end
```
thofma: Spacing a bit odd here.

thofma:
> > 1. The docstrings should ideally be removed, since the clutter up the REPL help mode.
> 
> That makes sense -- is it alright to comment them out, rather than remove them?

Sure.

> I meant that, as far as I could tell, any instance of the identity point `P` that would be created from the functions in this file, (e.g. calling `P = infinity(E)` for an elliptic curve E), would have `P.coordx` and `P.coordy` unset. Allowing in-place operations, though, allows for something like:
> 
> * Q is a `finite` point with `Q.coordx` set and `Q.coordy` set,
> * P is an `infinite` point, say `P=infinity(E)`
> * we do `sub!(Q,Q,Q)`
> * now `Q == P`, but `Q.coordx` and `Q.coordy` are set to some specific values whereas `P.coordx` and `P.coordy` are unset.
> 
> It feels like this is a little bit subtle (to me). I guess it should be fine, but I it's an added complication of doing affine arithmetic of points (i.e. we always have to check `Q.is_infinite` as a case before actually using the values).

I would say it is fine.

Thanks for adding this stuff here!

eoinmackall:
> There is one more TODO in your original post, but I think this was addressed, right?

That's right (I added tests for the in-place operations).
I think it should be ready to merge now.

thofma:
Great, thanks!

PR #2149, opened 2026-01-19
RieSrf Nicer printing

Tobias271828:
Added a `Base.show(...)` in order to make the output of code like
`using Hecke.RiemannSurfaces
QQxy, (x,y) = polynomial_ring(QQ, ["x","y"])
f = x^5 + x^4 + x^3 - x + 4 - y^2
RS = RiemannSurface(f)`
being hopefully more reasonable to users.

### DIFF
--- a/test/RieSrf/RiemannSurface.jl
+++ b/test/RieSrf/RiemannSurface.jl
@@ -16,6 +16,7 @@
     R("0.5515666804208422259") + R("0.7989131245853492630")*im]])
 
   @test contains(matrix(R2,tau), test_tau)
+  @test sprint(show, "text/plain", RS) isa String
 
   # an Elliptic curve
   f = x^3 + 1 - y^2


src/RieSrf/RiemannSurface.jl
```diff
@@ -301,6 +301,8 @@ mutable struct RiemannSurface
   end
 end
 
+#Nicer printing
+Base.show(io::IO, rs::RiemannSurface) = print(io, "Riemann Surface of genus $(rs.genus) defined by $(rs.defining_polynomial) = 0.")
```
lgoettgens:
```suggestion
Base.show(io::IO, rs::RiemannSurface) = print(io, "Riemann surface of genus $(rs.genus) defined by $(rs.defining_polynomial) = 0")
```

src/RieSrf/RiemannSurface.jl
```diff
@@ -301,6 +301,15 @@ mutable struct RiemannSurface
   end
 end
 
+#Nicer printing
+import AbstractAlgebra.is_terse
```
thofma: Please move the important statement to `src/RieSrf/RiemannSurface.jl`.

JHanselman:
> > Thanks. Is there someone who should approve this? I have no opinion on the printing of these objects.
> 
> I don't know - maybe @JHanselman ? He told me that I could make the printing of Riemann surfaces look nicer.

Yeah, I think it is fine like this. I don't think any additional information needs to be printed.

PR #2151, opened 2026-01-19
Fix issues revealed by OSCAR test suite

fingolfin:
I am not sure if this is all.

### DIFF
--- a/docs/src/manual/orders/orders.md
+++ b/docs/src/manual/orders/orders.md
@@ -41,10 +41,10 @@ any_order(K::AbsSimpleNumField)
 ```jldoctest
 julia> Qx, x = polynomial_ring(QQ, :x);
 
-julia> K, a = number_field(x^2 - 2, :a);
+julia> K, a = number_field(x^2 - 2, :a; cached = false);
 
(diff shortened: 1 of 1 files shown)


src/Misc/RatRecon.jl
```diff
@@ -376,11 +378,10 @@ function _modpResults(f, p::ZZRingElem, M::Int)
    Rc = f.parent
    l1 = fpPolyRingElem[]; l3 = ZZRingElem[]
    Np = listprimes([f], p, M)
-   Zx, Y = polynomial_ring(ZZ, "Y", cached=false)
    for j in 1:length(Np)
      RNp = Native.GF(Int(Np[j]), cached=false)
      Rp, t = polynomial_ring(RNp, "t", cached=false)
-     fp = Rp(f)
+     fp = change_base_ring(RNp, f; parent = Rp)
```
fieker, line 384: by the looks of it, this is much slower than the old function. The coercion maps directly to a flint-c-function, change_base_ring is pure julia

thofma: We have changed it, so that this calls flint directly.

PR #2159, opened 2026-01-31
closest vectors

simonbrandhorst:


src/QuadForm/CloseVectors.jl
```diff
@@ -209,6 +209,34 @@ function Base.iterate(C::LatCloseEnumCtx{X, elem_type}, start = nothing) where {
...
+        done = false
+        break
+      elseif l == cur_min
+        push!(cur_vec, v)
+      end
+    end
+  end
+
+  min = cur_min
+  return min, cur_vec
```
StevellM: I do not see the point of that, and `min` is already globally defined... Why not returning `cur_min, cur_vec` ? Anyway, this should not block merging, I am just confused by this. Similarly above with `cur_min = upperbound` where one could have directly defined `cur_min` as `upperbound` is defined.

src/QuadForm/CloseVectors.jl
```diff
@@ -209,6 +209,34 @@ function Base.iterate(C::LatCloseEnumCtx{X, elem_type}, start = nothing) where {
 end
 
 
+function _closest_vectors(L::ZZLat, v::Vector, elem_type::Type{S} = ZZRingElem; kw...) where S
+  lowerbound = 0
+  if gram_matrix(L)[1,1]<0
+    L = rescale(L,-1)
+  end
+  upperbound = abs(sum(diagonal(gram_matrix(L))))  # too big, better?
```
StevellM, line 217: What would happen is this is not big enough ? Can this happen actually ?

simonbrandhorst: it was just the most stupid upper bound which came to mind.

src/QuadForm/CloseVectors.jl
simonbrandhorst:
```suggestion
  return cur_min, cur_vec
```

PR #2160, opened 2026-02-01
Implemented point counting for elliptic curves in characteristic 3.

alexey-orlov-math:
Implemented point counting for elliptic curves in characteristic 3
- supersingular curves are handled through explicit isomorphism class computation (see http://arxiv.org/abs/2601.21756)
- the case of j in F_9 is handled via "Subfield Curve" approach; for j in F_3 we have explicit formulae, otherwise we use the embedding of F_9 into F_q to get the isomorphic curve defined over F_9
- otherwise we use AGM method from "An AGM-Type Elliptic Curve Point Counting Algorithm in Characteristic Three" by Gustavsen and Ranestad.

Also includes some style changes in point counting in characteristic 2, some extra tests, and some refactoring

### DIFF
--- a/src/EllCrv/Finite.jl
+++ b/src/EllCrv/Finite.jl
@@ -809,17 +809,19 @@ julia> order(E)
   p == 0 && error("Characteristic must be nonzero")
 
   if p == 2
-    j = j_invariant(E)
-    if j == 0
+    if j_invariant(E) == 0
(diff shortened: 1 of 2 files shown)


thofma:
Very nice, thanks! Feel free to restructure the source code and split the files where it makes sense.

PR #2168, opened 2026-02-04
Some improvments to AGM algorithms for point counting

alexey-orlov-math:
* Updated (qadic) Newton lift to handle polynomials in Z_q[x] directly too
  - renamed the existing implementation to `_newton_lift` and made it use q-adic $f$ and $f'$
  - added comments about implicit assumptions of the implementation (most notably the condition on the valuation of $f'(r)$)
  - `newton_lift` now accepts `f::Union{PolyRingElem{ZZRingElem}, PolyRingElem{QadicFieldElem}}` as input

* Tweaked `roots(Q::QadicField, f::ZZPolyRingElem; max_roots)` 
  - the docstring is now more explicit about simple roots
  - tweaked implementation to *ignore* non-simple roots (since in this case the answer doesn't make any sense and is completely wrong)
  - it now uses `_newton_lift` (the derivative is computed once, and both $f$ and $f'$ are converted to q-adic polynomials once)
  - added some extra tests for both `roots` and `newton_lift` to have as an extra contract for the requirements 

* Tweaked AGM implementation in characteristic 3 to use `newton_lift` with q-adic polynomial
For tests, I have tweaked the implementation to disable Conway polynomial check and to use generator of the residue field as a_6 coefficient (this way, since we cache q-adic fields, I can guarantee that the tests will use the same value)

For q=3^500

before:

```
julia> @time Hecke._trace_of_frobenius_char3_agm(t)
  3.720018 seconds (9.72 M allocations: 1.624 GiB, 0.47% gc time)
```

after:

```
julia> @time Hecke._trace_of_frobenius_char3_agm(t)
  1.218319 seconds (4.01 M allocations: 407.004 MiB, 1.14% gc time)
```

### DIFF
--- a/src/EllCrv/FinitePointCount.jl
+++ b/src/EllCrv/FinitePointCount.jl
@@ -689,8 +689,11 @@ end
 #
 ################################################################################
 
-# TODO: This should probably be part of Nemo, where we can do things more efficiently
-#   by working with padic_poly coefficients directly (instead of going through p-adic "coefficients")
+# NOTE: residue_field(R) creates a non-cached finite field and a map using it.
+#   In our context, the finite field element comes from outside, and although we
+#   create a qadic_field with the same residue field, the field objects differ,
+#   causing preimage(f::MapFromFunc, y) to fail its assertions.
+#   Thus we lift from the residue field to the qadic field manually.
 function _qadic_from_residue_element(R::QadicField, x::T; precision::Int=precision(R)) where T <: FinFieldElem
   z = R(precision=precision)
   for i in 0:degree(R)-1
@@ -862,7 +865,7 @@ function _trace_of_frobenius_char2_agm(a6::T) where T <: FinFieldElem
   # WARNING: FLINT pr: https://github.com/flintlib/flint/pull/2550
   # WARNING: for now we err for the exponent bigger than 91 (flint limit)
   d >= 92 && error("Currently exponents above 91 are not supported")
-  Qq,_ = qadic_field(2, d, precision=N)
+  Qq,_ = qadic_field(2, d, precision=N, cached=false)
 
   # TODO: we should implement a lift from F_q to Q_q (in Nemo) directly
   # TODO: instead of going through full padic coefficients
@@ -1052,8 +1055,8 @@ function _trace_of_frobenius_char3_agm(j::T) where T <: FinFieldElem
   # WARNING: FLINT pr: https://github.com/flintlib/flint/pull/2550
   # WARNING: for now we err for the exponent bigger than 57 (flint limit)
   d >= 58 && error("Currently exponents above 57 are not supported")
-  Q = qadic_field(3, d, precision=N)[1]
-  z = polynomial_ring(Q, "z")[2]
+  Q = qadic_field(3, d, precision=N, cached=false)[1]
+  z = polynomial_ring(Q, "z", cached=false)[2]
 
   # we are solving (z + 6)^3 − (z^2 + 3*z + 9)*D_k^3 = 0
   # j(E_{D_k}) = j(\mathcal{E}^k) mod 3^{k+1}
--- a/src/LocalField/Conjugates.jl
+++ b/src/LocalField/Conjugates.jl
@@ -1,10 +1,10 @@
 #XXX: valuation(Q(0)) == 0 !!!!!
 
-# silently assumes that f \in Z_q[X]
-# silently assumes that the lift is possible, that is v_p(f(r)) > 2*v_p(f'(r))
-#   see "Handbook of Elliptic and Hyperelliptic Curve Cryptography", lemma 12.8
-# since we assume the test for Newton Lift precondition to be done by the caller
-#   the derivative was already computed, so we can take it as an input argument
+# Assumes f has integral coefficients (i.e., f in Z_q[X])
+# Assumes the lift is possible, i.e., v_p(f(r)) > 2*v_p(f'(r));
+#   see "Handbook of Elliptic and Hyperelliptic Curve Cryptography", Lemma 12.8.
+# Takes the derivative df as input for efficiency (caller typically needs it
+#   anyway to verify the Newton lift precondition).
 function _newton_lift(f::PolyRingElem{QadicFieldElem}, df::PolyRingElem{QadicFieldElem}, r::QadicFieldElem, prec::Int = precision(parent(r)), starting_prec::Int = 2)
   Q = parent(r)
 
@@ -91,8 +91,10 @@ end
 @doc raw"""
     roots(Q::QadicField, f::ZZPolyRingElem; max_roots::Int = degree(f)) -> Vector{QadicFieldElem}
 
-The roots of $f$ in $Q$. Currently only simple roots are supported: that is
-if $f(a)=0$ then $f'(a) \neq 0$.
+The roots of $f$ in $Q$. The current implementation uses Newton lifting from
+the residue field, so only simple roots (in the residue field) are considered:
+a root $r$ in the residue field with $f(r) = 0$ is lifted to a root in $Q$
+only if $f'(r) \neq 0$.
 """
 function roots(Q::QadicField, f::ZZPolyRingElem; max_roots::Int = degree(f))
   k, mk = residue_field(Q)
--- a/test/LocalField/Poly.jl
+++ b/test/LocalField/Poly.jl
@@ -113,17 +113,15 @@
     X = polynomial_ring(ZZ, "X")[2]
     Y = polynomial_ring(Q, "Y")[2]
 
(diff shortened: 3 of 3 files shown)


test/LocalField/Poly.jl
```diff
@@ -84,10 +84,70 @@
...
+    Q = qadic_field(3, 4, precision=5)[1]
+    X = polynomial_ring(ZZ, "X")[2]
+    Y = polynomial_ring(Q, "Y")[2]
+
+    # For X^4-1 we have 4 roots:
+    # 1, -1, and a^3+a^2+1, -(a^3+a^2+1), where a is the generator of the residue field
+    f1 = X^4-1
+
+    # lift a^3+a^2+1
+    # TODO: we have _qadic_from_residue_element in src/EllCrv/FinitePointCount.jl
```
fieker: as a side: you know about `residue_field(Q)`? Which gives you a projection from Q to the finite field. That map has a pointwise inverse...

src/EllCrv/FinitePointCount.jl
```diff
@@ -1052,47 +1052,20 @@ function _trace_of_frobenius_char3_agm(j::T) where T <: FinFieldElem
   # WARNING: FLINT pr: https://github.com/flintlib/flint/pull/2550
   # WARNING: for now we err for the exponent bigger than 57 (flint limit)
   d >= 58 && error("Currently exponents above 57 are not supported")
-  Qq,_ = qadic_field(3, d, precision=N)
+  Q = qadic_field(3, d, precision=N)[1]
+  z = polynomial_ring(Q, "z")[2]
```
fieker: there should be `cached = false` added to both the qadic_field and the polynomial ring

src/LocalField/Conjugates.jl
```diff
@@ -75,14 +91,21 @@ end
 @doc raw"""
     roots(Q::QadicField, f::ZZPolyRingElem; max_roots::Int = degree(f)) -> Vector{QadicFieldElem}
 
-The roots of $f$ in $Q$, $f$ has to be square-free (at least the roots have to be simple roots).
+The roots of $f$ in $Q$. Currently only simple roots are supported: that is
+if $f(a)=0$ then $f'(a) \neq 0$.
```
fieker: this is slightly wrong, isn't it? the condition is that the root is simple in the residue field, not in ZZ? You need to avoid "ramified primes" (for now).
But
```
julia> roots(change_base_ring(qadic_field(3, 5; precision=10)[1], ans))
2-element Vector{QadicFieldElem}:
 3^0 + O(3^9)
 3^0 + 3^1 + O(3^9)

```
works already. Wierd. As `roots(qadic_field(3, 5; precision=10)[1], ans)` does not (same poly: `(x-1)*(x-1-3) = x^2 - 5*x + 4`

test/LocalField/Poly.jl
alexey-orlov-math: oh, i knew residue_field but completely missed that the map is returned! will update, thanks

src/EllCrv/FinitePointCount.jl
alexey-orlov-math: yes, fair

src/LocalField/Conjugates.jl
alexey-orlov-math: > the condition is that the root is simple in the residue field, not in ZZ? 

i am speaking here about the condition on newton lift: we must have v_p(f(r)) > 2 * v_p(f'(r)) in Z_q  [which is automatic for simple roots]
as an example

```
Q = qadic_field(3,4,precision=2)[1]
X = polynomial_ring(ZZ,"X")[2]
roots(Q, X^2-2*X+1)
```

gives

```
ERROR: DivideError: integer division error
Stacktrace:
 [1] inv(a::QadicFieldElem)
   @ Nemo ~/.julia/packages/Nemo/xSaoQ/src/flint/qadic.jl:452
 [2] newton_lift(f::ZZPolyRingElem, r::QadicFieldElem, prec::Int64, starting_prec::Int64)
   @ Hecke ~/math/software/Hecke.jl/src/LocalField/Conjugates.jl:15
 [3] newton_lift
   @ ~/math/software/Hecke.jl/src/LocalField/Conjugates.jl:3 [inlined]
 [4] roots(Q::QadicField, f::ZZPolyRingElem; max_roots::Int64)
   @ Hecke ~/math/software/Hecke.jl/src/LocalField/Conjugates.jl:85
```

which a bit different issue, but still.

`roots(change_base_ring(qadic_field(3, 5; precision=10)[1], ans))`
as far as i understand this is a *different* `roots` function - going into AbstractAlgebra solver for "general polynomials" (doing more work)

alexey-orlov-math: Ah, ok, i get what you mean about simple roots. I will rephrase the comment and elaborate. 

Just in case, I think we handle the case of `f'(r)=0` badly in general solver too, for example

`roots(change_base_ring(qadic_field(3, 5; precision=10)[1], X^4 + 2*X - 3))`
crashes for me: since 3 divides f'(1) we need to be more careful and we are not:

```
ERROR: BoundsError: attempt to access 0-element Vector{FqPolyRingElem} at index [1]
Stacktrace:
 [1] throw_boundserror(A::Vector{FqPolyRingElem}, I::Tuple{Int64})
   @ Base ./essentials.jl:15
 [2] getindex
   @ ./essentials.jl:919 [inlined]
 [3] Hecke.HenselCtxdr{QadicFieldElem}(f::AbstractAlgebra.Generic.Poly{QadicFieldElem}, lfp::Vector{FqPolyRingElem})
   @ Hecke ~/math/software/Hecke.jl/src/LocalField/Poly.jl:886
 [4] Hensel_factorization(f::AbstractAlgebra.Generic.Poly{QadicFieldElem})
   @ Hecke ~/math/software/Hecke.jl/src/LocalField/Poly.jl:835
 [5] roots(f::AbstractAlgebra.Generic.Poly{QadicFieldElem})
```

I can look into this later too, but for now i just fixed the thing that i run into with this roots function (i will recheck again, i think i was able to trigger wrong answer returned and not just crash)

test/LocalField/Poly.jl
alexey-orlov-math: i updated the tests, e.g.
https://github.com/thofma/Hecke.jl/pull/2168/changes#diff-44d342fe6bdd7009b3501a1d445e2096f848ebb1865eb5e94830b48d3f1401d7R134

and updated the comment in ellcurve implementation - we really need a special function there
https://github.com/thofma/Hecke.jl/pull/2168/changes#diff-57aad1aa59aca8011bdb358155efd6ab13b747979fa8ddc5d02d2823f99486abR692

src/EllCrv/FinitePointCount.jl
alexey-orlov-math: updated both char 2 and 3
https://github.com/thofma/Hecke.jl/pull/2168/changes#diff-57aad1aa59aca8011bdb358155efd6ab13b747979fa8ddc5d02d2823f99486abR868

https://github.com/thofma/Hecke.jl/pull/2168/changes#diff-57aad1aa59aca8011bdb358155efd6ab13b747979fa8ddc5d02d2823f99486abR1058

src/LocalField/Conjugates.jl
alexey-orlov-math: https://github.com/thofma/Hecke.jl/pull/2168/changes#diff-bca2a554a75d03a79c84ac1869578eec41a6e99c21b69897bdad0a554cfebacfR94

src/EllCrv/FinitePointCount.jl
```diff
@@ -689,8 +689,11 @@ end
 #
 ################################################################################
 
-# TODO: This should probably be part of Nemo, where we can do things more efficiently
-#   by working with padic_poly coefficients directly (instead of going through p-adic "coefficients")
+# NOTE: residue_field(R) creates a non-cached finite field and a map using it.
+#   In our context, the finite field element comes from outside, and although we
+#   create a qadic_field with the same residue field, the field objects differ,
+#   causing preimage(f::MapFromFunc, y) to fail its assertions.
+#   Thus we lift from the residue field to the qadic field manually.
```
fieker, line 696: What would be a useful API? Looking at residue_field, the finite field is cached on the unramifed extension. Thus given a finite field, one could construct the qadic field (you're doing it) and store (and create) the map at this point....

alexey-orlov-math: > Thus given a finite field, one could construct the qadic field (you're doing it) and store (and create) the map at this point....

Yeah, i agree something like this will work - i am essentially waiting for FLINT release to have my "create qadic with custom modulus" new api - at this point i will be indeed adding something akin to `qadic_field(R::FiniteField)`. I think this will be the perfect time/place to do this addition

PR #2170, opened 2026-02-06
feat: reduction and canonical symbols for `ZZGenus`

StevellM:
Implement reduction of lattice genus and canonical symbols

Towards https://github.com/StevellM/DefLatDB/issues/4

### DIFF
--- a/docs/src/Hecke.bib
+++ b/docs/src/Hecke.bib
@@ -1,13 +1,14 @@
+
 @Article{AGM20,
   author        = {Allcock, Daniel and Gal, Itamar and Mark, Alice},
   title         = {The Conway-Sloane calculus for 2-adic lattices},
+  mrnumber      = {4162280},
   journal       = {Enseign. Math.},
   fjournal      = {L'Enseignement Math\'ematique},
   volume        = {66},
-  year          = {2020},
   number        = {1-2},
   pages         = {5--31},
-  mrnumber      = {4162280},
+  year          = {2020},
   doi           = {10.4171/LEM/66-1/2-2}
 }
 


simonbrandhorst:
Cool! How well have you tested this? 
Please leave the old equality code alive for testing. (Since we are very confident that it is correct. )

StevellM:
Yes that's true. My current implementation is not current, it misses some cases which I need to look into.

StevellM:
The function for the canonical symbol should be fixed now. For what I have tried (over 25000 possible genera, with negative scales and up to rank 12), we obtain the same results as we the previous equality and hash functions. By saving the canonical symbol, we also tend to have less allocations and a slightly faster enumeration of possible symbols. I still need to find more tests for the test file; though it seems quite good for the moment.

StevellM:
Yes and no, here are the one I tried:
```
Hecke._local_genera(ZZ(2), 1:10, 1:6, 0, 5, false)
Hecke._local_genera(ZZ(2), 1:8, 1:6, -2, 2, false)
```
In the first case I allowed some large levels, the second one I allowed non-integral scales. I should probably try with examples where we have a long sequence of nontrivial constituents with adjacent scales (so like from 1 to 1024 for instance).

src/QuadForm/Quad/ZGenus.jl
```diff
@@ -930,30 +947,403 @@ function _is2adic_genus(symbol::Vector{Vector{Int}})
...
+
+@doc raw"""
+    canonical_symbol(g::ZZLocalGenus; odd_ones::Bool=true) -> String
+
+Return the canonical symbol for the genus of ``p``-adic lattices defined
+by `g`. The ouput is given in the form of a string.
+
+If ``p`` is odd, the symbol is uniquely determined by the invariants of `g`.
+
+If ``p == 2``, then we use Conway--Sloane canonicalization procedure following
```
simonbrandhorst:
```suggestion
If ``p == 2``, then we use the Conway--Sloane canonicalization procedure following
```

simonbrandhorst (approved):
This is a nice improvement. 

It is surprising that everything is formulated in terms of strings. 
I would have expected a `canonical_symbol` as a list of tuples of integers. 
and then some functions like `trains` and `compartments` etc. to structure things. 

But okay, the tests pass.

test/QuadForm/Quad/ZGenus.jl
```diff
@@ -613,3 +613,19 @@ end
   ## Issue 1103
   @test sprint(show, "text/plain", genus(root_lattice(:E, 8), 2)) isa String
 end
+
+@testset "Canonical symbols" begin
+  # From Allcock, Gal, Mark "The Conway--Sloane calculues for 2-adic lattices"
```
simonbrandhorst:
```suggestion
  # From Allcock, Gal, Mark "The Conway--Sloane calculus for 2-adic lattices"
```

docs/src/Hecke.bib
```diff
@@ -1,3 +1,15 @@
+@Article{AGM20,
+  author        = {Allcock, Daniel and Gal, Itamar and Mark, Alice},
+  title         = {The {C}onway-{S}loane calculus for 2-adic lattices},
+  journal       = {Enseign. Math.},
+  fjournal      = {L'Enseignement Mathématique},
```
StevellM:
```suggestion
  author        = {Allcock, Daniel and Gal, Itamar and Mark, Alice},
  title         = {The Conway-Sloane calculus for 2-adic lattices},
  journal       = {Enseign. Math.},
  fjournal      = {L'Enseignement Math\'ematique},
```

PR #2175, opened 2026-02-07
Changes to order computation for elliptic curve points

eoinmackall:
Fixed a bug in torsion_bound
- the check that p divides disc(E) was insufficient to verify that E has good reduction at primes above p
- a concrete example is as follows:
    ```julia
    F, a = quadratic_field(-2)
    E = elliptic_curve(F, [1-a,0])
    torsion_bound(E,10)
    ```
   This would previously error, as one prime above 3 has bad reduction and one has good reduction. It looks like SageMath avoids this issue by taking the norm of the discriminant of $E$ to $\mathbb{Q}$, and checking divisibility there (https://github.com/sagemath/sage/blob/develop/src/sage/schemes/elliptic_curves/ell_torsion.py#L282).

Also changed the way the order of an elliptic curve point is calculated
- previous algorithm was an iterative check for the order (check if the order is 1,2,3,4,...) up to `torsion_bound`.
- new algorithm uses that `torsion_bound` returns a multiple of the order. Requires factoring this number, and adjusting downward to find the exact order. Removed allocations by using in-place operations. (Note for the new algorithm there could be potential downsides in the -- unlikely(?) -- case where `torsion_bound` returns a large number and the input point for `order` has a small order.)

### DIFF
--- a/src/EllCrv/Torsion.jl
+++ b/src/EllCrv/Torsion.jl
@@ -352,7 +352,7 @@ map has trivial kernel when restricted to the torsion subgroup of $E$.
 function torsion_bound(E::EllipticCurve{AbsSimpleNumFieldElem}, n::Int)
   K = base_field(E)
   R = ring_of_integers(K)
-  disc_K = ZZ(discriminant(K))
+  disc_K = ZZ(discriminant(R))
   disc_E = R(discriminant(E))
 
   # If p>2 does not divide disc_K, then the kernel of the reduction


src/EllCrv/Torsion.jl
```diff
@@ -324,33 +342,50 @@ end
...
 """
 function torsion_bound(E::EllipticCurve{AbsSimpleNumFieldElem}, n::Int)
   K = base_field(E)
   R = ring_of_integers(K)
-  badp = bad_primes(E)
-  #First prime is 3
-  p = next_prime(2)
-  i = 0
-  bound = 0
   disc_K = ZZ(discriminant(K))
```
thofma:
```suggestion
  disc_K = ZZ(discriminant(R))
```
While we are at it, we might as will fix this.

PR #2184, opened 2026-02-16
fix: coercion for integral split

thofma:


thofma:
This works, if one does it like we do in the tests:
```
  k = GF(5)
  kx, x = rational_function_field(k, "x")
  kt = parent(numerator(x))
  ky, y = polynomial_ring(kx, "y")
  F, a = function_field(y^3+(4*x^3 + 4*x^2 + 2*x +2)*y^2 + (3*x+3)*y +2)
  Ofin = integral_closure(kt, F)
  R = localization(base_ring(F),degree)
  Oinfi = integral_closure(R,F)
```
Major pain point is that we do not have a function go get the polynomial ring out of a rational function field.

fieker:
I'm not sure what the interface should be. If Q(t) was constructed as field_of_fractions(R) then it would be "fine" to expect to be able to ask for R. However if Q(t) is "just" a rational function field, the question makes no sense.... (mathematically) or maybe as
```
Qt, t = rational_function_field(QQ, :t)
Qs, s = QQ[t]
```
to indicate I want the same indeterminate? or some form of `sub(Qt, t)`?

fieker:
or, more functorial via `embed(Qs, Qt, s->t)`, but then one would need to pass in maps rather than rings - which is also awkward and non natural...(unless categories)

PR #2188, opened 2026-02-20
Add some tests

joschmitt:
... and fix related typos.

As observed some time ago, the "quotient of order by ideal as algebra" functionality was not too thoroughly tested. I hope it is better like this; let's see what code coverage says.


joschmitt:
Looking at the coverage of the file: why is the "ideal by ideal" function for the absolute case in there three times? (And only covered once.)

But that is independent of this pull request, which should be done from my side.

thofma:
Merci

PR #2193, opened 2026-02-23
LMFDB: add some functions

StevellM:
Add some functions for the snippets of codes of the lattice part of the LMFDB.

### DIFF
--- a/src/QuadForm/ShortVectors.jl
+++ b/src/QuadForm/ShortVectors.jl
@@ -324,6 +324,10 @@ function successive_minima_with_vectors(L::ZZLat)
     cur_mi = S[cur_i + 1][2]
     next_i = findlast(x -> x[2] == cur_mi, S)
     @assert next_i > cur_i
+    # the following constructs a potential large marix
+    # better would be a "streaming" version which takes as input only the vector
+    # but we don't have this for rational matrices
+    # (we only care about rank = dimension of rational span)
     B = echelon_form(vcat(B, matrix(QQ, [x[1] for x in view(S, cur_i+1:next_i)])); trim = true)
     if nrows(B) > length(res)
       for _ in 1:(nrows(B) - length(res))


src/QuadForm/Quad/ZLattices.jl
```diff
@@ -2606,6 +2655,33 @@ function _short_vector_generators(L::ZZLat; up_to_sign::Bool=false)
 end
 
 
+@doc raw"""
+    _shortest_vectors_sublattice(L::ZZLat; check::Bool=true)
```
simonbrandhorst:
```suggestion
    shortest_vectors_sublattice(L::ZZLat; check::Bool=true)
```

src/QuadForm/Lattices.jl
```diff
@@ -517,6 +517,13 @@ Return whether the rational span of the lattice `L` is definite.
 """
 @attr Bool is_definite(L::AbstractLat) = is_definite(rational_span(L))
 
+@doc raw"""
+    is_indefinite(L::AbstractLat) -> Bool
+
+Return whether the rational space of the lattice `L` is indefinite.
```
simonbrandhorst:
```suggestion
Return whether the  `L` is indefinite.
```

src/QuadForm/Lattices.jl
```diff
@@ -517,6 +517,13 @@ Return whether the rational span of the lattice `L` is definite.
...
+"""
+is_indefinite(L::AbstractLat) = !is_definite(L)
```
simonbrandhorst: This is not correct, a semidefinite lattice is in general not indefinite.

simonbrandhorst:
```suggestion
is_indefinite(L::AbstractLat) = !is_definite(L) && !is_degenerate(L)
```

src/QuadForm/ShortVectors.jl
```diff
@@ -140,6 +140,33 @@ function minimum(L::ZZLat)
   return L.minimum
 end
 
+###############################################################################
+#
+#  Density
+#
+###############################################################################
+
+@attr Float64 function center_of_density(L::ZZLat)
```
simonbrandhorst:
```suggestion
@attr Float64 function center_density(L::ZZLat)
```
It is called "center density" in the literature.

simonbrandhorst: But why should this be a `Float64`? (If it has to be a float, then probably a `Float` and not a `Float64`.)

src/QuadForm/ShortVectors.jl
```diff
@@ -140,6 +140,33 @@ function minimum(L::ZZLat)
...
+  rho = sqrt(mu)
+  return rho^(rank(L))/sqrt(d)
+end
+
+###############################################################################
+#
+#  Hermite number
+#
+###############################################################################
+
```
simonbrandhorst, line 222: Docstring please :-). And same thing about Float applies.

src/QuadForm/ShortVectors.jl
StevellM: I think it does not work like `Int`, there is no such thing as `Float` afaik. Hence `Float64` :)

src/QuadForm/ShortVectors.jl
StevellM: Yes, right, this is planned, I wanted to find a good reference to define them properly. Probably SPLG defines them

src/QuadForm/Lattices.jl
StevellM: Good point, thanks!

src/QuadForm/ShortVectors.jl
StevellM: Anyway, I will turn them into a `RealFieldElem` (Tommy's suggestion)

simonbrandhorst: Ah right :-).

src/QuadForm/Quad/ZGenus.jl
```diff
@@ -1605,6 +1605,26 @@ function Base.show(io::IO, ::MIME"text/latex", g::ZZLocalGenus)
   print(io, str)
 end
 
+@doc"""
+    conway_symbol(G::ZZGenus) -> String
+
+Return a Conway symbol describing the genus $G$. Note that this is
+in general not unique.
+"""
+function conway_symbol(G::ZZGenus)
```
simonbrandhorst: Why not use the unique symbol that Stevell implemented?

src/QuadForm/Quad/ZLattices.jl
```diff
@@ -1064,11 +1064,39 @@ function norm(L::ZZLat)
...
+#  Level
 #
-################################################################################
+###############################################################################
+
+@doc raw"""
+    level(L::ZZLat) -> QQFieldElem
+
+Return the level of $L$, that is the smallest positive rational number $l$ such
+that $lL^\vee \subset L$ where $L^\vee$ is the dual lattice of $L$. If $L$ is
```
simonbrandhorst: Your definition is off for the zero lattice: The smallest positive rational number does not exist.

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -212,7 +212,7 @@ Input:
    classes in the genus of ``L`` to compute in the outer scope before stopping
    the enumeration.
 """
-function neighbours(
+function _neighbours(
```
simonbrandhorst, line 221: With the new changes this function looks like inconsistent mess even more than before.
I think we put too much logic in "algorithm". 
I suggest to have two keyword arguments.
1. `algorithm` in `[:orbit, :random]` 
2. `mode` in `[:spinor_generators, :isometry_classes]`
The second argument conflicts somewhat with the callback function so this needs to be documented carefully.

src/QuadForm/ShortVectors.jl
```diff
@@ -155,15 +263,80 @@ This is the number of non-overlapping spheres touching any
 other given sphere.
 """
 @attr Int function kissing_number(L::ZZLat)
-  @req rank(L) > 0 "Lattice must have positive rank"
+  @req is_definite(L) > 0 "Lattice must have positive rank"
```
simonbrandhorst: this look fishy

src/QuadForm/ShortVectors.jl
```diff
@@ -155,15 +263,80 @@ This is the number of non-overlapping spheres touching any
...
+
+Given a positive definite lattice $L$, return the successive minima of $L$.
+See also [`successive_minima_with_vectors`](@ref).
+"""
+function successive_minima(L::ZZLat)
+  return successive_minima_with_vectors(L)[1]
+end
+
+@doc raw"""
+    successive_minima_with_vectors(L::ZZLat) -> Vector{QQFieldElem}, Vector{ZZRingElem}
```
simonbrandhorst, line 290: Please add the definition. 
Also I believe the implementation should use `short_vectors_iterator` in order to not let the memory explode.

src/QuadForm/ShortVectors.jl
```diff
@@ -155,15 +263,80 @@ This is the number of non-overlapping spheres touching any
...
+  cur_mi = mi
+  res = QQFieldElem[cur_mi]
+  resv = [S[1][1]]
+  cur_i = findlast(x -> x[2] == mi, S)
+  B = echelon_form(matrix(QQ, [x[1] for x in view(S, 1:cur_i)]); trim = true)
+  while length(res) < n
+    cur_mi = S[cur_i + 1][2]
+    next_i = findlast(x -> x[2] == cur_mi, S)
+    @assert next_i > cur_i
+    B = echelon_form(vcat(B, matrix(QQ, [x[1] for x in view(S, cur_i+1:next_i)])); trim = true)
```
simonbrandhorst, line 331: This looks like you create a huge matrix. Perhaps better use 
`reduce_mod_rref!`  or `_row_span!`

src/QuadForm/Quad/ZGenus.jl
thofma: We decided to remove it for now.

src/QuadForm/Quad/ZLattices.jl
thofma: Thanks! Reformulated it.

src/QuadForm/ShortVectors.jl
thofma: I added a definition.

We also thought about `short_vectors_iterator`, but the algorithm requires the short vectors sorted by norm, but this is not guaranteed by the iterator variant.

src/QuadForm/ShortVectors.jl
thofma: Yeah, it is probably large. But `_row_span!` creates the span as a Z-module, but this is too expensive. I only care about the rank of the rational span. Maybe we can keep it as is and if there is some concrete example where it is too slow we can optimize it?

src/QuadForm/Quad/ZGenusRep.jl
StevellM: Implemented, should be better now

src/QuadForm/ShortVectors.jl
simonbrandhorst: Ok, I see the trouble. For the iterator version, you can provide upper and lower bounds and probably also resume a started iteration etc.? But maybe it is not worth the trouble.
Or you could put a safeguard that if there are more than 1 000 0000 vectors, that it throws instead of ooming.

src/QuadForm/ShortVectors.jl
simonbrandhorst: Okay. Maybe leave a comment in the code, that it can be optimized there?

thofma: done

PR #2194, opened 2026-02-24
ZZLat maintenance

simonbrandhorst:


src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -726,13 +771,13 @@ function enumerate_definite_genus(
       push!(res, L)
     end
   else
-    res = copy(known)
+    res = known#copy(known)
```
StevellM: This does not break anything ? I do not remember why I use a `copy` here, maybe it was unnecessary

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -393,9 +391,12 @@ function _neighbours(
       if !keep
         vain[] += 1
         continue
+      else
+        # we want the vain iterations since the last time we found a new class
+        vain[] = 0
       end
 
-      vain[] = Int(0)
+      vain[] = 0
```
StevellM:
```suggestion
      end
      # we want the vain iterations since the last time we found a new class
      vain[] = 0
```
If `!keep` we add 1 to `vain[]` and then continue the loop, otherwise we do whatever is after and the first thing is setting up `vain[]=0` so the `else` statement is redundant here.

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -707,9 +722,40 @@ function enumerate_definite_genus(
...
+    save_path::Union{String, Nothing}=nothing,
+    use_mass::Bool=true,
+    _missing_mass::Union{QQFieldElem, Nothing}=nothing,
+    vain::Base.RefValue{Int}=Ref{Int}(0),
+    stop_after::IntExt=inf,
+    max::IntExt=inf,
+    add_spinor_generators::Bool=true,
+    scaling_factor=nothing
+  )
+@req !save_partial || !isnothing(save_path) "No path mentioned for saving partial results"
```
StevellM:
```suggestion
  @req !save_partial || !isnothing(save_path) "No path mentioned for saving partial results"
```

src/QuadForm/Quad/ZGenusRep.jl
```diff
@@ -736,13 +782,13 @@ function enumerate_definite_genus(
       push!(res, L)
     end
   else
-    res = copy(known)
+    res = known
```
StevellM, line 781: This does not break anything ? I do not remember why I use a copy here, maybe it was unnecessary

src/QuadForm/Lattices.jl
```diff
@@ -1595,6 +1595,8 @@ Given a definite lattice `L`, return the order of the automorphism group of `L`.
 Setting the parameters `depth` and `bacher_depth` to a positive value may improve
 performance. If set to `-1` (default), the used value of `depth` is chosen
 heuristically depending on the rank of `L`. By default, `bacher_depth` is set to `0`.
+
+The underlying algorithm is by Plesken and Souvignier [PS97](@cite).
```
StevellM, line 1599: We had a problem when I added a reference to Hecke. Before we do any Hecke release, we should add the reference to the OSCAR bibliography too.

src/QuadForm/Torsion.jl
```diff
@@ -166,7 +166,7 @@ Gram matrix quadratic form:
   return T
 end
 
-discriminant_group(L::ZZLat, n::Int) = primary_part(discriminant_group(L), n)[1]
+discriminant_group(L::ZZLat, n::Union{Integer,ZZRingElem}) = primary_part(discriminant_group(L), n)[1]
```
StevellM:
```suggestion
discriminant_group(L::ZZLat, n::IntegerUnion) = primary_part(discriminant_group(L), n)[1]
```
We already have a type for that, this suggestion can be ignored

src/QuadForm/Lattices.jl
```diff
@@ -1654,7 +1666,7 @@ function is_isometric_with_isometry(L::AbstractLat{<: NumField}, M::AbstractLat{
   K = base_field(E)
   @assert base_ring(V) == base_ring(W)
   @assert base_ring(L) == base_ring(M)
-
+  is_definite(L) || error("not implemented")
```
StevellM:
```suggestion
  @req is_definite(L) "Only implemented for definite lattices"
```

src/QuadForm/Quad/ZGenusRep.jl
StevellM: I agree with removing the previous `vain[]+=1` since in fact we have not an actual neighbour, this should allow the algorithm to run a bit more before stopping random search.

src/QuadForm/Lattices.jl
simonbrandhorst: Oh, that is tiresome.

src/QuadForm/Quad/ZGenusRep.jl
simonbrandhorst: this is the `_...!` method, the old one is unchanged.

PR #2197, opened 2026-02-27
Rigorous integration along straight lines

stefnomden:
For straight lines, added a function which computes a rigorous bound for integrands on ellipses. Previously this was heuristic. The computation of this bound is based on [https://arxiv.org/pdf/2208.12377](url) .

### DIFF
--- a/src/RieSrf/PeriodMatrix.jl
+++ b/src/RieSrf/PeriodMatrix.jl
@@ -92,8 +92,10 @@ function big_period_matrix(RS::RiemannSurface)
   bound_temp = Vector{ArbFieldElem}()
   for path in paths
     for subpath in get_subpaths(path)
-      if path_type(subpath) == 0
+      if path_type(subpath) == 0 && RS.integration_method == "rigorous"
         compute_ellipse_bound_rigorous(subpath, dif_basis, int_group_rs, RS)
+      elseif path_type(subpath) == 0 && RS.integration_method == "heuristic"
+        compute_ellipse_bound_heuristic(subpath, embedded_differentials, int_group_rs, RS)
       else 
         compute_ellipse_bound_heuristic(subpath, embedded_differentials, int_group_rs, RS)
       end
@@ -343,7 +345,6 @@ end
 ###  Disney-Hogg, and Wuqian Effie Gao, as presented in https://arxiv.org/pdf/2208.12377. The implementation 
 ###  here is Strategy 1 from this paper.
 function compute_ellipse_bound_rigorous(subpath, dif_basis, int_group_rs, RS)
-
   subpath.integration_scheme_index = length(int_group_rs)
 
   v_start = start_point(subpath)
--- a/src/RieSrf/RiemannSurface.jl
+++ b/src/RieSrf/RiemannSurface.jl
@@ -139,6 +139,10 @@ mutable struct RiemannSurface
   # A list of bounds used duting computations
   bounds::Vector{ArbFieldElem}
 
+  #String specifying the integration method used when integrating differential forms.
+  #The options are "heuristic" and "rigorous". It is set to "heuristic" by default
+  integration_method::String
+
   #A collection of fields and error bounds needed to ensure correctness
   #TODO: Check which ones are actually used and necessary. Optimize this.
   prec::Int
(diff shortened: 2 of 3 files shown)


src/RieSrf/PeriodMatrix.jl
```diff
@@ -84,14 +84,19 @@ function big_period_matrix(RS::RiemannSurface)
   # Random precision. Probably needs to become a heuristic.
   max_prec = 187
   embedded_differentials = [embed_mpoly(g, v, max_prec) for g in differentials]
+  dif_basis = [omega.f for omega in basis_of_differentials(RS)]
```
Tobias271828:
```suggestion
  dif_basis = [omega.f for omega in diff_base]
```
Is there a reason, why the function call `basis_of_differentials(RS)` is used twice in this file?

src/RieSrf/PeriodMatrix.jl
```diff
@@ -333,6 +338,61 @@ function compute_ellipse_bound(subpath::CPath, differentials_test, int_group_rs,
...
+
+  subpath.integration_scheme_index = length(int_group_rs)
+
+  v_start = start_point(subpath)
+  v_end = end_point(subpath)
+  
+  #these values are the 't values' 
+  rs = [(2 * alpha - (v_start + v_end))/(v_end - v_start) for alpha in discriminant_points(RS)]
+
+  function construct_M(g, RS, centre, radius, rs, v_start, v_end)
```
Tobias271828: For me, it seems a bit weird to define a function in another function. As far as I know, this can result in unnecessary recompilations of the inner function (`construct_M`). Namely, each time the outer function (`compute_ellipse_bound_rigorous`) is being executed, the inner function gets overwritten, and thus possibly recompiled before the next call.

Is there a need for this function nesting here? Otherwise, you could just move the inner function outside.

src/RieSrf/PeriodMatrix.jl
```diff
@@ -333,6 +338,61 @@ function compute_ellipse_bound(subpath::CPath, differentials_test, int_group_rs,
...
+    gmin = minpoly(g)
+    lis = [denominator(coeff(gmin, i)) for i in (0:degree(gmin))]
+    gmin = lcm(lis) * gmin
+    CC = RS.complex_field
+    v = embedding(RS)
+    _, CCx = polynomial_ring(CC, "x")
+    coeffs = [numerator(coeff(gmin,i)) for i in (0:degree(gmin))]
+    #precompose with the path
+    if base_ring(base_ring(parent(gmin))) == QQ
+      coeffs = [sum(CC(coeff(a,i)) * ( (CCx+1)*v_end/2 + (1-CCx)*v_start/2 )^i for i in (0:length(coefficients(a)))) for a in coeffs]
```
Tobias271828: Where does this formula come from? By shortly glimpsing over the paper, I didn't saw it. I guess, I just overlooked something. If this is build out of several formulas in the paper, it might make sence to describe it in a comment in the code.

src/RieSrf/PeriodMatrix.jl
```diff
@@ -333,6 +338,61 @@ function compute_ellipse_bound(subpath::CPath, differentials_test, int_group_rs,
   end
 end
```
Tobias271828, line 342:
```suggestion

###  The following function compute_ellipse_bound_rigorous, together with the function construct_M,
###  are an implementation of the algorithm for rigorous integration along straight lines by Nils Bruin, Linden
###  Disney-Hogg, and Wuqian Effie Gao, as presented in https://arxiv.org/pdf/2208.12377.
```

src/RieSrf/PeriodMatrix.jl
stefnomden: Thank you for the question. This formula does not occur in the paper, it is merely a small detail: The functional equation that g satisfies has as coefficients rational functions which can be considered as maps from the straight line P -> CC connecting v_start and v_end. However, [BDG24] requires this functional equation to hold on [-1,1] so what we do here is precompose the coefficients with the formula for the straight path [-1,1] -> P as to get the desired form for our equation. 

I will add a comment that explains this.

Tobias271828: Thank you for the explaination!

JHanselman:
> Hi Simon. I've run them locally and it also times out. I've also traced down what the issue is: the test file RieSrf/RiemannSurface.jl contains a rather large example (genus 21). For the old version this was fine but for this version we calculate the minimal polynomial of a certain function field element which is incredibly big. This is where the test gets stuck.
> 
> Finding the minimal polynomial is essential to obtain a rigorous bound. I'd propose that we allow the user to decide the type of integration, either rigorous or heuristic (which is set to rigorous by default). Then we change the integration in the test to heuristic.
> 
> @JHanselman What do you think?

Yeah, sounds good to me. That was more or less my original plan anyway.

PR #2226, opened 2026-03-27
feat: convenient functions for `FinGenAbGroup`

StevellM:
- Add computation of ``n``-torsion subgroup (I do not think it is already available, at least I have not found it).
- Add some methods related to ``p``-types of finite abelian subgroups and common subgroups of a pair of finite abelian groups.

I do not have strong opinions on the name of the functions. I am happy to rework the "common subgroups" part if there are any ideas of cool features to add or better ways to approach the problem. The output may look weird; in the current implementation I have, for my purpose, this is how I use it. But i am OK to change this also.


src/GrpAb/SubgroupEnum.jl
```diff
@@ -1014,18 +1011,19 @@ function _primary_parts_subgroups(
...
       pps[p] = tmp
     end
   else
-    for (p, v) in factor(order(G))
+    for p in prime_divisors(last(elG))
+      v = sum(Int[valuation(a, p) for a in elG]; init=Int(0))
       tmp = Set{Vector{Int}}()
-      for _v in 1:v
-        union!(tmp, _psubgroups_types(G, p, _v))
+      for _v in 0:v
```
thofma: is this fixing a bug? (the change from `1` to `0`)?

StevellM: Kind of, I had problem with the trivial subgroup. But I just realize that there might be a flaw in the design suggested in this PR... the features are, maybe, too much compared to what I need. I will make it a draft for the moment until I come with the final version of my code. Then I will modify this part accordingly (probably I would just need the `_psubgroups_type` function but as a `_common_psubgroups_type` and the rest would be useless.

StevellM:
Ah yes, but this was fixing a bug in the code I added, not in a code already available in Hecke.. Actually the tests failed because when I made the last commit, I forgot to include the change to the test file... So there was a test, just missing in this PR. When I come up with the better design, there will be more tests.

thofma:
Thanks. Since the "new" functions are not exported, the name is fine. The torsion subgroup stuff is much welcome.

PR #2230, opened 2026-04-05
Function fields: some cleanup for divisors

alexey-orlov-math:
- Type Stability: parametrized `Divisor` with explicit constraints: `Divisor{S <: Generic.FunctionField, T1 <: PolyRing, T2 <: KInftyRing}`; added extra type assertions to `GenOrdIdl.basis_matrix`, and `GenOrdFracIdl.denominator` along with explicit matrix construction (from view) in `GenOrdIdl` constructor
- Added valuation for `GenOrdFracIdl`, fixed valuation of `GenOrdIdl` argument order
- Fixed Divisor.isless to check support primes of both divisors
- Extracted trivial helpers to remove excessive duplication of code handling finite/infinite support
- Added explicit is_separable guards to functions that assume it implicitly due to current implementation (separating_element, canonical_divisor, genus, index_of_speciality)
- Added lcm for divisors, added tests for gcd/lcm
- Added explicit Riemann-Roch theorem testing for function fields over QQ(x) and F_q(x) and across different genera

### DIFF
--- a/src/GenOrd/FractionalIdeal.jl
+++ b/src/GenOrd/FractionalIdeal.jl
@@ -154,10 +154,27 @@ Base.:*(A::T, B::T) where T <: GenOrdFracIdl = prod(A, B)
 
 
 function Base.:(+)(a::T, b::T) where T <: GenOrdFracIdl
-  d = lcm(denominator(a), denominator(b))
-  ma = divexact(d, denominator(a))
-  mb = divexact(d, denominator(b))
-  return GenOrdFracIdl(order(a)(ma)*numerator(a) + order(b)(mb) * numerator(b),d)
+  d = lcm(denominator(a; copy=false), denominator(b; copy=false))
+
+  ma = divexact(d, denominator(a; copy=false))
+  I  = order(a)(ma) * numerator(a; copy=false)
+
+  mb = divexact(d, denominator(b; copy=false))
+  J  = order(b)(mb) * numerator(b; copy=false)
+
+  return GenOrdFracIdl(I + J, d)
+end
+
+function Base.intersect(a::T, b::T) where T <: GenOrdFracIdl
+  d = lcm(denominator(a; copy=false), denominator(b; copy=false))
+
+  ma = divexact(d, denominator(a; copy=false))
+  I  = order(a)(ma) * numerator(a; copy=false)
+
+  mb = divexact(d, denominator(b; copy=false))
+  J  = order(b)(mb) * numerator(b; copy=false)
+
+  return GenOrdFracIdl(intersect(I, J), d)
 end
 
 ################################################################################
--- a/src/GenOrd/FractionalIdeal.jl
+++ b/src/GenOrd/FractionalIdeal.jl
@@ -356,7 +356,7 @@ function Hecke.factor(A::GenOrdFracIdl)
   primes = Dict{GenOrdIdl,Int}()
   for (f,e) in factors
     for (p,r) in prime_decomposition(O,f)
-      p_val = valuation(p,A_num) - valuation(p, A_den)
+      p_val = valuation(A_num, p) - valuation(A_den, p)
       if p_val != 0
         primes[p] = p_val
       end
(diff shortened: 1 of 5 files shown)


src/GenOrd/FractionalIdeal.jl
```diff
@@ -118,7 +118,7 @@ function assure_has_numerator_and_denominator(a::GenOrdFracIdl)
 
   B, d = integral_split(basis_matrix(a, copy = false), coefficient_ring(order(a)))
   a.num = GenOrdIdl(order(a), B)
-  a.den = d
+  a.den = d::elem_type(T)
```
thofma, line 121: Hm, why add this type conversion? I would guess that either `d` has the right type, or we need to do some coercion `(...)(d)`?

alexey-orlov-math: Essentially i was trying to make `@inferred gcd(x,y)` work for divisors. So i was going through all the places where we use `den` with "unknown" type and adding type assertions to help the compiler
As i understand this kinda acts like an assert on "d has the right type"

thofma: Hm yes, but think the field `a.den` has an abstract type, so adding the type assertion will not help. It should be the same as in the following example:
```
julia> mutable struct A
         x::Number
       end

julia> function f()
         z = A(1)
         z.x = 1::Int
         return z.x
       end;

julia> Base.return_types(f)
1-element Vector{Any}:
 Number
```
Any access to `z.x` will be inferred to to return `Number`, independent of what it was set to.

alexey-orlov-math: yes, but we are doing two things (again, maybe both are not needed, but this is still useful as an "assert"):
in here we set den with type assertion, but we also do below
https://github.com/thofma/Hecke.jl/pull/2230/changes#diff-414899046ec06ff89493ee2e29d5ed959dd989513d71d571fa2bc50acf7b49d7R132
with type assertion on return type of denominator "getter". So everywhere we use denominator compiler can infer the type of it. Arguably this is more important than setting den with type, but my thinking was exactly to have type assertion to happen at a place where "contract on types" is broken.
As a continuation: i was thinking about adding tests for GenOrd to be used with non-function-fields (as in - they should be usable with number fields even if less efficient), and in doing so adding some `@inferred` and trying to fix more type instabilities. The current changes are the ones that were caught with gcd/lcm (+, *, intersect)

thofma: OK, sounds good.

(And yes, I agree that it would be a good idea to throw the `GenOrd` things at Z-orders of a number field. I remember vaguely that we did this in the past by computing a maximal order/integral closure of something.)

fieker:
On Mon, Apr 06, 2026 at 11:46:03PM -0700, alexey-orlov-math wrote:
> 
> 
> 
> > @@ -118,7 +118,7 @@ function assure_has_numerator_and_denominator(a::GenOrdFracIdl)
>  
>    B, d = integral_split(basis_matrix(a, copy = false), coefficient_ring(order(a)))
>    a.num = GenOrdIdl(order(a), B)
> -  a.den = d
> +  a.den = d::elem_type(T)
> 
> yes, but we are doing two things (again, maybe both are not needed, but this is still useful as an "assert"):
> in here we set den with type assertion, but we also do below
> https://github.com/thofma/Hecke.jl/pull/2230/changes#diff-414899046ec06ff89493ee2e29d5ed959dd989513d71d571fa2bc50acf7b49d7R132
> with type assertion on return type of denominator "getter". So everywhere we use denominator compiler can infer the type of it. Arguably this is more important than setting den with type, but my thinking was exactly to have type assertion to happen at a place where "contract on types" is broken.
> As a continuation: i was thinking about adding tests for GenOrd to be used with non-function-fields (as in - they should be usable with number fields even if less efficient), and in doing so adding some ***@***.***` and trying to fix more type instabilities. The current changes are the ones that were caught with gcd/lcm (+, *, intersect)
> 
There is a cool application of the GenOrd for number fields: maxmal
orders in localisations. Localisations of Z are available and the GenOrd
stuff allows to compute integral closures in this case

PR #2233, opened 2026-04-11
Function fields: improvements for differentials

alexey-orlov-math:
- Fixed differential implementation bugs: `differential(f)` was wrong, missing the chain rule contribution from y; equality check was wrong (was comparing element with itself)
- Type stability improvements: FunFldDiff is now a parameterized immutable struct without a `function_field` member (it was a synonym to `parent(f))`; ensured basis_of_differentials has a concrete inferred return type
- Some minor polish, including `@req` instead of `@assert`, missing parent checks, and general code cleanup
- Tests rewritten: basic arithmetic tests for curves of different genus over QQ and several finite fields; basis of differentials tests for elliptic and hyperelliptic curves of different genus over QQ and several finite fields

Known Issues/Limitations:
- differential() requires the extension to be separable (we divide by `dp/dy` for `p` defining polynomial). As discussed on Slack, currently inseparable extensions are out of scope
- uncovered the issue with `is_zero(Divisor)`: it checks if the ideal is `1*O` and returns false for `u*O` with `u` unit (not equal to 1). Comment left in test explaining a workaround we do, but this should be looked into in a separate pr
- gen(FunctionField) [in AbstractAlgbera] has inferred return type Any


thofma:
@JHanselman do you want to have a look?

JHanselman:
> @JHanselman do you want to have a look?

Looks good to me.

PR #2240, opened 2026-04-19
ellcrv: refactor local data infrastructure (Tate's algorithm, local_height)

alexey-orlov-math:
- New internal abstraction: `DiscreteValuation` (`src/EllCrv/DiscreteValuation.jl`). A single interface for local algorithms that only care about "valuation" properties. For now it is kept in EllCrv module, but can be moved if it is deemed useful in other parts. Currently implemented for QQ, absolute simple number fields, and rational function fields
- New public `EllipticCurveLocalData` struct that collects local data. New `EllipticCurveReduction.ReductionType` enum: scoped inside an `EllipticCurveReduction` submodule, so values are spelled `EllipticCurveReduction.additive`, `EllipticCurveReduction.split_multiplicative`, etc. Added `tates_algorithm_local(E, P, ::Type{EllipticCurveLocalData})` overload to return `EllipticCurveLocalData` instead of tuple (old `tates_algorithm_local(E, P)` is preserved). Also refactored `KodairaSymbol` (simple code improvements).
- Tate's algorithm now uses `DiscreteValuation` (instead of closure-based arguments), the code prettified and commented (with explicit mapping to parts of Cremona/Silverman description)
- (Non-archimedean) Local Height code for QQ/NF unified using `DiscreteValuation`. Since we are using Tate's algorithm explicitly, use kodaira symbol and reduction type directly instead of checking valuation of c_4 and discriminant (which are theorems under Silverman's presentation). 
- Tests for Tate's algorithm: explicit LMFDB labels, prettified code, more tests, test all bad primes (and some good)
- Tests for (Non-archimedean) Local Heights: prettified code, more tests exercising different algorithm cases

I've also implemented `conductor` for elliptic curves over rational function fields, returning an array of `(prime, exponent)` tuples following Magma. It was implemented strictly for parity with ellcrv over QQ/NF. If we'd rather return a Divisor once/if rational-function-field divisors are in place, I can comment-out/revert the implementation + tests (~20 lines). Happy to go either way, but we need to make a decision.

No changes for existing callers. Public API additions: `EllipticCurveLocalData`, `EllipticCurveReduction` (used by `EllipticCurveLocalData`), and the `tates_algorithm_local` overload with `::Type{EllipticCurveLocalData}`. Exposing `EllipticCurveLocalData` rather than keeping it internal is intentional: a named struct is easier to use (and to extend) than a positional 5-tuple.


src/EllCrv/DiscreteValuation.jl
```diff
@@ -0,0 +1,185 @@
...
+# computes valuation of an element
+#
+# reduce(DiscreteValuation{T}, T)
+# maps an element of the field to the valuation's residue field
+#
+# lift(DiscreteValuation{T}, x)
+# lifts an element of the valuation's residue field to the field
+#
+# norm(DiscreteValuation{T}) -> ZZRingElem
+# norm of the valuation
```
thofma: Hm, the name is a bit confusing, since `x -> 2^v(x)` is a norm in the classical sense, if `v` is a discrete valuation. Maybe `residue_field_order`?

alexey-orlov-math: wouldnt norm of *element* be `c^v(x)`, or you mean like "sounds like returns a function"? Yeah, the idea was about a "norm of prime below" (which is exactly the order of residue field)
since the function name then becomes of the same length as impl, i guess i'll just kill it and use directly in local_height. Will push the change on the evening (with extra changes if needed from review)

thofma: I think I was just confused by "norm of a discrete valuation" (never read this phrase before). Thinking of it as a place, it sounds fine, since "norm of place" is the order of the corresponding residue field. So keeping `norm` like this is fine.

alexey-orlov-math: > I think I was just confused by "norm of a discrete valuation" (never read this phrase before). Thinking of it as a place, it sounds fine, since "norm of place" is the order of the corresponding residue field. So keeping `norm` like this is fine.

does it sound better if i update this to
```
# norm(DiscreteValuation{T}) -> ZZRingElem
# norm of the corresponding non-archimedean place
#
```

thofma: Looks good, thanks

alexey-orlov-math: tweaked head commit and force pushed
https://github.com/thofma/Hecke.jl/pull/2240/changes#diff-7f5f042b01d5e062d19597714861313d9b3911d7b4283b0572984d137a117914R34

src/EllCrv/DiscreteValuation.jl
```diff
@@ -0,0 +1,185 @@
...
+
+################################################################################
+#
+# Valuation on Rational Function Field k(t) (at the infinite place)
+#
+################################################################################
+
+struct RationalFunctionFieldDegreeValuation{T} <: DiscreteValuation{T}
+  K::AbstractAlgebra.Generic.RationalFunctionField
+  red_map
```
thofma: Any reason to not parametrize this?

src/EllCrv/Heights.jl
```diff
@@ -107,145 +107,64 @@ Computes the local height of a point $P$ on an elliptic curve defined over
...
   B = 2*y + a1*x + a3 # = psi2(P)
   C = 3*x^4 + b2 * x^3 + 3*b4*x^2 + 3*b6*x + b8 # = psi3(P)
 
-  # L always QQ
-  # N always ZZ
-  # M always QQ
+  # This is Proposition 3.4.1 in Cremona; the algorithm is coming from
+  #   Silverman's "Computing Heights on Elliptic Curves"
+  # NOTE: we adopt Cremona's normalization convention,
+  #   so our values are twice those from Silverman.
```
thofma, line 149: I remember that we/I had a bit of headache with the normalization in the past. Do you know how this compares to Magma/SageMath?

src/EllCrv/DiscreteValuation.jl
alexey-orlov-math: So i think the reason is similar to `RationalFunctionFieldValuation` where reduction map gets different types depending on the underlying field. In this case, though, the map type is more homogenous:
Q: `MapFromFunc{KInftyRing{QQFieldElem}, QQField}`
abs number field: `MapFromFunc{KInftyRing{AbsSimpleNumFieldElem}, AbsSimpleNumField}`
GF: `MapFromFunc{KInftyRing{FqFieldElem}, FqField}`
so we could parametrize on top by the residue field type, i guess? Any suggestions on making it more sane? 

Btw, same question for `RationalFunctionFieldValuation` but in there differences are bigger:
Q: `EuclideanRingResidueMap{RationalFunctionField{QQFieldElem, QQPolyRingElem}, EuclideanRingResidueField{RationalFunctionFieldElem{QQFieldElem, QQPolyRingElem}}}`
abs number field: `EuclideanRingResidueMap{RationalFunctionField{AbsSimpleNumFieldElem, Poly{AbsSimpleNumFieldElem}}, EuclideanRingResidueField{RationalFunctionFieldElem{AbsSimpleNumFieldElem, Poly{AbsSimpleNumFieldElem}}}}`
GF: `EuclideanRingResidueMap{RationalFunctionField{FqFieldElem, FqPolyRingElem}, EuclideanRingResidueField{RationalFunctionFieldElem{FqFieldElem, FqPolyRingElem}}}`

src/EllCrv/Heights.jl
alexey-orlov-math: so let me preface by saying that height output was NOT changed at all - this comment is here exactly to document the thing
all three systems agree (as in - using Cremona normalization), with the caveat of magma needing `Renormalization` parameter (as specified in docs), So (using our test curve). 
Magma:
```
E := EllipticCurve([1, -1, 1, -2063758701246626370773726978, 32838647793306133075103747085833809114881]);
P := E!([-30987785091199, 258909576181697016447]);
LocalHeight(P, 17: Renormalization := true);
```
gives `-1.88880889603747738683302307858`
Sage:
```
E = EllipticCurve([1, -1, 1, -2063758701246626370773726978, 32838647793306133075103747085833809114881])
P = E(-30987785091199, 258909576181697016447)
P.non_archimedean_local_height(17).n()
```
gives `-1.88880889603748` 

both agreeing with our value

src/EllCrv/DiscreteValuation.jl
thofma: I think we can just do
```
struct RationalFunctionFieldDegreeValuation{T, S} <: DiscreteValuation{T}
  K::AbstractAlgebra.Generic.RationalFunctionField
  red_map::S
```
and this should do the trick. Probably we also want to make the field for `K` concrete, or, it at least make `base_field` getter stable.

src/EllCrv/Heights.jl
thofma: Thanks for the explanation!

thofma, line 149:
```suggestion
  #   so our values are twice those from Silverman.
  # This is the same convention that SageMath and Magma (with `Renormalization := true`) use.
```

alexey-orlov-math: done
https://github.com/thofma/Hecke.jl/pull/2240/changes#diff-1fdd8d3204402da27b7856c0b091439593a86a64fdd4525e1154e43ba85230c1R150

src/EllCrv/DiscreteValuation.jl
alexey-orlov-math: https://github.com/thofma/Hecke.jl/pull/2240/changes/062f98fb5cad615719f23dab2d8be5239b104266

thofma:
Thanks. That cleanup was long overdue! Looks much better now.

PR #2241, opened 2026-04-20
Heuristic endomorphisms and Abel-Jacobi Map

JHanselman:
This adds code to heuristically compute homomorphisms between abelian varieties given by period matrices. It also improves the precision of the period matrix computations.

It additionally adds divisors and points on Riemann surfaces and the Abel-Jacobi map


simonbrandhorst (changes requested):
Would you please add documentation to the new methods?

PR #2252, opened 2026-04-23
Replaces old theta functions with theta functionality from flint

JHanselman:
Theta functions will be useful for testing the Riemann surfaces code (and for reconstructing curves from period matrices). This replaces my old theta functionality by the much faster flint implementation.


thofma (approved):

simonbrandhorst (approved):
Thanks for adding this functionality!

PR #2256, opened 2026-05-01
GenOrd: cleanup, fixes, type stability

alexey-orlov-math:
- Stricter typing: `KInftyRing{T}` and `KInftyElem{T}` are now `KInftyRing{T, U}`, and `KInftyElem{T, U}`, mirroring AbstractAlgebra's RationalFunctionField{T, U}. Also changed `KInftyRing` constructor to have cached as kwarg. 
- `degree(::GenOrdIdl)` now returns the inertia degree of the ideal (before it was `place` degree: for function field F/K with fields of constants K the place degree of P is [F_P:K]). Divisor code adjusted (divisors care about places).
- Generalization beyond function fields: `norm`, `minimum` and `prime_dec_nonindex` now work for any base ring.
- Fixed `prime_dec_nonindex` to generate normal two element representation. This also fixes issues with `factor`/`valuation` on totally inert primes.
- Ensured we get HNF matrix for principal ideal, thus we now have HNF matrices everywhere.
- Some fixes for type stability (most notably `factor` and `prime_decomposition`) and miscellaneous typos.
- Tests: removed root testset in function fields test (so that categories are visible), added per-field-type testsets to `GenOrdIdl` tests

If we care about somebody using `KInftyRing` and `KInftyElem`: the real breaking change is constructor (since it now demands two types), we can restore old constructor (with one type `T` and cached as positional argument); it was deemed unnecessary since this type is not used directly. 
Also, `degree(::GenOrdIdl)` is actual result change (i've also added explicit `inertia_degree` synonym), but it was deemed pure correction (Magma has degree == intertia_degree, and now our answers agree with Magma)

Known Issue: it turned out that `GenOrd` code cannot handle number fields with non-monic polynomials (Lenstra order is suborder of equation order); should be fixed separately if we really care about this.

### DIFF
--- a/src/FunField/Divisor.jl
+++ b/src/FunField/Divisor.jl
@@ -448,8 +448,27 @@ end
 Return the degree of D.
 """
 function degree(D::Divisor)
-  L = support(D)
-  return sum(degree(f)*e for (f, e) in L; init = zero(Int))::Int
+  # A divisor D = sum_P n_P * P has degree sum_P n_P * deg(P), where P ranges
+  #   over *places* of F/K (K the field of constants).
+  #
+  # For a place P lying over a (finite) prime p in the base ring,
+  #   deg(P) = [F_P : K] = f(P) * deg(p),
+  # where f(P) is the inertia degree and deg(p) is the degree of p as a
+  # polynomial in the base ring.
+  #
+  # For infinite places, p lies in KInftyRing.
+  # The formula above stays correct if we read deg(p) as the degree of p
+  # as a polynomial in the local parameter t = 1/x.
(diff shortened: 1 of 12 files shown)


src/FunField/Divisor.jl
```diff
@@ -448,8 +454,27 @@ end
 Return the degree of D.
 """
 function degree(D::Divisor)
-  L = support(D)
-  return sum(degree(f)*e for (f, e) in L; init = zero(Int))::Int
+  # A divisor D = sum_P n_P * P has degree sum_P n_P * deg(P), where P ranges
```
alexey-orlov-math, line 457: @JHanselman divisor's place-vs-ideal degree

PR #2258, opened 2026-05-01
fix: isometry of `TorQuadModule` as bilinear modules

StevellM:
Somehow, the `gram_matrix_bilinear` of the normal form of a quadratic module might not agree with the `gram_matrix_bilinear` of the normal form of the associated bilinear module. I had an example of `TorQuadModule` which are isometric as bilinear modules but the function said it was not. So I have added the option to create a "normal form" by only working with the `gram_matrix_bilinear` if we are only interested in the bilinear form.

### DIFF
--- a/src/QuadForm/Torsion.jl
+++ b/src/QuadForm/Torsion.jl
@@ -97,6 +97,7 @@ function torsion_quadratic_module(M::ZZLat, N::ZZLat; gens::Union{Nothing, Vecto
   T.value_module = QmodnZ(_modulus)
   T.value_module_qf = QmodnZ(_modulus_qf)
   T.is_normal = false
+  T.is_normal_bilinear = false
   return T
 end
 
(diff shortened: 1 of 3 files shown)


simonbrandhorst:
I don't understand what is going on. Maybe it is a misconception on our end. Can you give a minimal example? I would like to understand more.

src/QuadForm/Torsion.jl
simonbrandhorst, line 1758: is there any difference between the if and the else part?

StevellM:
In the test file, I have added a test which on the current Hecke branch does not work while it should. 
```
 T1 = torsion_quadratic_module(QQ[4//5 0; 0 1//5])
  T2 = torsion_quadratic_module(QQ[2//5 0; 0 2//5])
 is_isometric_with_isometry(T1, T2; as_bilinear_module=true)
```
The problem being that if we regard a `TorQuadModule` as a quadratic module or as a bilinear module, you will not necessarily get the same `gram_matrix_bilinear` for the normal form. Hence we need a `normal_form` function for torsion quadratic modules where we disregard the quadratic form and just focus on the bilinear one.

EDIT: when I wrote the first version of the code for isometry between quadratic modules seen as bilinear modules, I did not check that `normal_form` could give different `gram_matrix_bilinear` depending if the modulus quadratic is the same or different to the modulus bilinear. Maybe it is an expected feature. So the current code just provide a way to only produce a normal form as if the moduli were equal.

src/QuadForm/Torsion.jl
StevellM: Yes, if you have a `TorQuadModule` whose moduli (bilinear and quadratic) are different but you want to treat as a finite bilinear module, then you want to use the `gram_matrix_bilinear` and the `modulus_bilinear_form`. So here we want to use `modulus_quadratic_form` only if we care about the quadratic form and $p$ is 2.

simonbrandhorst (approved):

simonbrandhorst:
thx

PR #2265, opened 2026-05-10
Ellcrv: min model/discriminant for rational function fields

alexey-orlov-math:
## Additions:

- added `_bad_prime_candidates(E)`: internal helper returning the primes dividing the discriminant of the supplied Weierstrass model (the old `bad_primes` behavior). It is still used internally both as candidates for `bad_primes` and when we need "bad" primes for a given model, for example `_global_minimality_obstruction_ideal_integral`.
- added per-curve cache of `EllipticCurveLocalData` on `EllipticCurve` (via `@attributes`). Data is cached **only** for primes of bad reduction.
- added `bad_primes(E)` for rational function fields. 
- added `minimal_model(E, p)` for rational function fields, and unified dispatch across everything we care about
- added `minimal_model(E)` for rational function fields: finite-place minimal model
- added `minimal_discriminant(E)` for rational function fields: returns the divisor as `Vector{Tuple{T, ZZRingElem}}` of `(place, valuation)` pairs
- added `valuation(x::T, p::T)` for `RationalFunctionFieldElem` (we were missing it, yet needed to copy the logic of unpacking the element into either prime poly or 1/x)
- added `EllipticCurveLocalData.discriminant_valuation`: Tate's algorithm already computes it

## Changes:

- `bad_primes(E)` is now strict across all base fields (QQ, NF, RFF): the result is the support of the minimal discriminant, i.e. the primes of bad reduction of the *curve* (matching Magma's `BadPrimes` and `BadPlaces`). The result is now cached.
- `minimal_discriminant` for QQ now returns `ZZRingElem` (was `QQFieldElem`), docs updated to explicitly mention that the positive generator is returned (this was already positive, just docs updated); also agrees with pari's `ellminimaldisc`
- all of `bad_primes`, `minimal_model`, `minimal_discriminant`, `has_global_minimal_model`, `semi_global_minimal_model`, `global_minimality_class` now accept non-integral models for QQ/NF
- similar to `bad_primes` for curves over `QQ`, the output for curves over number fields and rational function fields is also sorted (the sort key is partial — primes that tie on it appear in implementation-defined order). Further, since we now cache them, the order is stable within a session.

## Fixes:

- fixed `minimal_model` over number fields with class number 1 returning a non-minimal model: it was completely ignoring obstruction ideal, allowing discriminant inflation. This fixes `@test_broken` in `test/EllCrv/MinimalModels.jl`

## Tests:

- `@test_broken` is now `@test`
- all the changes covered with extra tests and existing tests for QQ/NF are expanded 
- cleaned up both `test/EllCrv/MinimalModels.jl` and `test/EllCrv/LocalData.jl` to consistently use `cached = false` kwarg, and prefer symbol to string for var names


thofma:
Hm, thanks for spotting the problem with `bad_primes`. It is unfortunate that our `bad_primes` deviates from Magma. Also, as you noticed, our `bad_primes` do not return the list of bad primes, but possible a strict superset.

I am leaning towards changing it to return the primes of bad reduction and would consider the current behaviour slightly buggy. Maybe @JHanselman or @simonbrandhorst may have an opinion?

JHanselman:
I think I noticed that behavior before, but figured it was a design choice. I do think we need a function that returns the primes of bad reduction if we don’t have it yet.

Creating this function and naming it something like bad_reduction_primes or changing bad_primes to return the primes of bad reduction both seem valid to me.

If we do change bad_primes we need to be careful that it doesn’t break any other functions that could potentially rely on its current behavior.

alexey-orlov-math:
> I think I noticed that behavior before, but figured it was a design choice. I do think we need a function that returns the primes of bad reduction if we don’t have it yet.
> 
> Creating this function and naming it something like bad_reduction_primes or changing bad_primes to return the primes of bad reduction both seem valid to me.

Ok, i will do this in the same pr. I will go through the usages and will check if there are places where indeed we do not care about guaranteed bad (we need to create local minimal model for it, so some extra work).

thofma:
I guess we will still need a name to get the divisors/places dividing the discriminant of the model? Probably only needed internally.

alexey-orlov-math:
> I guess we will still need a name to get the divisors/places dividing the discriminant of the model? Probably only needed internally.

Most of the times we use `bad_primes` ourselves we are feeding it to "let's do some tate's algorithm on these". The only exception seem to be minimal models machinery. So while keeping internal might make sense, we shouldn't worry about name too much, i would say; so `_bad_prime_candidates` should do the trick

alexey-orlov-math:
bad_primes + local data caching:
https://github.com/thofma/Hecke.jl/pull/2265/changes/21a20a04a1d5c3b5f3439808052bba3c4cf0c5ab
will update pr desc

src/EllCrv/LocalData.jl
```diff
@@ -658,42 +670,111 @@ valuation at that place. Finite places are represented by irreducible
...
+    e > 0 && push!(result, (p, e))
   end
-  e_inf = _tates_algorithm(E, degree).conductor_valuation
-  e_inf > 0 && push!(result, (1//gen(K), e_inf))
 
   return result
 end
 
-#Magma returns the primes that divide the minimal discriminant
 @doc raw"""
```
thofma, line 683: Either the github editor is acting up, or the docstring is not attached to `bad_primes` anymore.

alexey-orlov-math: https://github.com/thofma/Hecke.jl/pull/2265/changes#diff-d34e0ed3f19a2a3c6041b48fde1b02105e57c2212f3d637fd833f1c869791e6fR686

https://github.com/thofma/Hecke.jl/pull/2265/changes#diff-d34e0ed3f19a2a3c6041b48fde1b02105e57c2212f3d637fd833f1c869791e6fR710

https://github.com/thofma/Hecke.jl/pull/2265/changes#diff-d34e0ed3f19a2a3c6041b48fde1b02105e57c2212f3d637fd833f1c869791e6fR741

thofma (approved):

Issue #2266, opened 2026-05-12
Issue with ideal factorization in function fields with constant basefield QQBar


stefnomden:
I’m running into an issue with function fields over `QQBar`  and I’m wondering whether this is a bug or an unimplemented feature. When I run the following code I get an error but not when I set `K = QQ` or `K` equal to some other number field.
```
K = QQBar
Kx, x = rational_function_field(K,"x")
KxY, Y = polynomial_ring(Kx, "Y")
P = Y^2 - x^3 - x^2
kC, y = function_field(P, "y")
OC = finite_maximal_order(kC)
I = ideal(OC(x),OC(y))
factor(I) #gives error
```
The error I get is
```
MethodError: no method matching _radical_zero(::StructureConstantAlgebra{EuclideanRingResidueFieldElem{AbstractAlgebra.Generic.Poly{QQBarFieldElem}}})
The function `_radical_zero` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  _radical_zero(::Hecke.AbstractAssociativeAlgebra{T}) where T<:Union{QQFieldElem, NumFieldElem}
   @ Hecke ~/Hecke.jl/src/AlgAss/radical.jl:55

Stacktrace:
  [1] _radical(A::StructureConstantAlgebra{EuclideanRingResidueFieldElem{AbstractAlgebra.Generic.Poly{QQBarFieldElem}}})
    @ Hecke ~/Hecke.jl/src/AlgAss/radical.jl:39
  [2] _issemisimple
    @ ~/Hecke.jl/src/AlgAss/AbsAlgAss.jl:984 [inlined]
  [3] _decompose(A::StructureConstantAlgebra{EuclideanRingResidueFieldElem{AbstractAlgebra.Generic.Poly{QQBarFieldElem}}})
    @ Hecke ~/Hecke.jl/src/AlgAss/Decompose.jl:88
  [4] decompose
    @ ~/Hecke.jl/src/AlgAss/Decompose.jl:48 [inlined]
[...]

PR #2268, opened 2026-05-15
fix: radical for algebras and residue fields for QQbar

thofma:
- Relax signature to allow QQBar in radical computation
- Make the residue_field of QQBar[x] equal to QQBar

Fixes the remaining issues of #2266.

@alexey-orlov-math does the fix with the residue field make sense to you? Otherwise we would have to implement `factor` for R[X], where `R::EuclideanRingResidueField{AbstractAlgebra.Generic.Poly{QQBarFieldElem}}`.


alexey-orlov-math:
> @alexey-orlov-math does the fix with the residue field make sense to you? Otherwise we would have to implement `factor` for R[X], where `R::EuclideanRingResidueField{AbstractAlgebra.Generic.Poly{QQBarFieldElem}}`.

yeah changes make sense

PR #2270, opened 2026-05-17
EllCrv: order in characteristic 2 and 3 for arbitrary degree

alexey-orlov-math:
Nemo 0.55.1 includes the possibility (flint 3.5.0) to construct qadic field with custom modulus. We can now enable p-adic methods to compute curve order for any base field degree. 
Fixed previously missed Int64 truncation (we never hit it since we were limiting the degree).
Added more tests, prettified existing (mostly precision kwarg handling and field caching), and added benchmark for AGM2


thofma:
Nice, thanks

PR #2275, opened 2026-06-02
FunField/GenOrd: adapt to stricter FunctionField{T, U} typing in AbstractAlgebra

alexey-orlov-math:


thofma:
can you update the AA/Nemo version in this pr here?

PR #2281, opened 2026-06-10
FunField: Riemann-Roch test coverage, plus fixes found along the way

alexey-orlov-math:
# Context

This is step 0 towards optimized Riemann-Roch computation. Adds an extensive test suite for Riemann-Roch and fixes the bugs and inefficiencies it uncovered in FunField/Divisor and GenOrd.

The next steps would be
- Implement proper (modular) HNF for function field ideals
- Implement Riemann-Roch via divisor reduction

Originally we added test suite and implemented Riemann-Roch only to find that there were no performance gains, but only because everything else is very slow. This PR is literally "make sure the tests we added run in sane amount of time" (plus some bug fixes) before embarking on bigger optimizations

# Tests

- New `test/FunField/RiemannRoch.jl`: verifies the Riemann-Roch theorem, genus, and canonical/different divisors across elliptic (F_5), hyperelliptic (F_3, F_5), non-monic, Artin-Schreier (F_2, F_3, F_5), and superelliptic (Q, F_4) curves.
- Expanded `GenOrd`/`Ideal` tests: fractional ideal inversion, norm/minimum, prime decomposition, containment, colon, and scaling by base-field elements.

# Bug fixes

- `prime_decomposition`: gate the non-index fast path on `is_defining_polynomial_nice` — the index is ill-defined for non-monic/non-integral defining polynomials (mirrors the number field convention).
- `zero_divisor`/`pole_divisor` of a field element: the integral_split-based construction was wrong; now derived from divisor(f) via lcm with the trivial divisor.
- `in(a, O)` for field elements called data on a FieldElem; membership now works for both field and order elements.
- `is_prime`: handle prime ideals without a two-element representation instead of accessing an undefined gen_two.

# Performance

- `degree(D)`: computed from ideal norms when the support is unknown, avoiding a forced factorization.
- `is_effective(D)`: reduced to integrality checks on the underlying fractional ideals (no factorization).
- `genus`: rewritten via discriminant degrees of the finite/infinite maximal orders instead of `basis_of_differentials`; cached with @attr, as are `different_divisor` and `canonical_divisor`.
- Fractional ideal `inv`: factors the scalar denominator out instead of routing through colon.

We note again that faster HNF (for function fields) is a goal of a subsequent PRs. 

# Minor

- Multiplication of `GenOrdIdl`/`GenOrdFracIdl` by rational function field elements (always returning a fractional ideal, for type stability).
- Type assertions in `basis_matrix(::GenOrdFracIdl)` for inference; @req same-order checks on ideal arithmetic.

### DIFF
--- a/src/GenOrd/Ideal.jl
+++ b/src/GenOrd/Ideal.jl
@@ -652,14 +652,26 @@ function Hecke.factor(A::GenOrdIdl{S, T}) where {S, T}
 end
 
 function prime_decomposition(O::GenOrd{S, T}, p::RingElem, degree_limit::Int = degree(O), lower_limit::Int = 0; cached::Bool = true) where {S, T}
-  #Index not well-defined for infinite maximal order
-  if !isa(base_ring(O), KInftyRing) && !(divides(index(O), p)[1])
+  # Fast path needs:
+  # 1. well-defined equation order: we need defining polynomial to be "nice",
+  #    that is, monic and integral
+  # 2. finite maximal order setting, since index is ill-defined for KInftyRing
+  # 3. p coprime to index
+  # Note: for non-nice polynomials the index has a denominator.
+  #       We could check divisibility by p of both numerator and denominator,
+  #       but that is a bit hacky, so we mimic number fields and instead
+  #       check that the defining polynomial is nice.
+  if !isa(base_ring(O), KInftyRing) &&
+     is_defining_polynomial_nice(O.F) &&
+     !(divides(index(O), p)[1])
     return prime_dec_nonindex(O, p, degree_limit, lower_limit)
   else
     return prime_dec_gen(O, p, degree_limit, lower_limit)
   end
 end
 
+# TODO: prime_dec_gen currently does NOT set gen_two
+# TODO: we should mimic number fields by adding a search for a uniformizer
 function prime_dec_gen(O::GenOrd{S, T}, p::RingElem, degree_limit::Int = degree(O), lower_limit::Int = 0) where {S, T}
   Ip = pradical(O, p)
   lp = _decomposition(O, ideal(O, p), Ip, ideal(O, one(O)), p)
@@ -895,7 +907,17 @@ function is_prime(A::GenOrdIdl)
     if norm(A) != norm(P)
       continue
     end
-    if P.gen_two in A
+
+    # from factorization we know already that A is in P
+    # TODO: for non-maximal orders we need to check both generators
+    # TODO: though most of our code around primes assumes maximal order
+    PinA = if has_2_elem(P)
+      P.gen_two in A
+    else
+      A == P
+    end
+
+    if PinA
       A.is_prime = 1
       A.splitting_type = P.splitting_type
       return true
--- a/src/GenOrd/Ideal.jl
+++ b/src/GenOrd/Ideal.jl
@@ -287,6 +287,8 @@ function Base.isequal(a::GenOrdIdl{S, T}, b::GenOrdIdl{S, T}) where {S, T}
 end
 
 function Base.:(*)(a::GenOrdIdl{S, T}, b::GenOrdIdl{S, T}) where {S, T}
+  @req order(a) === order(b) "Ideals must have same order"
+
   O = order(a)
   Ma = basis_matrix(a)
   Mb = basis_matrix(b)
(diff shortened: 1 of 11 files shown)


thofma:
@fieker do you want to have a look?

src/GenOrd/Ideal.jl
```diff
@@ -652,14 +652,26 @@ function Hecke.factor(A::GenOrdIdl{S, T}) where {S, T}
 end
 
 function prime_decomposition(O::GenOrd{S, T}, p::RingElem, degree_limit::Int = degree(O), lower_limit::Int = 0; cached::Bool = true) where {S, T}
-  #Index not well-defined for infinite maximal order
-  if !isa(base_ring(O), KInftyRing) && !(divides(index(O), p)[1])
+  # Fast path needs:
+  # 1. well-defined equation order: we need defining polynomial to be "nice",
+  #    that is, monic and integral
+  # 2. finite maximal order setting, since index is ill-defined for KInftyRing
```
fieker, line 658: I've read this now a couple of time, but why is the index ill-defined in the KInftyRing?
The index is always between two orders. If the poly is nice, then the equation order is an order and usually "the" second order. But what is the connection to KInfty? In the future, class field theory, using Artin-Schreier equations, the polynomials will never (almost) never be nice in the finite case, but frequently in the infinite case. I think the "correct" concept is to check if the monic minpoly is integral in the current order. Don't know if we have that (yet)

fieker:
I am happy with the progress, it moves in the right direction and should be merged. I still stand by my claim though that the divisors need a re-design: we really should store the inverses of the ideals and optionally a factored element...

src/GenOrd/Ideal.jl
alexey-orlov-math: > but why is the index ill-defined in the KInftyRing?

i will be completely honest: i don't know - it is just the rephrase of the existing comment 
> #Index not well-defined for infinite maximal order

i will look into it later (i plan to revisit to make sure that index-divisor code path sets second generator: this is also important to optimize the case of constructing divisors from primes where we can do HNF on smaller 2n x n matrix instead of n^2 x n)

alexey-orlov-math:
> I still stand by my claim though that the divisors need a re-design: we really should store the inverses of the ideals and optionally a factored element...

I wholeheartedly agree. I have TODO item to, at least, make sure we support support/ideal "in parallel" (that is, some operation are easier to do with support if we have it already, and materialize ideals only when needed). In similar vein, i want to investigate the possibility of NOT enforcing HNF: e.g. for polynomial matrices we could do a lot with weak popov and compute hnf only when we need an actual basis. I guess i will add "store inverses" to "need research" list ;-)

src/GenOrd/Ideal.jl
alexey-orlov-math: > I've read this now a couple of time, but why is the index ill-defined in the KInftyRing? 

@fieker I think i got the main idea behind this. We compute index as 
```
function Hecke.index(O::GenOrd)
  return is_equation_order(O) ? O.R(1) : O.R(det(basis_matrix_inverse(O)))
end
```
which works in finite case when poly is monic, because `y` is integral and this is indeed the matrix in power basis (this is indeed the restatement of what you have said)

But when we speak about infinity - `y` is almost never integral so we need to tweak the computation. Case in point: for
```
kx, x = rational_function_field(GF(3), :x; cached = false)
ky, y = polynomial_ring(kx, :y; cached = false)
# has no finite affine points
F, a = function_field(y^2 - x^5 - 2*x^4 - 2*x^3 - x^2 - 2; cached = false)
Ofin = @inferred finite_maximal_order(F)
Oinf = @inferred infinite_maximal_order(F)
```
the basis matrix of Oinf is `(1, 0; 0, 1/x^3)` and det of inverse is `x^3` - not in `KInftyRing`. But this is not the matrix in power basis, because it uses y which is not integral.

I hope my mumbling made some sense 8). I will look into this, yes

fieker:
On Sun, Jun 21, 2026 at 12:18:04AM -0700, alexey-orlov-math wrote:
> 
> 
> 
> > @@ -652,14 +652,26 @@ function Hecke.factor(A::GenOrdIdl{S, T}) where {S, T}
>  end
>  
>  function prime_decomposition(O::GenOrd{S, T}, p::RingElem, degree_limit::Int = degree(O), lower_limit::Int = 0; cached::Bool = true) where {S, T}
> -  #Index not well-defined for infinite maximal order
> -  if !isa(base_ring(O), KInftyRing) && !(divides(index(O), p)[1])
> +  # Fast path needs:
> +  # 1. well-defined equation order: we need defining polynomial to be "nice",
> +  #    that is, monic and integral
> +  # 2. finite maximal order setting, since index is ill-defined for KInftyRing
> 
> > I've read this now a couple of time, but why is the index ill-defined in the KInftyRing? 
> 
> @fieker I think i got the main idea behind this. We compute index as 
> ```
> function Hecke.index(O::GenOrd)
>   return is_equation_order(O) ? O.R(1) : O.R(det(basis_matrix_inverse(O)))
> end
> ```
> which works in finite case when poly is monic, because `y` is integral and this is indeed the matrix in power basis (this is indeed the restatement of what you have said)
> 
> But when we speak about infinity - `y` is almost never integral so we need to tweak the computation. Case in point: for
> ```
> kx, x = rational_function_field(GF(3), :x; cached = false)
> ky, y = polynomial_ring(kx, :y; cached = false)
> # has no finite affine points
> F, a = function_field(y^2 - x^5 - 2*x^4 - 2*x^3 - x^2 - 2; cached = false)
> Ofin = @inferred finite_maximal_order(F)
> Oinf = @inferred infinite_maximal_order(F)
> ```
> the basis matrix of Oinf is `(1, 0; 0, 1/x^3)` and det of inverse is x^3 - not in `KInftyRing`. But this is not the matrix in power basis, because it uses y which is not integral.
> 
> I hope my mumbling made some sense 8). I will look into this, yes
> 
All of this is correct and makes sense. My point was/is slightly
different and possibly pointless as I am not quite aware of our
conventions wrt integrallity and monicness of polynomials:
  y^p-y+a(x)
in char p defines a cyclic extension of degree p. It is at most ramified
at places Q where v_Q(a(x)) < 0. In particular if y is integral, then
it is unramified.
The class field theory will use those polynomials. Here, a(x) will
almost never be k[x]-integral, but frequently could by
k[1/x]-integral...

PR #2285, opened 2026-06-20
FunField: some optimization related to basis_of_differential

alexey-orlov-math:
# Context

We are continuing the quest of speeding up the `basis_of_differentials`. 
Alas, currently we cannot speed up the `Q(x)` case. We will research the normalized 2-element representation of ideals, but this is out of scope for this PR.
This caveat aside, for function fields over `F_q(x)` the Riemann-Roch machinery itself takes quite minuscule proportion of the running time of `basis_of_differentials`, so we have embarked on fixing the issues around (that is - general inefficencies in Divisor/GenOrd code).

# Changes

The notable changes:

- Enforce the `basis_matrix` of `GenOrdIdl` to be the canonical lower-left HNF.
- Added lower-left modular HNF for ideal operations. Since we now gurantee the basis matrix to be in lower-left HNF, computing the largest elementary divisor is robust and we can use it for HNF operations. This optimization is specifically disabled for the case of base field `Q(x)`: in here the coefficient swell is the main performance issue, and adding extra `gcdx` and `mod` operations actively harms performance. For this case we use `hnf_kb!` from AbstractAlgebra.

Smaller changes, which are all trivial by themselves but bring real savings both in terms of memory and time:

- cache representation matrices of basis elements on `GenOrd`: this allows better `representation_matrix` for order element, but more importantly more efficient construction of trace matrix and ideal multiplication matrix stack
- specialized mutliplication of ideal by a base-ring scalar: in this case we can scale the basis matrix directly
- added zero/unit short-circuits for ideals `+`/`-`/`intersect` and ideal `divexact`
- surprisingly check for the function field to be separable is generating a lot of garbage: cached on FunctionField. We use this check in several places where we assume separability; since we have not yet decided how we will move forward in regards to non-separable defining polynomials, it is better to keep the check but at least make it more efficient.
- optimized `differential` to skip the second (y-derivative) term when the element has no actual y-dependence
- used `pseudo_inv` in `codifferent` to invert trace matrix (since it is dense, way more efficient than usual row reduction in inv)

# Why bother

We are using the polynomial from Jeroen Hanselman to construct function field
```
f = 18*x^4*y^9 - 4*x^4*y^8 - 20*x^4*y^7 - 2*x^4*y^6 + 7*x^4*y^5 - 6*x^4*y^4 + 8*x^4*y^2 + 12*x^4*y - 19*x^4 - 17*x^3*y^9 + 8*x^3*y^8 - 20*x^3*y^7 + 13*x^3*y^6 + 13*x^3*y^5 + 6*x^3*y^4 + 14*x^3*y^3 + 2*x^3*y^2 + 5*x^3*y + 18*x^3 + 9*x^2*y^9 + 18*x^2*y^8 - x^2*y^6 + 6*x^2*y^5 + 8*x^2*y^4 + 6*x^2*y^3 - 2*x^2*y^2 + 5*x^2*y + 10*x^2 + 18*x*y^9 + 3*x*y^8 + 5*x*y^7 - 16*x*y^6 + 9*x*y^5 + 7*x*y^4 + 7*x*y^3 + 14*x*y^2 - 14*x*y + 14*x - 19*y^9 - 9*y^8 - 18*y^7 - 12*y^6 + 8*y^5 - 18*y^4 - 15*y^3 + 17*y^2 + 5*y - 2;
```
and time the second call to `basis_of_differentials(F)` after running `GC.gc()`.

For a function field over `GF(10000000000000000051)`, 
before
```
julia> @time basis_of_differentials(F);
  1.050605 seconds (18.84 M allocations: 1.578 GiB, 39.03% gc time)
```
after
```
julia> @time basis_of_differentials(F);
  0.052124 seconds (1.64 M allocations: 130.196 MiB)
```

For a function field over `finite_field(13, 5)`,
before
```
julia> @time basis_of_differentials(F);
  5.783126 seconds (125.10 M allocations: 10.567 GiB, 22.64% gc time)
```
after
```
julia> @time basis_of_differentials(F);
  0.534014 seconds (9.63 M allocations: 865.026 MiB, 28.14% gc time)
```

EDIT: while implementing two-elem normal form (of course on top of these changes) i have found a corner case issue with modular HNF. History rewritten, regression tests added


alexey-orlov-math:
tagging @fieker and @thofma for review

src/FunField/Factor.jl
```diff
@@ -106,12 +102,6 @@ end
 
 # squarefree factorization a la Trager-Gianni
 
-function is_separable(f::PolyRingElem{<:FieldElem})
```
thofma, line 109: What happend to this function?

thofma (approved):
Looks good, thanks

src/FunField/Factor.jl
alexey-orlov-math: it was (internal to module) copy of `AbtsractAlgebra.is_separable`. All the uses were changed to `Hecke._is_separable` with field argument

src/GenOrd/GenOrd.jl
```diff
@@ -869,6 +951,41 @@ end
...
+# Keeps elimination in base ring! This is especially important for function fields.
+# NOTE: for finite order all the denominators are 1, so the function could be simplified
+#   but this is not true for infinite order, so we keep the trivial lcm gathering
+# NOTE: this is NOT a replacement for general matrix inv
+# Trace matrix is dense with polynomial entries of moderate degrees, so normal inv
+#   will spend a lot of time doing divexact bringing to reduced row-echelon
+# But, for example, basis_matrix or inv used in colon have a triangular matrix
+#   and in this case general matrix inv is immediate (back substitution essentially)
+#   while pseudo_inv needs to do some work.
+function _fraction_free_inv(T::MatElem)
```
fieker, line 972: for which rings is this better? Should this be hooked up with the normal inv? So more benefits?

src/GenOrd/Ideal.jl
```diff
@@ -5,16 +5,29 @@
 ################################################################################
 
 ideal(O::GenOrd, a::RingElement, b::RingElement) = GenOrdIdl(O, a, b)
-
 ideal(O::GenOrd, a::RingElement) = GenOrdIdl(O, a)
 
-function ideal(O::GenOrd, M::MatElem)
-  return GenOrdIdl(O, hnf(M, :lowerleft))
+function ideal(O::GenOrd, M::MatElem; M_in_hnf::Bool = false)
```
fieker, line 10: I don't like the name of the kw arg as my matrices are not called M...maybe just in_hnf?

src/GenOrd/Ideal.jl
```diff
@@ -105,6 +117,168 @@ function show(io::IO, id::GenOrdIdl)
...
+        M[kr, jc] = mod(t1, g)
+      end
+    end
+  end
+
+  return M
+end
+
+# Specialization for Q(x)
+# The main issue here is coefficient swell, not degree; so by itself hnf_modular_eldiv_left! cannot help
```
fieker: have you tried different normalizations for element in Q(x)? integral primitive denominator, monic, normalize last non zero coeff...?

src/GenOrd/Ideal.jl
```diff
@@ -387,19 +607,24 @@ end
...
+
   O = order(A)
   b = _make_canonical_in(O, b)
-  bm = divexact(basis_matrix(A), b)
-  B = GenOrdIdl(O, bm)
+  is_unit(b) && return A
+
+  bm = divexact(basis_matrix(A; copy = false), b)
+  B = ideal(O, bm; M_in_hnf = true)
+
```
fieker, line 680: why do we keep the block with if false?

src/GenOrd/Ideal.jl
```diff
@@ -426,10 +651,10 @@ function assure_has_minimum(A::GenOrdIdl)
 
   O = order(A)
 
-  if isone(basis(O, copy = false)[1])
-    A.minimum = deepcopy(basis_matrix(A, copy = false)[1, 1])
+  if isone(basis(O; copy = false)[1])
+    A.minimum = deepcopy(basis_matrix(A; copy = false)[1, 1])
```
fieker, line 713: this is where we need the correct hnf! or a different method

src/GenOrd/Types.jl
```diff
@@ -166,11 +162,10 @@ end
 
   function GenOrdIdl(O::GenOrd, T::Vector{<:GenOrdElem})
     @assert all(x -> parent(x) === O, T)
-    # One should do this block by block instead of the big matrix
+
+    # TODO: One should do this block by block instead of the big matrix
```
fieker, line 166: don't be too sure. All in one might be better as the algorithm can make choices (eg if one of the matrices is diagonal...)

fieker (approved):
in general I am happy. I commented at a few places, but nothing serious. Thanks

src/GenOrd/GenOrd.jl
alexey-orlov-math: > for which rings is this better?

i tested for function fields only, actually. I'll recheck with more and specialize when/if needed

> Should this be hooked up with the normal inv?

no, i tried to explain in a comment - we use inv in 3 places essentially - here, colon, and inverse of basis matrix. Now, `colon`/`basis_matrix_inverse` have essentially HNF as input so doing inv here is "free" - since matrices are over field, row-reduced echelon form is used which essentially does nothing. But for `trace_matrix` it is different: since it is dense, we win here by the virtue of doing arithmetic in "integers" and not fractions. 
Actually we already have this in AbstractAlgebra (src/Solve.jl) 
```
matrix_normal_form_type(::Field) = RREFTrait()

# The fflu approach is the fastest over a fraction field (see benchmarks on PR 661)
matrix_normal_form_type(::FracField) = FFLUTrait()

matrix_normal_form_type(::FracField{T}) where {T <: PolyRingElem} = MatrixInterpolateTrait()
```
so i wonder if we can go even faster. Anyway, i'll recheck function fields over Q(x) and nf/localization to decide on specialization

src/GenOrd/Ideal.jl
alexey-orlov-math: this was copied from nford ideals (same in `AlgAssAbsOrdIdl`) - i prefered to have exactly same keyword as existing primitives

src/GenOrd/Ideal.jl
alexey-orlov-math: No, i would rather first try two-element representation before trying to fine-tune things

fieker:
I agree with keeping it the same. I still don't like the name, even though
it might have been mine

src/GenOrd/Ideal.jl
alexey-orlov-math: yes, part of this pr is exactly to ensure that matrices are all in lower-left HNF. So this now works always

src/GenOrd/GenOrd.jl
alexey-orlov-math: Ok i did a quick benchmark. For number fields (and localizations) `inv` wins slightly (well it is all in microseconds) because inv of matrix over Q goes into flint, which does fraction-free inverse up to some matrix size (and yeah - you won't beat flint here)
For both function fields over F_q(x) and Q(x) `_fraction_free_inv` is significantly faster (with gains growing with degree). As mentioned above this is because we stay in k[x] instead of k(x).
I will specialize the implementation

alexey-orlov-math: https://github.com/thofma/Hecke.jl/pull/2285/changes/ef81bf45ba6617d0cd2b243931b2b2b39ac2a67b

thofma:
@fieker please merge if your are happy

src/GenOrd/Ideal.jl
thofma: do you want to adress this @alexey-orlov-math or should we just merge and clean it up afterwards?

alexey-orlov-math: @thofma i would say let's do a cleanup of the existing code separately from doing new "features"/optimizations 8)

alexey-orlov-math:
i have found a corner case with modular hnf - moving to draft until i make a fix
EDIT: fixed force pushed

src/GenOrd/Ideal.jl
alexey-orlov-math: as i am doing updates - i will delete if false branch too

PR #2288, opened 2026-06-22
feat:Conic curves and twists/reconstruction of genus 2 curves

JHanselman:
**Adds conic curves:**

- Finding a minimal model of a conic over QQ
- Finding a rational point on a conic over number fields, finite fields and function fields
- Parametrisation of conics

**Implements the reconstruction and twists of genus 2 curves based on the hyperelliptic package by Lercier, Ritzenthaler and Sijsling**
- Reconstruct genus 2 curves from their Igusa invariants or G2 invariants (over all fields)
- Compute all twists of a given genus 2 curve (finite fields)


src/ConCrv/ConCrv.jl
```diff
@@ -0,0 +1,881 @@
...
+          x = divexact(x, c)
+          y = divexact(y, c)
+          z = divexact(z, c)
+        end
+
+        if T <: AbstractAlgebra.Generic.FunctionFieldElem
+          a = gen(K)
+          x = x(a)
+          y = y(a)
+          z = z(a)
```
thofma: I don't quite understand why we are doing this.

src/ConCrv/ConCrv.jl
```diff
@@ -0,0 +1,881 @@
...
+  B, U, sol = lll_gram_indef_isotropic(M; base = true)
+
+  W = W * transpose(U)
+  W = W/content(W)
+  #Return minimized conic Cmin, map from Cmin -> C, and corresponding matrix.
+  return conic_curve(K, B), W * [x,y,z], W 
+
+end
+
+function algorithm22(M::ZZMatrix, p::ZZRingElem)
```
thofma, line 671: Can you mention the reference in a comment?

thofma (commented):
I can't comment on the mathematics, but looks good overall. I left two comments.

src/ConCrv/ConCrv.jl
JHanselman: I think I needed to do that because numerator turns x, y and z into elements of a polynomial ring. And I want them to be elements of the function field. If I leave this part out the coercing into K at lines 92-94 will give an error.

The error gets triggered when working over a field extension over a rational function field. For example in 
the test over F, t = rational_function_field(GF(7), :t) in G2Rec.jl
(The reconstruction itself is not possible over GF(7)(t), so at some point during the algorithm we move to a quadratic extension of GF(7)(t)).

But maybe there is a better way to deal with it?

PR #2302, opened 2026-07-10
Two-element normal presentations for GenOrd ideals

alexey-orlov-math:
Similar to NfOrd ideals, GenOrd ideals now can carry a P-normal two-element presentation <`gen_one`, `gen_two`> with normal prime set defined by `gens_normal`. This is used for fast ideal arithmetic when the order is known maximal. 

### Main Changes
- Added Order maximality tracking, similar to NfOrd: `is_maximal::Int` field on GenOrd (0 unknown / 1 maximal / 2 not), set by maximal_order and integral_closure, and queried via `is_maximal(O)` (might compute maximal order and compare) and `is_maximal_known_and_maximal(O)`.
- Prime decomposition, Dedekind-Kummer path: extended to locally-nice polynomials (monic, denominators coprime to `p`) and infinite orders. 
- Prime decomposition, general case: use idempotent from the structure-constant split to find uniformizer. The current algorithm is completely deterministic, and uses element of valuation exactly 1 to construct normal presentation. To find such an element we consider basis of our ideal: in Dedekind domain we must have some basis element like this, since otherwise all of them are in $P^2$ which is impossible. If such an element is not found (in our test suite we have exactly one occurence of this in specially constructed non-p-maximal order) we proceed as before with non-normal presentation.
- Implemented special case for 2-elem normal ideals for multiplication/inverse/minimum/norm and basis_matrix computation

### Explicit non-goals (deferred to next pr)
- For powering we kept using Base.power_by_squaring. For ideals over Q(x), this is significantly faster than "naive" normal-presentation powering (due to smaller polynomial on each step). We might revisit it for ideals of function fields over F_q(x) or over number fields, but for now this is not important.
- No search for 2-element and 2-element-normal presentation for a given ideal
- No "mixed" specializations, like 2-element (non-normal) multiplication  with ideal with matrix etc.


alexey-orlov-math:
tagging @fieker and @thofma

src/FunField/IntClsZx.jl
```diff
@@ -150,6 +150,8 @@ function Hecke.integral_closure(Zx::ZZPolyRing, F::Generic.AbsSimpleFunctionFiel
   else
     oo2 = order(o3, integral_split(inv(T2)*basis_matrix(o2), Zx)..., check = false)
   end
+
+  oo2.is_maximal = 1
```
fieker, line 154: is it clear, by context, in what sense the GenOrd is maximal? I suspect the anser is yes, but mathematically being maximal or integrally closed is a bivariate function...
Is it always guaranteed to mean: maximal = integral_closure(coeff_ring, field)?

alexey-orlov-math: am i missing something but isnt it (in commutative case) maximal order == integral closure always?

fieker:
On Mon, Jul 13, 2026 at 05:14:27AM -0700, alexey-orlov-math wrote:
> 
> 
> 
> > @@ -150,6 +150,8 @@ function Hecke.integral_closure(Zx::ZZPolyRing, F::Generic.AbsSimpleFunctionFiel
>    else
>      oo2 = order(o3, integral_split(inv(T2)*basis_matrix(o2), Zx)..., check = false)
>    end
> +
> +  oo2.is_maximal = 1
> 
> am i missing something but isnt it (in commutative case) maximal order == integral closure always?
> 
IntCls(A, B) is the closure of A in B, ie the elements in B that are
roots of monic polynomials over A
E.g. we have
IntCls(Q[x], F = Q(x)[t]/f) = finite max order
IntCls(Q[1/x], F) = infinte max order
IntCls(Z[x], F) = the weird case
IntCls(Q(x), F) = F ...

But we can also do
IntCls(ZZ, K) for a number field K as well as
IntCls(ZZ_(p), K) for a localisation

so a ring R is integrally closed if IntCls(R, Q(R)) == R for the fraction
field Q(R). It's just for "normal number fields" one is lazy

src/FunField/IntClsZx.jl
thofma: Yes, I think maximal means maximal R-order of Q(S), where R = base_ring of the order S. And then, as @alexey-orlov-math said, it coincides with the integral closure of R in Q(S).

alexey-orlov-math: In `src/GenOrd/GenOrd.jl` we have
```
# Support for generic maximal orders over any PID
#
#   final result:
#     integral_closure(R, F)
#     where R is "any" PID and F a finite extension of some quotient field of R
```
so `integral_closure` is explicitly about quotient fields (as @thofma wrote above). 

Further, apart from `Hecke.integral_closure(Zx::ZZPolyRing, F::Generic.AbsSimpleFunctionField)`, all other `integral_closure` are explicitly doing
```
O = GenOrd(S, F)
return Hecke.maximal_order(O)
```
so i'd say it is safe to assume we are consider only "easy" case (R-orders in fraction fields tower)

fieker:
On Mon, Jul 13, 2026 at 09:10:07AM -0700, Tommy Hofmann wrote:
> 
> 
> 
> > @@ -150,6 +150,8 @@ function Hecke.integral_closure(Zx::ZZPolyRing, F::Generic.AbsSimpleFunctionFiel
>    else
>      oo2 = order(o3, integral_split(inv(T2)*basis_matrix(o2), Zx)..., check = false)
>    end
> +
> +  oo2.is_maximal = 1
> 
> Yes, I think maximal means maximal R-order of Q(S), where R = base_ring of the order S. And then, as @alexey-orlov-math said, it coincides with the integral closure of R in Q(S).
OK, this I wanted to clarify.
For my curiosity: this also coincides with S being closed in Q(S)?

thofma:
> OK, this I wanted to clarify. For my curiosity: this also coincides with S being closed in Q(S)?

Yes, I think so.

thofma:
Are you happy with this PR @fieker?

PR #2304, opened 2026-07-13
feat: Add Weierstrass places and gap numbers

JHanselman:
Implements some of the algorithms from Hess, An Algorithm for computing Weierstrass Points.
Adds functions to compute:
- Ramification divisor
- Gap numbers
- Weierstrass places
- Prime power representations
- Differentiations


JHanselman:
@alexey-orlov-math Feel free to make any changes. The Weierstrass places are currently kind of slow, but I expect that that will be changed with your upcoming PRs.

PR #2309, opened 2026-07-17
feat: more on `as(_finite)_bilinear_module` possibilities

StevellM:
- Add keyword argument `as_bilinear_module` for the constructors `(is_genus_with_)genus(::TorQuadModule, ::NTuple{2, Int})`.
- Preserve generators in `_as_finite_bilinear_module`
The first change is to be able to fix a bug in OSCAR (problem with embeddings some even lattices into odd unimodular lattices). The second one is to remove a duplicate function in OSCAR


simonbrandhorst (approved):

StevellM:
@thofma Any idea about what is wrong here ?

Issue #2311, opened 2026-07-22
ZLatAutoCtx init/fingerprint: UndefRefError with >2 simultaneous Gram matrices (w[2] typo should be w[k])


gkorpal:
`Hecke.ZLatAutoCtx`'s `init` (in `src/QuadForm/Morphism.jl`) throws
`UndefRefError` when constructed with more than two simultaneous Gram
matrices. Root cause is a one-character typo: a loop that should fill
`w[k]` for each `k` instead always writes `w[2]`.

## Minimal reproducer

```julia
using Hecke

G1 = 2 * identity_matrix(ZZ, 3)
G2 = 3 * identity_matrix(ZZ, 3)
G3 = 5 * identity_matrix(ZZ, 3)
G4 = 3 * identity_matrix(ZZ, 3)   # deliberately == G2, see note below

C = Hecke.ZLatAutoCtx([G1, G2, G3, G4])
Hecke.init(C, true)
```

Expected: succeeds. Actual:

```
ERROR: LoadError: UndefRefError: access to undefined reference
Stacktrace:
  [1] getindex
    @ ./essentials.jl:920 [inlined]
  [2] fingerprint(C::Hecke.ZLatAutoCtx{ZZRingElem, ZZMatrix, ZZMatrix})
    @ Hecke ~/.julia/packages/Hecke/kYInM/src/QuadForm/Morphism.jl:999
  [3] macro expansion
    @ ~/.julia/packages/AbstractAlgebra/Y7um3/src/Assertions.jl:270 [inlined]
  [4] init(C::Hecke.ZLatAutoCtx{ZZRingElem, ZZMatrix, ZZMatrix}, auto::Bool, bound::ZZRingElem, use_dict::Bool; depth::Int64, bacher_depth::Int64, is_lll_reduced_known::Bool, known_short_vectors::Tuple{Int64, Vector{Any}})
    @ Hecke ~/.julia/packages/Hecke/kYInM/src/QuadForm/Morphism.jl:193
  [5] init (repeats 2 times)
    @ ~/.julia/packages/Hecke/kYInM/src/QuadForm/Morphism.jl:116 [inlined]
  [6] init(C::Hecke.ZLatAutoCtx{ZZRingElem, ZZMatrix, ZZMatrix}, auto::Bool)
    @ Hecke ~/.julia/packages/Hecke/kYInM/src/QuadForm/Morphism.jl:116
  [7] top-level scope
[...]

PR #2312, opened 2026-07-22
Fix UndefRefError in ZLatAutoCtx init for r>2 (#2311)

gkorpal:
The short-vector length loop always wrote w[2] instead of w[k],
   leaving w[3:r] undefined whenever r > 2. Fixes issue #2311

### DIFF
--- a/src/QuadForm/Morphism.jl
+++ b/src/QuadForm/Morphism.jl
@@ -170,7 +170,7 @@ function init(
     w = Vector{ZZRingElem}(undef, r)
     w[1] = numerator(cand[2])
     for k in 2:r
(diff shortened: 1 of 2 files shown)


StevellM:
Hi, thank you for the PR. If you fix a bug, could you please add your previously non-working examples as a new test (in the corresponding file) so that we keep track of it. It also allows to verifies that this is indeed fixed now. :-)

gkorpal:
Added a regression test in `test/QuadForm/Morphism.jl` reproducing the original crash.

thofma:
Thanks for catching (and fixing) this!

PR #2314, opened 2026-08-05
opt: Require assumptions in isomorphism test for fractional ideals

wfsteiner:
Since the isomorphism test currently terminates with a wrong answer for fractional ideals in noncommutative algebras, check the needed assumptions.

### DIFF
--- a/src/AlgAssAbsOrd/ICM.jl
+++ b/src/AlgAssAbsOrd/ICM.jl
@@ -189,11 +189,8 @@ function ideals_containing(S::T, a::T2, R::T) where { T <: Union{ AbsNumFieldOrd
     return [ a ]
   end
 
-  K = _algebra(S)
   d = degree(S)
-
   potential_basis = Vector{elem_type(_algebra(S))}(undef, d)
-  ideals = fractional_ideal_type(R)[]
 
   for i = 1:mQ.offset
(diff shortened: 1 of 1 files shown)


thofma:
Is the check expensive?

wfsteiner:
In the case of `AbsNumFieldOrder` this short circuits to true, see

https://github.com/thofma/Hecke.jl/blob/12736ed797cbeecdd8d12a072c9188c919f19793/src/NumFieldOrd/NumFieldOrd.jl#L31

In the case of `AlgAssAbsOrd`, I assume it depends. If the order happens to be constructed via a polynomial, commutativity is recorded here

https://github.com/thofma/Hecke.jl/blob/12736ed797cbeecdd8d12a072c9188c919f19793/src/AlgAss/StructureConstantAlgebra.jl#L344

and it also short circuits. This would be the case for the tests. If one happens to work with an order in an arbitrary algebra, the worst case I assume is that the commutativity test for the structure constant algebra would be called while it actually is commutative.

Overall I would not count it as expensive.

thofma:
If it is not too much work, I would suggest to always do the `is_commutative` check and to remove the `check` keyword argument.

(You can just use `@req is_commutative(R)`)

PR #2315, opened 2026-08-05
fix: Fix isomorphism tests and calculations for modules over algebras

wfsteiner:
Add some minor repairs; fix branches that were crashing because of non-implemented methods, and fix a branch condition that was always evaluating to false, resulting in type instability. While there, remove some unused variables in function bodies.


thofma:
good catch, thanks

PR #2318, opened 2026-08-09
feat: add `multiplicative_generator` for finite fields

thofma:
This adds `multiplicative_generator(::FinField)`, which promises to cache the result.

### DIFF
--- a/src/Misc/FiniteField.jl
+++ b/src/Misc/FiniteField.jl
@@ -125,7 +125,7 @@ end
 #
 ################################################################################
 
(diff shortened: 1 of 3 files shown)


thofma:
@fieker I think this is what we agreed upon a while ago?

fieker:
On Sun, Aug 09, 2026 at 10:16:20PM -0700, Tommy Hofmann wrote:
> thofma left a comment (thofma/Hecke.jl#2318)
> 
> @fieker I think this is what we agreed upon a while ago?

I think so. In a second step we might (might!) consider making it stable
across runs as well. But this is what we agreed on

fieker (approved):

PR #2329, opened 2026-08-16
fix: return the structure constant algebra of a cyclic algebra

lucifer1004:
## Summary

- return the structure constant algebra already stored by `CyclicAlgebra`
- cover the public accessor with a regression

`StructureConstantAlgebra(c.sca)` invokes a conversion that returns an algebra-map pair. The accessor asserts a structure-constant-algebra return type, so calling it currently raises a type error. Returning `c.sca` directly matches both the documented contract and the existing cyclic-algebra representation.

## Test

`julia --project=. -e 'using Hecke, Test; include("test/AlgAss/AlgCyc.jl")'`

Result: 22/22 tests passed.

### DIFF
--- a/src/AlgAss/AlgCyc.jl
+++ b/src/AlgAss/AlgCyc.jl
@@ -65,7 +65,7 @@ end
 function structure_constant_algebra(
   c::CyclicAlgebra{T, S},
 ) where {T, S}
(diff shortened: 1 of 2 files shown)


thofma:
@wfsteiner makes sense?

wfsteiner:
Makes sense, thanks!

PR #2339, opened 2026-08-16
fix: clear rational local norm denominators correctly

lucifer1004:
## Summary

- clear the denominator of a rational local-norm query by multiplying by the norm of the denominator
- add quadratic and quartic regression tests for rational norms and non-norms

## Rationale

For `a = n/d` and `[K : QQ] = m`, multiplication by the norm of the embedded rational scalar `d` preserves the norm class:

```
a * N_{K/QQ}(d) = (n/d) * d^m = n * d^(m - 1).
```

The previous exponent `m` therefore changed the queried norm class instead of merely clearing the denominator.

## Testing

```
julia --startup-file=no --project=. -e 'using Test, Hecke; include("test/setup.jl"); include("test/LocalField/neq.jl")'\n```\n\nAll LocalField tests passed; the `LocalNorm` testset passed 6/6.

### DIFF
--- a/src/LocalField/neq.jl
+++ b/src/LocalField/neq.jl
@@ -1479,7 +1479,7 @@ function is_local_norm(k::Hecke.AbsSimpleNumField, a::Integer)
 end
 
 function is_local_norm(k::Hecke.AbsSimpleNumField, a::QQFieldElem)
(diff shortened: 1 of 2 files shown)


simonbrandhorst (approved):
Thanks

PR #2342, opened 2026-08-16
RFC implementation: expose quadratic Kummer generators

### LINKED ISSUE #2340: RFC: expose a quadratic Kummer generator without materializing the class field
## Status

Proposed API; this issue is intended to settle the public contract before the implementation is finalized.

## Motivation

A formal quadratic `ClassField` already specifies an exact index-two Artin quotient. Some callers need only a Kummer representative of that quotient—for example, to evaluate Frobenius characters or to feed a later construction. Today the available construction path continues through relative-field materialization and descent even when those objects are not needed.

The requested operation is not a search: it should extract the quadratic Kummer square class determined by the existing ray class quotient.

## Mathematical contract

Let `C` be a formal class field over `K` with `degree(C) == 2`. The operation should return a factored element `a` of `K` such that `K(sqrt(a))/K` realizes the Artin quotient encoded by `C.rayclassgroupmap` and `C.quotientmap`. The representative should be reduced modulo squares.

For an admissible prime ideal `P`, the Kummer Frobenius coordinate obtained from `a` must agree with

```julia
C.quotientmap(preimage(C.rayclassgroupmap, P))
```

under the canonical identification with the group of order two.

## Proposed API

```julia
quadratic_kummer_generator(C::ClassField) -> FacElem
```

The method would:

- require `degree(C) == 2`;
- return a `FacElem` over `base_field(C)`, reduced modulo squares;
- leave the formal input unmaterialized: in particular, it would not construct `C.A` or `C.cyc`;
- avoid introducing a wrapper that duplicates the ray class and quotient maps already owned by `C`.

## Implementation outline

The existing ray class implementation already contains the required computations. A minimal implementation can create an ephemeral `ClassField_pp`, reuse the S-unit and inverse-Kummer steps, reduce the resulting factored generator modulo squares, and stop before `radical_extension` and descent.

This is intended to reuse the current mathematical kernel, not to add a second class-field algorithm.

## Verification

Regression tests should:

1. construct a formal quadratic ray class field;
2. obtain the factored Kummer generator;
3. verify that the formal `ClassField` remains unmaterialized;
4. compare `canonical_frobenius(P, kummer_extension(2, [a]))` with the defining Artin quotient at several admissible primes.

### Return a dedicated result wrapper

A wrapper could retain the ray class map, quotient map, factored generator, and evaluated generator. This duplicates data already present in `ClassField` and creates a larger public API without adding mathematical information.

### Expose only a low-level RCF helper

This would minimize public surface area, but callers would need to construct and mutate the internal `ClassField_pp` state correctly. A method on formal `ClassField` gives the operation a clearer precondition and owner.

### Continue through `number_field(C)`

This is correct but materializes objects that are unnecessary when the desired result is only the exact quadratic square class or its Frobenius character.

## Non-goals

- search or conductor-budget policy;
- application-specific evidence or replay types;
- a new Frobenius wrapper (the existing Kummer machinery is sufficient);
- support for higher-degree Kummer quotients in this proposal.

## Open questions

1. Is `quadratic_kummer_generator` the preferred public name, or should this be a more general `kummer_generator` API restricted initially to degree two?
2. Should the public result remain factored only, or should the method return both the factored and evaluated representative?
3. Should the computation be cached on `ClassField`, or remain a non-materializing operation with only the existing lower-level caches?

lucifer1004:
## Design status

Draft implementation for #2340. The issue is the API/design discussion; this PR is intentionally draft until the public name, return shape, and caching policy are agreed.

## Summary

- add `quadratic_kummer_generator(C::ClassField) -> FacElem`
- reuse the existing S-unit and inverse-Kummer computations
- reduce the result modulo squares and stop before `radical_extension` and descent
- keep the formal `ClassField` unmaterialized
- verify the returned Kummer character against the defining Artin quotient

The implementation does not introduce a result wrapper, duplicate the ray class maps, or add a second Frobenius API.

## Mathematical contract

For a formal degree-two class field `C` over `K`, the returned factored element `a` lies in `K` and `K(sqrt(a))/K` realizes the Artin quotient encoded by `C`.

The test compares

```julia
C.quotientmap(preimage(C.rayclassgroupmap, P))
```

with `canonical_frobenius(P, kummer_extension(2, [a]))` at six admissible primes.

## Local verification

The focused quadratic Kummer testset passes 12/12 under Julia 1.12.7. It checks that `C.A` and `C.cyc` remain undefined before and after extraction, that the returned element is a nonsquare over the base field, and that all six Frobenius comparisons agree.

Closes #2340

### DIFF
--- a/src/RCF/rcf.jl
+++ b/src/RCF/rcf.jl
@@ -104,6 +104,35 @@ function ray_class_field_cyclic_pp(CFpp::ClassField_pp; using_stark_units::Bool 
   return CFpp
 end
 
+@doc raw"""
+    quadratic_kummer_generator(C::ClassField) -> FacElem
+
+Return a factored Kummer generator for the formal quadratic class field `C`,
+reduced modulo squares. This does not materialize the relative class field.
+"""
+function quadratic_kummer_generator(CF::ClassField{S, T}) where {S, T}
+  degree(CF) == 2 || throw(ArgumentError(
+    "Kummer generator extraction requires a quadratic class field"))
+
+  CFpp = ClassField_pp{S, T}()
+  CFpp.rayclassgroupmap = CF.rayclassgroupmap
+  CFpp.quotientmap = CF.quotientmap
+  _rcf_S_units(CFpp)
+  _rcf_find_kummer(CFpp)
+
+  isdefined(CFpp, :o) && CFpp.o == 2 || error(
+    "failed to find a quadratic Kummer generator")
+  isdefined(CFpp, :defect) && isone(CFpp.defect) || error(
+    "quadratic Kummer generator has a descent defect")
+  isdefined(CFpp, :sup_known) && CFpp.sup_known || error(
+    "support of the quadratic Kummer generator is not known")
+
+  a = reduce_mod_powers(CFpp.a, 2, CFpp.sup)
+  base_ring(a) === base_field(CF) || error(
+    "quadratic Kummer generator is not defined over the class field base")
+  return a
+end
+
 ################################################################################
 #
 #  Using norm relations to get the S-units
--- a/src/exports.jl
+++ b/src/exports.jl
@@ -809,6 +809,7 @@ export push!
 export push_through_isogeny
 export quadratic_defect
 export quadratic_field
+export quadratic_kummer_generator
 export quadratic_lattice
 export quadratic_product
 export quadratic_space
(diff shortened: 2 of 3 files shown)


thofma:
We can have this if this is helpful for you.

lucifer1004:
Thanks @thofma, I have marked the PR as ready for reivew.

PR #2344, opened 2026-08-19
Compute rational signature_tuple via leading principal minors

simonbrandhorst:
## Summary
- `signature_tuple(q::QuadSpace{QQField,QQMatrix})` now determines the signature from the sequence of leading principal minors $D_0=1, D_1, \dots, D_m$ of a Gram matrix of the non-degenerate part of `q`, using Jacobi's theorem (the number of sign changes in this sequence equals the number of negative eigenvalues). This avoids computing an explicit basis-change matrix.
- Each minor is checked as soon as it is computed, and the function bails out to the fallback the moment one is zero, instead of computing the full sequence of (increasingly expensive) minors first and only then checking whether any of them vanished.
- When the Gram matrix is already non-degenerate (kernel trivial, the common case), the basis-completion step and the two matrix multiplications it would otherwise take to restrict to the non-degenerate part are skipped entirely, since they would be a no-op (identity) in that case.
- When a leading principal minor vanishes (e.g. a hyperbolic plane sitting on the diagonal, where Jacobi's theorem does not directly apply), the function falls back to the previous exact Gram-Schmidt diagonalization, which handles this case via pivoting.
- The result is still exact (`QQFieldElem` arithmetic throughout), so `signature_tuple` remains provably correct.

## Benchmarks

With #2345 merged, `diagonal`/`_gram_schmidt` itself got much faster, so it was worth re-checking whether this PR is still worthwhile on top of that. Compared `signature_tuple` from this PR against always calling `diagonal(q)` (i.e. the pre-this-PR behavior, now benefiting from #2345), on top of current `master`:

**Generic full-rank matrices (no fallback triggered):**

| n | diag (ms) | minors (ms) | speedup | diag alloc | minors alloc | alloc ratio |
|---|---|---|---|---|---|---|
| 5 | 0.018 | 0.004 | 4.54x | 25,951 | 5,106 | 5.08x |
| 10 | 0.121 | 0.018 | 6.71x | 163,859 | 18,912 | 8.66x |
| 20 | 7.69 (*) | 0.175 | 43.9x (*) | 1,303,573 | 91,434 | 14.26x |
| 40 | 38.27 | 1.86 | 20.55x | 12,344,300 | 1,461,624 | 8.45x |
| 80 | 769.32 | 40.84 | 18.84x | 107,604,624 | 17,098,267 | 6.29x |
| 120 | 3461.02 | 381.00 | 9.08x | 300,421,408 | 64,551,560 | 4.65x |

(*) the n=20 "diag" timing looks like a noisy outlier relative to its neighbors; treat that one speedup figure loosely, the rest of the trend is consistent. Still a clear win across the board, roughly 5x-20x depending on size, with allocations lower by a similar margin.

**Matrices with a zero leading diagonal entry** (e.g. a hyperbolic plane in the standard basis, common for real lattice Gram matrices), which force the fallback path — across the three iterations of this PR:

| case | original (compute all minors first) | + early bail-out | + skip basis completion when already non-degenerate |
|---|---|---|---|
| n=10 | 0.72x | 0.81x | 0.95x |
| n=20 | 0.71x | 0.81x | 0.95x |
| n=40 | 0.37x | 0.81x | 0.91x |

Originally, computing the full sequence of minors before discarding them made this case increasingly slower than the plain `diagonal(q)` fallback as `n` grew (down to 0.37x at n=40, unbounded further out). The early bail-out capped that at a flat ~19% overhead regardless of `n`. Since these particular matrices are already full rank (trivial kernel), the basis-completion step was pure unnecessary overhead here too; skipping it brings the remaining gap down to ~5-9%, essentially just the cost of the one trivial determinant before falling back.

## Test plan
- [x] Added a `signature_tuple via leading principal minors` testset in `test/QuadForm/Quad/Spaces.jl` covering: empty matrix, isotropic/definite 1x1, hyperbolic plane (forces the fallback path), rank-1 degenerate forms, and a mixed degenerate/indefinite 4x4 example.
- [x] Cross-checked against an explicit Gram-Schmidt diagonalization on 131 random/edge-case matrices (including degenerate and hyperbolic-plane cases that trigger the fallback), re-run after each change in this PR.
- [x] `test/QuadForm/Quad/Spaces.jl` (2632 tests) passes.
- [x] `test/QuadForm/Quad/ZGenus.jl` (1332 tests, 1 unrelated harness-only error from running the file standalone) passes.
- [x] `test/QuadForm/Quad/ZLattices.jl` (1371 tests, 1 pre-existing unrelated broken test) passes.

### DIFF
--- a/src/QuadForm/Quad/Spaces.jl
+++ b/src/QuadForm/Quad/Spaces.jl
@@ -2117,10 +2117,6 @@ end
     signature_tuple(q::QuadraticSpace{QQField,QQMatrix) ->Tuple{Int,Int,Int}
 
 Return the number of (positive, zero, negative) inertia of this rational quadratic space.
-
-This is computed from the sequence of leading principal minors of a Gram
-matrix of the non-degenerate part of `q` via Jacobi's theorem, which is
-usually faster than computing an explicit diagonalization.
 """
 @attr Tuple{Int,Int,Int} function signature_tuple(q::QuadSpace{QQField,QQMatrix})
   g = gram_matrix(q)
--- a/src/QuadForm/Quad/Spaces.jl
+++ b/src/QuadForm/Quad/Spaces.jl
@@ -2149,7 +2149,7 @@ Return the number of (positive, zero, negative) inertia of this rational quadrat
 
   # Jacobi's theorem does not apply directly when a leading principal minor
   # vanishes (e.g. a hyperbolic plane on the diagonal); fall back to an
-  # explicit diagonalization, which handles this case via pivoting.
+  # explicit diagonalization
   D = diagonal(q)
   pos = count(d > 0 for d in D)
   neg = count(d < 0 for d in D)
--- a/src/QuadForm/Quad/Spaces.jl
+++ b/src/QuadForm/Quad/Spaces.jl
@@ -2117,13 +2117,47 @@ end
     signature_tuple(q::QuadraticSpace{QQField,QQMatrix) ->Tuple{Int,Int,Int}
 
 Return the number of (positive, zero, negative) inertia of this rational quadratic space.
+
+This is computed from the sequence of leading principal minors of a Gram
+matrix of the non-degenerate part of `q` via Jacobi's theorem, which is
+usually faster than computing an explicit diagonalization.
 """
 @attr Tuple{Int,Int,Int} function signature_tuple(q::QuadSpace{QQField,QQMatrix})
+  g = gram_matrix(q)
+  n = ncols(g)
+  n == 0 && return (0, 0, 0)
+
+  K = kernel(g, side = :left)
+  k = nrows(K)
+  k == n && return (0, n, 0)
+
+  B = complete_to_basis(K)
+  gg = B[k+1:end, :] * g * transpose(B[k+1:end, :])
+  m = ncols(gg)
+
+  # Leading principal minors D_0 = 1, D_1, ..., D_m of the non-degenerate part.
+  minors = Vector{QQFieldElem}(undef, m + 1)
+  minors[1] = one(QQ)
+  for r in 1:m
+    minors[r + 1] = det(@view gg[1:r, 1:r])
+  end
+
+  if all(!iszero, minors)
+    # Jacobi's theorem: if all leading principal minors of a symmetric matrix
+    # are nonzero, the number of negative eigenvalues equals the number of
+    # sign changes in the sequence D_0, D_1, ..., D_m.
+    neg = count(r -> (minors[r] > 0) != (minors[r + 1] > 0), 1:m)
+    pos = m - neg
+    return (pos, k, neg)
+  end
+
+  # Jacobi's theorem does not apply directly when a leading principal minor
+  # vanishes (e.g. a hyperbolic plane on the diagonal); fall back to an
+  # explicit diagonalization, which handles this case via pivoting.
   D = diagonal(q)
-  pos = count(d>0 for d in D)
-  zero = count(d==0 for d in D)
-  neg = count(d<0 for d in D)
-  return (pos, zero, neg)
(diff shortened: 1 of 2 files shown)


test/QuadForm/Quad/Spaces.jl
```diff
@@ -5,14 +5,45 @@
...
+        g = A + transpose(A)
+        V = quadratic_space(QQ, g)
+        D = diagonal(V)
+        pos = count(d -> d > 0, D)
+        zer = count(d -> d == 0, D)
+        neg = count(d -> d < 0, D)
+        @test Hecke.signature_tuple(V) == (pos, zer, neg)
+      end
+    end
+  end
```
simonbrandhorst:
```suggestion
```

simonbrandhorst (commented):
AI changes look good to me.

src/QuadForm/Quad/Spaces.jl
```diff
@@ -2138,13 +2117,47 @@ end
     signature_tuple(q::QuadraticSpace{QQField,QQMatrix) ->Tuple{Int,Int,Int}
 
 Return the number of (positive, zero, negative) inertia of this rational quadratic space.
+
+This is computed from the sequence of leading principal minors of a Gram
+matrix of the non-degenerate part of `q` via Jacobi's theorem, which is
+usually faster than computing an explicit diagonalization.
```
simonbrandhorst:
```suggestion
```

src/QuadForm/Quad/Spaces.jl
```diff
@@ -2140,11 +2119,41 @@ end
...
+    # are nonzero, the number of negative eigenvalues equals the number of
+    # sign changes in the sequence D_0, D_1, ..., D_m.
+    neg = count(r -> (minors[r] > 0) != (minors[r + 1] > 0), 1:m)
+    pos = m - neg
+    return (pos, k, neg)
+  end
+
+  # Jacobi's theorem does not apply directly when a leading principal minor
+  # vanishes (e.g. a hyperbolic plane on the diagonal); fall back to an
+  # explicit diagonalization, which handles this case via pivoting.
```
simonbrandhorst:
```suggestion
  # explicit diagonalization
```

test/QuadForm/Quad/Spaces.jl
```diff
@@ -5,14 +5,31 @@
...
+    # hyperbolic plane: the leading principal minor D_1 vanishes, forcing the
+    # fallback diagonalization
+    @test Hecke.signature_tuple(quadratic_space(QQ, QQ[0 1; 1 0])) == (1, 0, 1)
+    # rank-1 degenerate form
+    @test Hecke.signature_tuple(quadratic_space(QQ, QQ[1 1; 1 1])) == (1, 1, 0)
+    @test Hecke.signature_tuple(quadratic_space(QQ, QQ[2 1; 1 2])) == (2, 0, 0)
+    @test Hecke.signature_tuple(quadratic_space(QQ, QQ[-2 1; 1 -2])) == (0, 0, 2)
+    # hyperbolic plane plus a definite direction, with a degenerate direction
+    g = matrix(QQ, [0 1 0 0; 1 0 0 0; 0 0 -3 0; 0 0 0 0])
+    @test Hecke.signature_tuple(quadratic_space(QQ, g)) == (1, 1, 2)
```
simonbrandhorst, line 25:
```suggestion
    @test Hecke.signature_tuple(quadratic_space(QQ, g)) == (1, 1, 2)
  end
```

simonbrandhorst:
Let's wait for #2345 and then measure whether this is still an improvement.

thofma:
https://github.com/thofma/Hecke.jl/pull/2345 has landed

simonbrandhorst:
Ready to go from my side. More could still be done but I guess that would be over engineering it.

PR #2345, opened 2026-08-19
Make _gram_schmidt use elementary row/column operations

simonbrandhorst:
## Summary
- `_gram_schmidt` (used throughout the quadratic/Hermitian space machinery, e.g. `diagonal`, `signature_tuple`, `lll_gram_indef_with_transform`, isometry testing, ...) previously built a dense $n \times n$ transformation matrix $T$ at every pivot step and computed $T \cdot F \cdot \overline{T}^t$ via two full matrix multiplications, even though $T$ only ever differs from the identity in a handful of entries (a single row swap, a single off-diagonal entry, or a single column's worth of elimination coefficients). This costs $O(n^3)$ per step, i.e. $O(n^4)$ overall.
- This PR applies the same elementary transforms directly with `add_row!`/`add_column!`/`swap_rows!`/`swap_cols!`, which only touch the rows/columns that actually change, bringing the complexity back down to the usual $O(n^3)$ of symmetric Gaussian elimination.
- Also replaces `iszero(F[i, j])`-style zero tests (which allocate a fresh ring/field element via `getindex` just to test it) with `is_zero_entry(F, i, j)`, which for `QQMatrix`/`ZZMatrix` reads the FLINT entry directly with no allocation, and falls back to the same behavior for other matrix types.
- The pivoting logic and `nondeg` error-on-non-full-rank semantics are unchanged; only *how* each elementary transform is applied to `F` and `S` changed, not *which* transforms are chosen.

## Benchmarks

Timed the old implementation (kept temporarily under a different name) against the new one on random full-rank symmetric `QQMatrix` inputs of size `n x n` (`nondeg = true`), warmed up, averaged over multiple reps:

| n | old (ms) | new (ms) | speedup | old alloc (bytes) | new alloc (bytes) | alloc ratio |
|---|---|---|---|---|---|---|
| 5 | 0.57 | 0.60 | 0.95x | 21,640 | 17,920 | 1.21x |
| 10 | 0.23 | 0.14 | 1.59x | 154,766 | 140,272 | 1.10x |
| 20 | 3.32 | 1.40 | 2.37x | 1,162,929 | 1,145,341 | 1.02x |
| 40 | 193.5 | 32.28 | 6.0x | 50,442,099 | 11,413,936 | 4.42x |
| 60 | 1980.1 | 185.47 | 10.68x | 406,120,592 | 37,585,112 | 10.81x |
| 80 | 13940.84 | 675.64 | 20.63x | 1,853,004,168 | 99,493,104 | 18.62x |

At tiny sizes (n=5, n=10) the two are roughly on par, dominated by constant overhead/noise rather than algorithmic complexity. From n~20 up, the gap grows as expected for an $O(n^3)$ vs $O(n^4)$ difference: at n=80 the rewrite is over 20x faster and allocates 18x less.

## Test plan
- [x] Added a `_gram_schmidt pivoting` testset to `test/QuadForm/Quad/Spaces.jl` covering the row-swap branch, the off-diagonal/hyperbolic-pair branch, and random full-rank matrices, checking `is_diagonal(F)` and `S * M * transpose(S) == F`.
- [x] Cross-checked the new implementation against the old one (kept temporarily under a different name) on 463 random/edge-case matrices over `QQMatrix` (both `nondeg = true` and `nondeg = false`) and over a genuine number field with a nontrivial involution, comparing both the returned `F` and that `S` satisfies the defining identity.
- [x] `test/QuadForm/Quad/Spaces.jl` (2799 tests) passes.
- [x] `test/QuadForm/Herm/{Spaces,Genus,GenusRep,Lattices}.jl` (117 tests) passes.
- [x] `test/QuadForm/indefiniteLLL.jl` passes, which exercises `_gram_schmidt(H, QQ)` — i.e. `a` being a coercion callable rather than `identity`.

### DIFF
--- a/src/QuadForm/Spaces.jl
+++ b/src/QuadForm/Spaces.jl
@@ -335,37 +335,41 @@ diagonal_with_transform(V::AbstractSpace)
 ################################################################################
 
 # Clean this up
+#
+# Unlike a textbook Gram-Schmidt process, the elementary transformations here
+# are applied to F and S via targeted row/column operations (add_row!,
+# add_column!, swap_rows!, swap_cols!) instead of building a dense n x n
+# transformation matrix T and computing T * F * transpose(_map(T, a)) at
+# every step. The latter costs O(n^3) per pivot, i.e. O(n^4) overall, whereas
+# the elementary operations only ever touch the rows/columns that actually
+# change, giving the usual O(n^3) complexity of symmetric Gaussian
+# elimination. Zero-testing uses is_zero_entry(A, i, j), which for e.g.
+# QQMatrix/ZZMatrix avoids allocating a field/ring element just to inspect
+# a single entry (unlike A[i, j]).
 function _gram_schmidt(M::MatElem, a, nondeg = true)
   F = deepcopy(M)
   K = base_ring(F)
   n = nrows(F)
   S = identity_matrix(K, n)
-  T = identity_matrix(K, n)
-  okk = is_diagonal(F)
-  if !okk
+  if !is_diagonal(F)
     for i in 1:n
-      if iszero(F[i,i])
-        zero!(T)
-        for i in 1:nrows(T)
-          T[i, i] = 1
-        end
+      if is_zero_entry(F, i, i)
         ok = 0
         for j in (i + 1):n
-          if !iszero(F[j, j])
+          if !is_zero_entry(F, j, j)
             ok = j
             break
           end
         end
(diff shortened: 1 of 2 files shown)


simonbrandhorst:
The AI thingy looks good to me. I did not check the exact order of each row/colum operation, but I am confident that any nonsense would blow up in the tests.

thofma:
New strategy makes sense. I restarted the failing job(s).

PR #2352, opened 2026-08-20
fix: `can_solve` for binary quadratic forms missed solutions with y = 0

simonbrandhorst:
The enumeration in `can_solve_with_solution(f::QuadBin, n)` ran `for y in 1:floor(sqrt(4*a*n/abs(D)))`, so solutions with `y == 0` were never examined.

```julia
julia> f = binary_quadratic_form(1, 0, 5)
Binary quadratic form over ZZ: x^2 + 5*y^2

julia> f(2, 0)
4

julia> can_solve(f, 4)
false
```

The bound itself is right — from `4*a*f(x, y) = (2*a*x + f[2]*y)^2 - D*y^2` one gets `abs(D)*y^2 <= 4*a*n` — but `f(x, y) == f(-x, -y)` only lets one drop the *negative* values of `y`, not `y == 0`. The loop now starts at `0`.

Two further problems on the same line:

```julia
julia> can_solve(binary_quadratic_form(1, 0, 1), -1)
ERROR: DomainError with -1.0:
sqrt was called with a negative real argument ...
```

For `n` of the sign opposite to the leading coefficient the bound is the square root of a negative rational. A definite form represents nothing of the opposite sign, so this returns `false` early now. And the `Int(...)` conversions overflowed for large arguments; the remaining arithmetic is done over `ZZ`.

**Behaviour change to note:** `can_solve(f, 0)` is now `true` with solution `(0, 0)`. That matches the docstring (`f(0, 0) = 0`) and the convention used elsewhere in Hecke that every quadratic form represents 0.

**Testing.** Compared against a naive search over all 46604 pairs of a definite form `[a, b, c]` with entries in `-6..6` and a target in `-30..30`: the results agree everywhere, and every reported solution evaluates to the target. The existing assertions in `test/QuadForm/QuadBin.jl` are unchanged and still pass; a smaller exhaustive comparison is now part of the test file.

### DIFF
--- a/src/QuadForm/QuadBin.jl
+++ b/src/QuadForm/QuadBin.jl
@@ -158,24 +158,38 @@ end
 
 For a binary quadratic form `f` with negative discriminant and an integer `n`,
 return the tuple `(true, (x, y))` if $f(x, y) = n$ for integers `x`, `y`.
-If no such integers exist, return `(false, (0, 0))`
+If no such integers exist, return `(false, (0, 0))`.
+
+Note that $f(0, 0) = 0$, so `n == 0` is always solvable.
 """
 function can_solve_with_solution(f::QuadBin, n::IntegerUnion; sol::Bool = true)
-  @req discriminant(f) < 0 "f must have negative discriminant"
-  for y in 1:Int(floor(sqrt(Int(4*f[1]*n)//abs(Int(discriminant(f))))))
+  D = discriminant(f)
+  @req D < 0 "f must have negative discriminant"
+  a = f[1]
+  _n = ZZ(n)
+  # A definite form only represents values of the sign of its leading
+  # coefficient, so there is nothing to enumerate otherwise.
+  if sign(a) * sign(_n) < 0
+    return false, (ZZ(0), ZZ(0))
+  end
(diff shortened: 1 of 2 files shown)


src/QuadForm/QuadBin.jl
```diff
@@ -158,24 +158,38 @@ end
...
+  D = discriminant(f)
+  @req D < 0 "f must have negative discriminant"
+  a = f[1]
+  _n = ZZ(n)
+  # A definite form only represents values of the sign of its leading
+  # coefficient, so there is nothing to enumerate otherwise.
+  if sign(a) * sign(_n) < 0
+    return false, (ZZ(0), ZZ(0))
+  end
+  # From 4*a*f(x, y) = (2*a*x + f[2]*y)^2 - D*y^2 we get, for f(x, y) = n,
```
simonbrandhorst:
```suggestion
  end
  # f = a x^2 + b xy + c y^2
  # D = b^2 - 4ac
  # From 4*a*f(x, y) = (2*a*x + b*y)^2 - D*y^2 we get, for f(x, y) = n,
```

PR #2355, opened 2026-08-22
feat: implement basis improvement of Allombert-Chenevier

thofma:
cc @simonbrandhorst


simonbrandhorst (commented):
I think `reduce` is exported. You sure we want this kind of interface?
reduce can also be meant in the sense of lattice reduction, then
there are several algorithms that one can should of lll, bkz etc. 
So perhaps call it _reduce_allombert_chenevier

thofma:
Oops, I think I wanted to call it `_reduce` and not export it. I will fix it with the name you suggest.

PR #2361, opened 2026-08-26
Use weak popov form for fractional ideals over function fields

alexey-orlov-math:
## Goal

Make fractional ideal arithmetic over function fields fast enough for divisor computations over `Q(x)`, where canonical HNF suffers from severe coefficient swell. 

Before this, for fractional ideals we had a canonical representation (integral ideal with denominator) and basis matrix in base field (used only for input). Now, we do not require a canonical basis for fractional ideals: the second representation becomes (first-class citizen) basis matrix (in coordinate ring) with denominator (shared with ideal). To an extent, it can be said that we now (optionally) store integral ideal basis matrix, but we allow it to be non-canonical. To actually gain performance, we introduce the reduction trait based on the coefficient ring type: HNF by default (no semantic changes here) and weak Popov for poynomial rings (ideals in function fields over `Q(x)` and `F_q(x)`). 
All ideal arithmetic (`*`, `+`, `intersect`, `colon`, `inv`) is unified behind this trait: there are special cases for integer/fractional ideals, but the final "let's build matrix stack and reduce it to the basis" is shared.
NOTE: after lots of back-and-forth it was decided that in the case of HNF reduction, fractional ideal will always create integral ideal (and never keep the matrix): first of all the basis matrix in this case is exactly the basis matrix of integral ideal, it also allows us to better use intergal-ideal-specific optimizations dependent on having HNF form.

## New files

- `GenOrd/Matrix.jl`: matrix kernels. Modular lower-left HNF (moved from Ideal.jl) and a Hecke-owned weak Popov kernel. We keep our own Popov copy because AA's `weak_popov!` mutates entries in place and because we added the content-stripping for `Q[x]`.
- `GenOrd/TwoElement.jl`: generator-based fast paths. Some were moved some were added (most notably principal ideals are handled way better now, and allow mixing with two-element representation). 
- `GenOrd/IdealArithmetic.jl`: trait-dispatched stack-and-reduce implementations, and actual GenOrdIdl/GenOrdFracIdl implemetations

## Smaller fixes / changes along the way

- better and more robust `is_zero`/`is_one`.
- better handling of principal ideals, better/more strict construction of ideals from one/two generators.
- better and more full handling of ideal multiplication by scalar
- stricter typed norm of fractional ideal
- applied similar change to inverse basis matrices of fractional ideals (matrix over coefficient ring + denominator). As a sideeffect, `containment_by_matrices` now works with `KInftyRing`.

## Result

Over `Q(x)` Popov weak form is way faster and has better coefficient/degree growth than HNF, giving great speedups. Similarly over `F_q(x)` Popov is way better, apart from two possible caveats:
- fractional ideals `+` might be a bit faster with HNF, since it operates on small $2n \times n$ stack with very good modulus (gcd of minima).
- fractional ideals intersect might be a bit faster, since with Popov we need to do Popov reduction with transform (to find kernel) and then extra reduction to reduce a basis.

Note that these exceptional cases were happening on simpler fields and as "standalone" operations.

Tests run comparison (not super indicative, since the fields/ideals involved are small/simple, and these are tests; but at least it shows there is a universal speedup)
Before:
```
julia> @btime include("test/FunField/RiemannRoch.jl");
<...>
  2.368 s (40559673 allocations: 2.47 GiB)
julia> @btime include("test/FunField/Divisor.jl");
<...>
  731.136 ms (15883503 allocations: 829.67 MiB)
```
After:
```
julia> @btime include("test/FunField/RiemannRoch.jl");
<...>
  1.371 s (25576825 allocations: 1.49 GiB)
julia> @btime include("test/FunField/Divisor.jl");
<...>
  607.231 ms (12115001 allocations: 653.70 MiB)
```

Now for real workloads, we use one of the Jeroen's test polynomials. The setup is:
```
kx, x = rational_function_field(QQ, :x; cached = false);
ky, y = polynomial_ring(kx, :y; cached = false);
f = y^3*((y^4-16)^2 - 2*x) - x^12;
L, t = function_field(f; cached = false);
Ofin = finite_maximal_order(L);
Oinf = infinite_maximal_order(L);
x = Hecke.separating_element(L);
dx = differential(x);
codiff_divisor = Hecke.divisor(codifferent(Ofin), codifferent(Oinf));
D = (-Hecke.divisor(dx.f) + 2*pole_divisor(L(x)) + codiff_divisor);
```

Before:
```
julia> @btime (-Hecke.divisor(dx.f) + 2*pole_divisor(L(x)) + codiff_divisor);
  633.798 ms (2204242 allocations: 768.79 MiB)

julia> @btime zero_divisor(D) - pole_divisor(D);
  27.479 s (30101654 allocations: 99.76 GiB)
```

After:
```
julia> @btime (-Hecke.divisor(dx.f) + 2*pole_divisor(L(x)) + codiff_divisor);
  679.291 μs (49488 allocations: 1.70 MiB)

julia> @btime zero_divisor(D) - pole_divisor(D);
  142.803 ms (1445719 allocations: 121.85 MiB)
```

Same but with 
```
R, _ = finite_field(17, 5; cached = false)
kx, x = rational_function_field(R, :x; cached = false);
```

Before:
```
julia> @btime (-Hecke.divisor(dx.f) + 2*pole_divisor(L(x)) + codiff_divisor);
  21.674 ms (963967 allocations: 75.80 MiB)

julia> @btime zero_divisor(D) - pole_divisor(D);
  298.057 ms (16162232 allocations: 1.45 GiB)
```

After:
```
julia> @btime (-Hecke.divisor(dx.f) + 2*pole_divisor(L(x)) + codiff_divisor);
  1.557 ms (103469 allocations: 5.74 MiB)

julia> @btime zero_divisor(D) - pole_divisor(D);
  160.853 ms (4774038 allocations: 512.81 MiB)
```


alexey-orlov-math:
tagging @thofma @fieker @JHanselman

thofma:
Nice! I wonder if we should have some benchmarking list for matrices over k[x] (various k) arising in these computations. @stelip42 has some ideas for a new algorithm and some real life test cases would be nice.

thofma (commented):
have not checked everything in detail, but looks reasonable

@fieker if you are happy, just merge

src/GenOrd/FractionalIdeal.jl
```diff
@@ -313,17 +395,35 @@ Base.:*(I::GenOrdIdl, c::Generic.RationalFunctionFieldElem) = c * I
...
+
+  # The norm of a fractional ideal is a quotient of two coefficient ring elements,
+  #   and it has to stay in the fraction field of the *coefficient* ring:
+  #   for example, factor() splits it into numerator and denominator and hands the parts to
+  #   prime_decomposition, which for the infinite order needs KInftyRing elements
+  #   rather than k[x] polynomials.
+  # Since we consider orders as extensions of the fraction field (of coefficient ring),
+  #   this is sound mathematically too. For the infinite order (KInftyRing),
+  #   the coefficient ring is not k[x]
+  # NOTE: Nemo's fraction_field(ZZ) returns QQField, so we special case here
```
thofma, line 423:
```suggestion
  # NOTE: Nemo's fraction_field(ZZ) returns QQField, so we special case here (upstream bug: https://github.com/Nemocas/Nemo.jl/issues/2351)
```

fieker:
I will merge - but have questino for future thinking: hat about hashing? And can we use a similar trick as with the HNF to get intersections?

alexey-orlov-math:
> I will merge - but have questino for future thinking: hat about hashing? And can we use a similar trick as with the HNF to get intersections?

So for hashing i left the current "hash numerator and denominator" so it forces HNF. We should definitely revisit it
For intersection/add with interim HNF - will revisit too; i am currently optimizing infinite order operations and tweaking some things already

Issue #2218, opened 2026-03-20
gcd_sircana crash+assert


alexey-orlov-math:
When writing tests for HessQR i have encountered

```
julia> F, mF = residue_ring(ZZ, 12); x = F[:x][2];

julia> Hecke.gcd_sircana(x^2+1,x+1)
ERROR: DomainError with -1:
Index must be non-negative
Stacktrace:
 [1] coeff(x::ZZModPolyRingElem, n::Int64)
   @ Nemo ~/.julia/packages/Nemo/xSaoQ/src/flint/fmpz_mod_poly.jl:49
 [2] fun_factor(f::ZZModPolyRingElem)
   @ Hecke ~/math/software/Hecke.jl/src/Misc/nmod_poly.jl:1050
 [3] gcd_sircana(f::zzModPolyRingElem, g::zzModPolyRingElem)
   @ Hecke ~/math/software/Hecke.jl/src/Misc/nmod_poly.jl:829
 [4] top-level scope
   @ REPL[20]:1
```

and

```
julia> F, mF = residue_ring(ZZ, 210); x = F[:x][2];

julia> Hecke.gcd_sircana(x^2+1,x+1)
ERROR: AssertionError: f * qf == _F
Stacktrace:
 [1] gcd_sircana(f::zzModPolyRingElem, g::zzModPolyRingElem)
   @ Hecke ~/math/software/Hecke.jl/src/Misc/nmod_poly.jl:852
 [2] top-level scope
   @ REPL[24]:1
```

tagging @thofma since he knows a thing or two about this algorithm ;-)

thofma:
It has been a while, but I will have a look.

thofma:
fixed in #2367

PR #2367, opened 2026-08-30
Fixed gcd sircana corner cases

alexey-orlov-math:
Fixed several missed corner cases in sircana gcd implementation:
- Use full powers of the coprime-base factors as CRT moduli so that their product recovers the original, possibly nonsquarefree, modulus.
- Record and unwind Euclidean quotients to reconstruct cofactors of the original polynomials (we can have several Euclidean steps). Include content and Weierstrass unit factors in the reconstructed cofactors as well.
- Build the coprime base from the coefficients of both inputs and make each CRT component primitive before calling fun_factor. Together, these conditions ensure that every nonzero input to fun_factor has a unit coefficient.

Also refactored the implementation to be easier to follow.
Added tests (from https://github.com/thofma/Hecke.jl/issues/2218) and restored disabled HessQR tests

To ease the review, it is advisable to check two commits one-by-one: first is a pure bug fix, the second is pure refactoring (the only "real" change is `@assert` -> `@hassert` for the result check.

### DIFF
--- a/src/Misc/nmod_poly.jl
+++ b/src/Misc/nmod_poly.jl
@@ -809,45 +809,86 @@ function gcd_sircana(f::PolyRingElem{T}, g::PolyRingElem{T}) where T <: ResElem{
   @assert isone(content(f))
   @assert isone(content(g))
   local qf, qg
+  quots = typeof(f)[]
   while !iszero(g)
     if !is_unit(leading_coefficient(g))
-      cp = coprime_base(vcat(map(x->gcd(lift(x), modulus(R)), coefficients(g)), [modulus(R)]))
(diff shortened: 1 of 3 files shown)


alexey-orlov-math:
tagging @thofma

thofma:
```
F, = residue_ring(ZZ, 4)
Fx, x = polynomial_ring(F, :x)

f = x^2 + x + 2       # (x + 2)*(x + 3)
g = x^2 + 3*x + 2     # (x + 2)*(x + 1)

d, qf, qg = Hecke.gcd_sircana(f, g)
```
This yields `d == 1`, but `h = x + 2` is a proper divisor `f` and `g`.

I am not sure we can just scale with the content when normalizing, since it might be a non-unit?

PR #2369, opened 2026-08-30
fix: ray class group map for factored ideals

### LINKED ISSUE #1289: Ray class group map could be a bit more intelligent
We can map factored ideals to ray class groups, but the individual factors need to be coprime to the modulus:

https://github.com/thofma/Hecke.jl/blob/60a055cf6d9a54784a3dca18e56c090b20282e50/src/NumFieldOrd/NfOrd/RayClassGrp.jl#L785-L792

I think we can do better. Here is a quick example to reproduce the issue.

```julia
julia> K, a = rationals_as_number_field();

julia> OK = maximal_order(K);

julia> M = 12*OK;

julia> R, f = ray_class_group(M);

julia> I = FacElem([9*OK, 3*OK, 5 * OK], ZZRingElem[1, -2, 1]);

julia> Hecke.assertions(true);

julia> f\(numerator(evaluate(I)))
Element of R with components [0 1]

julia> f\(I)
ERROR: AssertionError: $(Expr(:escape, :(is_coprime(J, m))))
Stacktrace:
```

thofma:
Closes #1289


fieker:
its an improvement, but if this is ever critical, I'd change it to run simplify only on the part that is not coprime...ie. go through the factored ideal one-by-one, collect the non-coprime factors elsewhere, processing the ok ones directly. In the end, simplify the non-coprime ones and process

PR #2386, opened 2026-09-05
Fix closure boxes and test for them in CI

### LINKED ISSUE #2167: Use Test.detect_closure_boxes
https://github.com/JuliaLang/julia/pull/60478

fingolfin:
Resolves #2167.

Disclaimer: I did *not* verify every single fix. More notes at the end. First, the Claude summary:

---

A captured variable that is assigned more than once, assigned inside a closure, or used before its definition is stored in a `Core.Box` and inferred as `Any`. `Test.detect_closure_boxes` (Julia 1.14, JuliaLang/julia#60478) flagged 258 methods in Hecke and 2 in GAPExt. This PR fixes all of them and adds `test/ClosureBoxes.jl`, which fails on any new box (no-op below Julia 1.14, so it only bites on the nightly CI job). Same approach as Nemocas/AbstractAlgebra.jl#2498.

Fix idioms: single-assigned names, if-expressions, loop-local aliases, `Ref` for closure-mutated state, helper functions for two-branch or recursive code.

Commits are split by area to ease review. Happy to split into several PRs if preferred.

### Behaviour changes found on the way

- `_quadratic_form_with_invariants` over `QQ` referenced an undefined `q` when `-det` is a local square at a prime in `finite`; it now raises the intended error.
- `_shortest_vectors_span_with_is_shorter` divided the returned map by `d` twice.
- `random_ideal_with_norm_up_to` called the nonexistent `find`; now `findall`.
- `multiplicative_group` errors on an unknown `task` instead of hitting undefined variables.
- Unreachable code after `return` removed in `neq.jl`, `bley_johnston.jl`, `LocalField/Elem.jl`, `MultDep.jl`, `SubgroupEnum.jl`; a disabled debug block in `MultDep.jl` removed.

---

I realize this is *big*. We can split this up. E.g. I could turn each commit in here into a separate PR. Or something else. Let me know what we prefer.

Also, some of these fixes are more clear than other. I don't claim to fully understand all at this point...

so perhaps a better way of splitting this up would be alone the type of fix, instead of along files. Hmmm


thofma:
Can you do the separate PR per commit thing?

fingolfin:
I force-pushed a rewrite of the history. The final tree is unchanged, but the commits are now grouped by kind of change instead of by area, from trivial to invasive:

1. `Fix closure boxes by using fresh variable names`: pure renames, a reassigned captured variable gets a new name instead
2. `Fix closure boxes by using if-expressions`: `x = if ... else ... end` or a ternary instead of assigning in each branch
3. `Fix closure boxes by returning values from closures`: do-blocks and callbacks return the value or update a `Ref`
4. `Fix closure boxes by extracting helper functions`: multi-branch code and recursive local functions moved to top-level helpers
5. `Fix remaining closure boxes`: loops, broadcasting, in-place updates
6. `Remove dead code and fix bugs found via closure boxes`: the only commit with behaviour changes, listed in its message
7. `Test for closure boxes in CI`

I feel that it is actually easier to review, because one has to switch less "mentally" between different methods. Also, the first one (renaming vars) should be fairly uncontroversial, while others may need more discussion. Indeed, I already spotted a few changes I found questionable and got the AI to modify. E.g. it now dropped the `comp_summand_fmpz` rename: the unused `Int` method is deleted instead.

Next, I'll open separate PRs based on slices.

fingolfin:
One more disclaimer: I did not test the separate commits, so there may be CI failures. I will try to fix them as they appear (and when I am awake again and at a computer ;-) )

JHanselman:
I just want to remark that the Riemann surfaces code in the master branch is extremely outdated. Some parts of it are completely rewritten in #2241 . It might be better to fix the closure boxes for the Riemann surface stuff in that branch.

fingolfin:
@JHanselman OK, understood, so what is the status of #2241 in general, when will it be merged?

I see that it reports merge conflicts, possibly caused (in parts or entirely) due to the changes here. If you want, I can make a PR targeting your branch to resolve them. Though depending on what @thofma wants, we should decided in which order: either we merge all the "closure box" fixes first, and then fixup your branch (and possibly others, my offer extends to all of them -- between @lgoettgens and myself, we should have enough free Claude minutes). Or we fix your branch now, and merge it before resuming with the closure box work.

JHanselman:
> @JHanselman OK, understood, so what is the status of #2241 in general, when will it be merged?
> 
> I see that it reports merge conflicts, possibly caused (in parts or entirely) due to the changes here. If you want, I can make a PR targeting your branch to resolve them. Though depending on what @thofma wants, we should decided in which order: either we merge all the "closure box" fixes first, and then fixup your branch (and possibly others, my offer extends to all of them -- between @lgoettgens and myself, we should have enough free Claude minutes). Or we fix your branch now, and merge it before resuming with the closure box work.

We'll try to get it merged now. It's inside a module anyway, so it won't affect regular users, and anyone who is working with Riemann surfaces that I know of is using branch #2241 anyway. (It's more stable and bug free than the version in master.)

PR #2387, opened 2026-09-05
Fix closure boxes by using fresh variable names

fingolfin:

Give each new value its own name, or alias a captured value under a name that is assigned exactly once. No behaviour change.

---

First commit of PR #2386. While it is big, I feel it is fairly easy to review anyway, as this is only renaming variables. Still: if you prefer, I can split it further.

### DIFF
--- a/src/LocalField/neq.jl
+++ b/src/LocalField/neq.jl
@@ -83,7 +83,8 @@ function Nemo.basis(K::FinField, k::FinField)
   b = basis(K)
   K = base_field(K)
   while absolute_degree(K) > absolute_degree(k)
-    b = [x*y for x = basis(K) for y = b]
+    bb = b
+    b = [x*y for x = basis(K) for y = bb]
     K = base_field(K)
   end
   if K != k
@@ -203,7 +204,8 @@ end
 function coordinates(a::Union{QadicFieldElem, LocalFieldElem}, k)
   c = [coeff(a, i) for i=0:degree(parent(a))-1]
   while absolute_degree(parent(c[1])) > absolute_degree(k)
-    c = reduce(vcat, [[coeff(x, i) for i=0:(degree(parent(c[1]))-1)] for x = c])
+    d = degree(parent(c[1]))
+    c = reduce(vcat, [[coeff(x, i) for i=0:d-1] for x = c])
   end
   if parent(c[1]) != k
     if isa(parent(c[1]), QadicField) && degree(parent(c[1])) ==1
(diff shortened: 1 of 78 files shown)


src/NumFieldOrd/NfOrd/Clgp/Sunits.jl
thofma, line 156:
```suggestion
      return log_fac(FacElem(a))
```

src/NumField/NfRel/NEQ.jl
```diff
@@ -57,13 +57,13 @@ function is_norm_fac_elem(K::RelSimpleNumField{AbsSimpleNumFieldElem}, a::AbsSim
...
-  k, _mk = snf(k)
-  mk = _mk*mk
+  k0, mk0 = kernel(No)
+  k, _mk = snf(k0)
+  mk = _mk*mk0
   gk = [mU(mk(g)) for g = gens(k)]
   pushfirst!(gk, mU(so))
-  if length(S) > 0
-    v = matrix(ZZ, length(gk), length(S), [valuation(g, P)*norm(P) for g = gk for P = S])
+  if length(Sc) > 0
```
thofma, line 65: `Sc` is never defined

fingolfin: Yeah the split made some mistakes. I'll get in it -- but first, sleep

fingolfin:
@thofma this is "good to go" from my perspective. The one failing test looks like it might be out of memory? I can't restart it myself, though.

src/LocalField/neq.jl
```diff
@@ -261,10 +263,10 @@ function solve_1_units(a::Vector{T}, b::T) where T
...
 
     val_offset = e .* map(valuation, absolute_basis(K))
-    cur_b = setprecision(b, k+e)
-    cur_a = [setprecision(x, k+e) for x = a]
-    b = setprecision(b, k+e)
-    a = [setprecision(x, k+e) for x = a]
+    b_prec = setprecision(b, k+e)
+    a_prec = [setprecision(x, k+e) for x = a]
+    cur_b = b_prec
+    cur_a = a_prec
```
fieker, line 269: cur_a, a_prec are now identical - hopefully that anti-aliasing implicit in the old code is not important

fingolfin: Aliasing is fine here: `cur_a` and `cur_b` are only ever rebound (`cur_a = [...]`, `cur_b = cur_b^p`, `cur_b = divexact(...)`), never mutated in place, and `a_prec`/`b_prec` are only read. The old code built two equal copies via `setprecision`; the separate copies were never relied upon.

PR #2388, opened 2026-09-05
Fix closure boxes by using if-expressions

fingolfin:

Assign captured variables once via `x = if ... else ... end` or a ternary instead of assigning in each branch. No behaviour change.

---

First commit of PR #2386. It also contains a few variable renaming which one might argue would fit better into PR #2387, but untangling them would be quite a bit more work, so I hope that's fine.

Again, this is fairly big: if you prefer, I can split it further (e.g. along directories or so).


thofma:
should we do this next? if so, we need to resolve the conflict first

fingolfin:
updated

fingolfin:
@thofma the failed Oscar.jl test ended in a segfault. Could it be a memory issue?

fingolfin:
@JHanselman OK, understood, so what is the status of #2241 in general, when will it be merged?

I see that it reports merge conflicts, possibly caused (in parts or entirely) due to the changes here. If you want, I can make a PR targeting your branch to resolve them. Though depending on what @thofma wants, we should decided in which order: either we merge all the "closure box" fixes first, and then fixup your branch (and possibly others, my offer extends to all of them -- between @lgoettgens and myself, we should have enough free Claude minutes). Or we fix your branch now, and merge it before resuming with the closure box work.

src/QuadForm/Quad/GenusRep.jl
```diff
@@ -1546,12 +1546,7 @@ function _genus_representatives_binary_quadratic_definite_helper(L::QuadLat; max
   F, z = number_field(t^2 - de^2 * d, "z", cached = false)
   # TODO: Use automorphism_group (once implemented for relative extensions)
   a1, a2 = automorphism_list(F)
-  local sigma::morphism_type(F)
-  if a1(z) == z
-    sigma = a2
-  else
-    sigma = a1
-  end
+  sigma = a1(z) == z ? a2 : a1
```
simonbrandhorst, line 1549: Can you please explain why the old code was bad? So that I can avoid it in the future?
I don't really see the difference to the new code.

src/QuadForm/Torsion.jl
```diff
@@ -2265,11 +2265,9 @@ function _biproduct(
 )
   mbf = modulus_bilinear_form(x[1])
   @req all(q -> modulus_bilinear_form(q) == mbf, x) "All torsion quadratic modules must have the same bilinear modulus"
+  mqf = as_bilinear_module ? mbf : modulus_quadratic_form(x[1])
   if !as_bilinear_module
-    mqf = modulus_quadratic_form(x[1])
     @req all(q -> modulus_quadratic_form(q) == mqf, x) "All torsion quadratic modules must have the same quadratic modulus"
-  else
-    mqf = mbf
   end
```
simonbrandhorst, line 2271: Or similar here: Why use the ? :  construction instead of an if else? I thought this was purely a matter of taste?

src/QuadForm/Quad/GenusRep.jl
thofma: There are explanations here: https://github.com/oscar-system/Oscar.jl/pull/6238/changes/21f030e8bb3591e8e959bf5255e96683c74f85df

simonbrandhorst: Thanks

PR #2389, opened 2026-09-05
Fix closure boxes by returning values from closures

fingolfin:

Closures (do-blocks, callbacks) that assigned outer variables now return the value instead, or update a `Ref`. No behaviour change.

---

Third commit of PR #2386.


thofma:
Is it possible to make a PR without the changes from https://github.com/thofma/Hecke.jl/pull/2388?

PR #2397, opened 2026-09-09
sb/fix reduced characteristic vectors

simonbrandhorst:
Fix `_reduced_characteristic_vectors` for negative definite lattices. We hardcoded the positive definite cartan matrix.
If the lattice is negative definite, that bites.


thofma:
deja vu

simonbrandhorst:
Merged master. Now the diff should make sense. Sorry.

thofma (approved):

PR #2399, opened 2026-09-11
Pass on keywords in representatives(::ZZGenus)

simonbrandhorst:
- **forward keyword arguments from representatives(::ZZGenus) to enumerate_definite_genus**

We remove the @attr macro because that chokes on `kwargs...`

### DIFF
--- a/src/QuadForm/IntegerLattices/ZLattices.jl
+++ b/src/QuadForm/IntegerLattices/ZLattices.jl
@@ -767,7 +767,7 @@ end
 
 Return representatives for the isometry classes in the genus of `L`.
 """
-function genus_representatives(_L::ZZLat)
+function genus_representatives(_L::ZZLat; kwargs...)
   if rank(_L) <= 1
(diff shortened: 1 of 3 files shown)


src/QuadForm/IntegerLattices/ZLattices.jl
```diff
@@ -785,7 +785,7 @@ function genus_representatives(_L::ZZLat)
       push!(res, _to_ZLat(N; K=QQ))
     end
   elseif is_definite(L)
-    res = enumerate_definite_genus(L)
+    res = enumerate_definite_genus([L], :default; kwargs...)
```
StevellM:
```suggestion
    res = first(enumerate_definite_genus([L], :default; kwargs...))
```
I think this version of `enumerate_definite_genus` has 3 outputs (missing mass and dictionary of invariants). This should probably solve the breaking test

StevellM (approved):
Modulo the small comment, looks good to me.

